PR #1, opened 2021-07-06
Refactor BasedOnUbi check and corresponding unit test

bcrochet:
Just passing through the params to a validate method seemed over
the top just to test. This breaks out the gathering of the output
from the actual test. This allows us to unit test the actual part
of the check without caring how that data was collected.


certification/internal/shell/base_on_ubi.go
```diff
@@ -32,6 +29,10 @@ func (p *BaseOnUBICheck) validate(podmanEngine cli.PodmanEngine, image string, l
 
 	lines := strings.Split(runReport.Stdout, "\n")
 
+	return p.checkBasedOnUbi(lines)
+}
+
+func (p *BaseOnUBICheck) checkBasedOnUbi(lines []string) (bool, error) {
```
samira-barouti, line 35: Can we refactor this into a generic function (sth like `parse`) and have it included in the interface? It seems like all of our tests need to parse a command's output.

certification/internal/shell/base_on_ubi.go
```diff
@@ -12,10 +12,7 @@ type BaseOnUBICheck struct{}
 
 func (p *BaseOnUBICheck) Validate(image string, logger *logrus.Logger) (bool, error) {
```
samira-barouti, line 13: I like the idea of splitting the command execution from the output parsing. It however does not solve the public interface testing; I mean we are not still unit testing the `Validate()` method which is exposed externally. 

If there is no easy way to unit test `Validate()` method, I am good with testing the `checkBasedOnUbi` 👍

tonytcampbell: +1 to @samira-barouti comment.  BaseOnUBICheck.checkBasedOnUbi makes the code less readable IMO.  Would simplifying the Validate function signature help?  Instead of passing in logger for example, we just import it 

`import (
  "os"
)`

bcrochet: I am definitely in favor of removing logger. That would simplify that interface. There are some other tricks that can be employed to test the public interface. Let me take a stab at it.

certification/internal/shell/base_on_ubi.go
bcrochet: Not all do that. And I'd like to see the output from the engine become more structured. See #2

komish: +1 to @samira-barouti's request here. If we have multiple validations that require "parsing the output of `containercli run ...` then we probably would see a benefit of abstracting that out and testing that independently. 

Could we have that drafted into a new issue+PR?

---
Separately, bit of a nitpick regarding naming - I'd prefer to keep this as `validate` over `checkBasedOnUbi`. I don't see an issue with reusing the public method name as a private method name with the internal working defined.

certification/internal/shell/base_on_ubi_test.go
```diff
@@ -1,41 +1,38 @@
...
-			RunReportStderr: "",
-		}
+`, "\n")
 	})
 
 	Describe("Checking for UBI as a base", func() {
 		Context("When it is based on UBI", func() {
 			It("should succeed the check", func() {
-				ok, err := base_on_ubi.validate(engine, "dummy/image", logger)
+				ok, err := base_on_ubi.checkBasedOnUbi(lines)
```
komish, line 25: Prefer the `validate` name as mentioned in an earlier comment. Just leaving this comment here so we consciously address it or leave it be (based on whatever choice is made).

komish (changes requested):

certification/internal/shell/base_on_ubi.go
komish: @bcrochet I'm missing it, what challenges arise from the logger being present in the signature?

bcrochet: @komish The logger itself doesn't really introduce challenges, but it really doesn't belong IMHO. See #13 . The problem of the exec needed to be mocked is still a concern.

komish: Understood. Agreed, don't like it in the signature.

certification/internal/shell/base_on_ubi.go
bcrochet: If it ends up being part of the interface, that's great. But I don't think we necessarily need to add it just yet. Let's see how many end up actually needing it.

certification/internal/shell/base_on_ubi.go
bcrochet: logger piece handled in another PR. The Validate part should be handled now.

certification/internal/shell/base_on_ubi_test.go
bcrochet: I think I've addressed all of this.

certification/internal/shell/base_on_ubi_test.go
```diff
@@ -1,41 +1,65 @@
...
+				podmanEngine = fakeEngine
+			})
 			It("should succeed the check", func() {
-				ok, err := base_on_ubi.validate(engine, "dummy/image")
+				ok, err := baseOnUbiCheck.validate(lines)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(ok).To(BeTrue())
+			})
+			It("should pass Validate", func() {
+				ok, err := baseOnUbiCheck.Validate("dummy/image")
```
samira-barouti, line 44: Now I have the peace of mind ;) Thanks!

IMO testing the Validate() method is enough. It will test the validate() automatically and we don't need to maintain implementation in both tests and the method (e.g. line 27: if at some point we decide to split lines based on another char, we have to update the code and the test).

komish:
Runs on my machine 

```json
time="2021-07-08T13:15:12-05:00" level=info msg="check completed: BasedOnUbi" result=PASSED
{
    "image": "registry.access.redhat.com/ubi8/ubi:latest",
    "validation_lib_version": {
        "version": "0.0.0",
        "commit": "c891eab1641ca1fdad35285f323e75bfa8bc01ff"
    },
    "results": {
        "Passed": [
            {
                "description": "Checking if the container's base image is based on UBI",
                "level": "best",
                "knowledge_base_url": "https://connect.redhat.com/zones/containers/container-certification-policy-guide",
                "check_url": "https://connect.redhat.com/zones/containers/container-certification-policy-guide"
            }
        ],
        "Failed": [],
        "Errors": []
    }
}
```

certification/internal/shell/base_on_ubi_test.go
```diff
@@ -1,41 +1,65 @@
 package shell
```
samira-barouti, line 1: Is it possible to move the tests to a separate directory?

komish (approved):
Looks good. I think the test suite could probably use some comments to give readers some insight into the implementation of the test (how we're faking and why), but it works, so 👍 

/lgtm

samira-barouti (commented):
Two small comments

certification/internal/shell/base_on_ubi_test.go
komish: @samira-barouti Technically, yes! For example, an integration test might live elsewhere and don't really need to live next to packages.

With that said, storing test code for a package alongside the package is idiomatic for most unit test cases from what I've seen. Also the below snippet from the golang testing docs (though we use Ginkgo which integrates well with the go test tooling):

> To write a new test suite, create a file whose name ends _test.go that contains the TestXxx functions as described here. Put the file in the same package as the one being tested. The file will be excluded from regular package builds but will be included when the "go test" command is run. For more detail, run "go help test" and "go help testflag".
reference: https://golang.org/pkg/testing/

Storing the test also gives @bcrochet the ability to leverage any local-only variables/functions etc. that he might need or want to test. I think the only one that is in scope here is the podman engine.

As an aside (and not contextually related to this PR), you can also create a test file in the same package using a new package name `<packagename>_test` which gives you the ability to do black-box testing, and only focus on exported functionality.

samira-barouti: Understood. I was just thinking as we add more checks, this directory will become cumbersome . I am good with it at the moment, however we need to start splitting this package into subpackages (probably leveraging container vs bundle sub packages).

certification/internal/shell/base_on_ubi_test.go
bcrochet: I was literally thinking about that as well. Let me address that and add some comments about the fakery and then I'll repush.

samira-barouti (approved):
Looks good to me!

Just a small note: is it possible to test for the erroneous condition too? In the case that podmanEngine.Run fails and returns an error.

certification/internal/shell/base_on_ubi_test.go
bcrochet: The way that I'm doing the mocking would break if we moved the tests to a different directory/package. As @komish pointed out, this is the idiomatic way.

samira-barouti (approved):

certification/internal/shell/base_on_ubi.go
```diff
@@ -1,21 +1,28 @@
...
 
 type BaseOnUBICheck struct{}
 
 func (p *BaseOnUBICheck) Validate(image string) (bool, error) {
-	podmanEngine := PodmanCLIEngine{}
-	return p.validate(podmanEngine, image)
+	lines, err := p.getDataToValidate(image)
+	if err != nil {
+		log.Debugf("Stdout: %s", lines)
+		return false, fmt.Errorf("%w: %s", errors.ErrFormattingResults, err)
```
komish, line 19: Hmm, should we use a different error here? Under the premise that "Results" as a concept refers to the validation results in its entirety?

certification/internal/shell/base_on_ubi_test.go
```diff
@@ -3,42 +3,70 @@ package shell
...
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/cli"
+	log "github.com/sirupsen/logrus"
 )
 
+// podmanEngine is a package-level variable. In some tests, we
+// override it with a "happy path" engine, that returns good data.
+// In the unhappy path, we override it with an engine that returns
+// nothing but errors.
```
komish, line 13: This comment is epic and very clear! Nothing to change here, just wanted to thank you for adding this. Resolve this thread whenever 😆

certification/internal/shell/base_on_ubi_test.go
```diff
@@ -3,42 +3,70 @@ package shell
...
 			RunReportStdout: `ID="rhel"
 NAME="Red Hat Enterprise Linux"
 `,
 			RunReportStderr: "",
 		}
+		originalEngine = podmanEngine
+	})
+	AfterEach(func() {
+		podmanEngine = originalEngine
 	})
```
komish, line 33: Topic for thought: This block basically becomes a required part of every series of tests in this suite, right? So every check that leverages the FakePodmanEngine will need to do this BeforeEach and AfterEach task to make sure they're not inadvertently touching the package-level variable and impacting other tests?

Again - nothing to change here. Just trying to zoom out a bit. Feel free to resolve whenever.

certification/internal/shell/fake_podman_engine_test.go
```diff
@@ -0,0 +1,55 @@
+package shell
```
komish, line 1: Nit: To @samira-barouti's request elsewhere, I think _this_ can live in its own package. I know you added the test _test suffix to make sure it's only called into scope when testing, but I wonder if we can organize it elsewhere. Up to you if you think it's worth doing now or subsequently.

I also think `"github.com/redhat-openshift-ecosystem/openshift-preflight/cli"` should move elsewhere as well. We can discuss later outside of this PR.

komish (changes requested):
One or two small change requests, and a couple of comments/discussion points. Thanks @bcrochet!

certification/internal/shell/base_on_ubi.go
bcrochet: Sure. I'll change that. Probably just a case of me being lazy.

certification/internal/shell/base_on_ubi_test.go
bcrochet: I will address that in a different PR. (Probably #2) once this lands.

certification/internal/shell/fake_podman_engine_test.go
bcrochet: Sounds good. Let's go with future us.

samira-barouti (approved):

komish (approved):
/lgtm

PR #3, opened 2021-07-06
Remove references to unimplemented design concepts and add basic usage statement to top-level readme

komish:
There were several concepts in the initial dev docs draft that we did not implement (e.g. `inputmanager`). These are being removed from the dev docs.

In addition, I just added a simple text blurb to the README indicating the intention of the project, and reflecting the current build's `--help` output. We'll refine this more as time goes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,1 +1,30 @@
 # Preflight
+
+**Preflight** is a commandline interface for validating if
+[OpenShift](https://www.openshift.com/) operator bundles meet minimum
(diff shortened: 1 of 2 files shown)


samira-barouti (approved):
Do we plan to add a `developer guide` doc as well?

komish:
> Do we plan to add a `developer guide` doc as well?

Definitely! I think both developer and user guides, as well as the various use case example guides are in scope for the future. I'll be looking to help write those as we get more stable!

PR #5, opened 2021-07-06
Add version flag support

komish:
Just a quick addition - I meant to wire this up previously. This just adds the ability to call `./preflight --version` and have the CLI respond.

I added a `String()` method to format a `VersionContext` on a single line, which we then use in our cobra root.go definition. Cobra, then, handles the actual addition of the `-v` or `--version` flag.

Finally, this updates the Makefile so that we can build the CLI with a RELEASE_TAG environment variable and have it injected into the binary. Example:

```
$ RELEASE_TAG=1.2.3 make build
go build -o preflight -ldflags "-X github.com/redhat-openshift-ecosystem/openshift-preflight/version.commit=52f97749b3ac5fbcdf4877bd31582221e592bce6 -X github.com/redhat-openshift-ecosystem/openshift-preflight/version.version=1.2.3" main.go

$ ./preflight --version
preflight version 1.2.3 <commit: 52f97749b3ac5fbcdf4877bd31582221e592bce6>
```

By default, this will inject `0.0.0` indicating that no release tag was provided. 

I also updated the README to include the help output which shows the version flag. I'm noticing that the available test and formatters output changed, and that's probably because we internally use a slice there and I'm assuming ordering is changing. When README gets more refined, we'll need to be cognizant of that.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -9,12 +9,14 @@ import (
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/errors"
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/formatters"
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/runtime"
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/version"
 	"github.com/spf13/cobra"
 )
 
 var rootCmd = &cobra.Command{
-	Use:   "preflight <container-image>",
-	Short: "Preflight Red Hat certification prep tool.",
+	Use:     "preflight <container-image>",
+	Short:   "Preflight Red Hat certification prep tool.",
+	Version: version.Version.String(),
 	Long: "A utility that allows you to pre-test your bundles, operators, and container before submitting for Red Hat Certification." +
 		"\nChoose from any of the following checks:" +
 		"\n\t" + strings.Join(engine.AllChecks(), ", ") +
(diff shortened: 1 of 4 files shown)


jomkz (approved):
/lgtm

cmd/root.go
```diff
@@ -9,12 +9,14 @@ import (
...
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/formatters"
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/runtime"
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/version"
 	"github.com/spf13/cobra"
 )
 
 var rootCmd = &cobra.Command{
-	Use:   "preflight <container-image>",
-	Short: "Preflight Red Hat certification prep tool.",
+	Use:     "preflight <container-image>",
```
samira-barouti, line 17: For OpenScap scan we need the root privilege. So this should be `sudo preflight <container-image>`

samira-barouti (changes requested):

cmd/root.go
komish: This was originally `preflight <container-image>`, and this PR doesn't change anything other than spacing (auto-alignment done by the linter) changed due to the new `Version` line being longer.

That said, I'm not sure I agree that we should indicate users should user `sudo` here. Given the enabled checks are configurable, and the check utilizing openscap is only a single check, I think it makes more sense to leave this as is (and document elsewhere if cases where `sudo` should be used).

samira-barouti: I believe we should indicate running all checks and HasMinimalVulnerabilities requires the root privilege, otherwise the policy execution will fail. I could probably take care of it in the other PR that I opened.

samira-barouti (approved):

PR #6, opened 2021-07-06
Migrate check to scan the containers for unacceptable vulnerabilities

samira-barouti:
* The check is using the `openscap-podman` command output to identify the number of vulnerabilities.
* This check requires `root` privileges.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,9 @@
 # Output of the go coverage tool, specifically when used with LiteIDE
 *.out
 
+# openscap vulnerability report
+vuln.html
+
 # Binaries for programs and plugins
 *.exe
 *.exe~
@@ -12,6 +15,7 @@
 *.dylib
 bin
 preflight
+openshift-preflight
 preflight-*.log
 
 # Test binary, build with `go test -c`
--- a/certification/internal/shell/has_minimal_vulns.go
+++ b/certification/internal/shell/has_minimal_vulns.go
@@ -1,15 +1,107 @@
 package shell
 
 import (
+	"bytes"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"regexp"
+
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification"
-	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/errors"
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/internal/utils/file"
+	log "github.com/sirupsen/logrus"
+)
+
+const (
+	ovalFilename = "rhel-8.oval.xml.bz2"
+	ovalUrl      = "https://www.redhat.com/security/data/oval/v2/RHEL8/"
+	reportFile   = "vuln.html"
 )
 
 type HasMinimalVulnerabilitiesCheck struct{}
 
 func (p *HasMinimalVulnerabilitiesCheck) Validate(image string) (bool, error) {
-	return false, errors.ErrFeatureNotImplemented
+
+	ovalFileUrl := fmt.Sprintf("%s%s", ovalUrl, ovalFilename)
+	dir, err := ioutil.TempDir(".", "oval-")
+
+	if err != nil {
+		log.Error("Unable to create temp dir", err)
+		return false, err
+	}
+	log.Debugf("Oval file dir: %s", dir)
+	defer os.RemoveAll(dir)
+
+	ovalFilePath := filepath.Join(dir, ovalFilename)
+	log.Debugf("Oval file path: %s", ovalFilePath)
+
+	err = file.DownloadFile(ovalFilePath, ovalFileUrl)
+	if err != nil {
+		log.Error("Unable to download Oval file", err)
+		return false, err
+	}
+	// get the file name
+	r := regexp.MustCompile(`(?P<filename>.*).bz2`)
+	ovalFilePathDecompressed := filepath.Join(dir, r.FindStringSubmatch(ovalFilename)[1])
+
+	err = file.Unzip(ovalFilePath, ovalFilePathDecompressed)
+	if err != nil {
+		log.Error("Unable to unzip Oval file: ", err)
+		return false, err
+	}
+
+	numOfVulns, err := p.numberOfVulnerabilities(image, ovalFilePathDecompressed)
+	if err != nil {
+		return false, err
+	}
+
+	log.Debugf("The number of found vulnerabilities: %d", numOfVulns)
+
+	return numOfVulns == 0, nil
 }
+
+// numberOfVulnerabilities takes the `image` and the path to the OVAL file. It runs
+// `oscap-podman` against `image`, and parses the output to figure out how many vulnerabilities
+//  exist. It also saves results to the vuln.html file in the current directory,
+// for the end users' reference.
+
+func (p *HasMinimalVulnerabilitiesCheck) numberOfVulnerabilities(image string, ovalFilePathDecompressed string) (int, error) {
+
+	// run oscap-podman command and save the report to vuln.html
+	cmd := exec.Command("oscap-podman", image, "oval", "eval", "--report", reportFile, ovalFilePathDecompressed)
+	var out bytes.Buffer
+	cmd.Stdout = &out
+	var stderr bytes.Buffer
+	cmd.Stderr = &stderr
+	err := cmd.Run()
+
+	if err != nil {
+		log.Error("unable to execute oscap-podman on the image: ", cmd.Stderr, reportFile)
+		return -1, err
+	}
+
+	// get the current directory
+	path, err := os.Getwd()
+	if err != nil {
+		log.Error("unable to get the current directory: ", err)
+	}
+
+	log.Debugf("The path to vulnerability report: %s/%s ", path, reportFile)
+
+	// use regex to count the number of vulnerabilities
+
+	// oscap-podman writes `Definition oval:com.redhat.<CVE#>: true` to stdout
+	// if the vulnerability exist in the container, and `Definition oval:com.redhat.<CVE#>: false`
+	// if the vulnerability does not exist
+	r := regexp.MustCompile("Definition oval:com.redhat.*: true")
+	matches := r.FindAllStringIndex(string(out.String()), -1)
+
+	// count the number of matches
+	return len(matches), nil
+}
+
 func (p *HasMinimalVulnerabilitiesCheck) Name() string {
 	return "HasMinimalVulnerabilities"
 }
--- a/certification/internal/utils/file/helper.go
+++ b/certification/internal/utils/file/helper.go
@@ -1,0 +1,50 @@
+package file
+
+import (
+	"compress/bzip2"
+	"io"
+	"net/http"
+	"os"
+)
+
+func DownloadFile(filename string, url string) error {
+	resp, err := http.Get(url)
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+
+	out, err := os.Create(filename)
+	if err != nil {
+		return err
+	}
+	defer out.Close()
+
+	// Write the body to file
+	_, err = io.Copy(out, resp.Body)
+	return err
+}
+
+func Unzip(bzipfile string, destination string) error {
+
+	f, err := os.Open(bzipfile)
+	if err != nil {
+		return err
+	}
+	defer f.Close()
+
+	in := bzip2.NewReader(f)
+
+	out, err := os.Create(destination)
+
+	if err != nil {
+		return err
+	}
+	_, err = io.Copy(out, in)
+
+	if err != nil {
+		return err
+	}
+	out.Close()
+	return nil
+}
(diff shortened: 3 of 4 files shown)


certification/internal/utils/utils.go
```diff
@@ -0,0 +1 @@
+package utils
```
tonytcampbell, line 1: I believe Golang best practices recommend avoiding packages named util as the package could be a catch all for lots of unrelated things. Is there a more descriptive package name we could use?

samira-barouti: That makes sense. How does `preflightutils` sound?

.gitignore
```diff
@@ -4,14 +4,17 @@
...
+
 # Binaries for programs and plugins
 *.exe
 *.exe~
 *.dll
 *.so
 *.dylib
 bin
-preflight
+openshift-preflight
```
komish, line 17: I think adding `openshift-preflight` is fine, but I think the Makefile still gives us a `preflight` binary. Let's keep both?

certification/internal/shell/has_minimal_vulns.go
```diff
@@ -1,16 +1,96 @@
...
+		logger.Error("unable to get the current directory: ", err)
+	}
+
+	logger.Debugf("The path to vulnerability report: %s/%s ", path, reportFile)
+
+	r := regexp.MustCompile("Definition oval:com.redhat.*: true")
+	matches := r.FindAllStringIndex(string(out.String()), -1)
+	return len(matches), nil
+}
+
```
komish, line 93: Should this be a struct method on `HasMinimalVulnerabilitiesCheck`?

certification/internal/utils/utils.go
komish: Maybe just remove *utils.go` so we're not defining a utils package? Then you keep the import path? Would that work @tonytcampbell?

The blurb on package names, for reference:

> Avoid meaningless package names. Packages named util, common, or misc provide clients with no sense of what the package contains. This makes it harder for clients to use the package and makes it harder for maintainers to keep the package focused. Over time, they accumulate dependencies that can make compilation significantly and unnecessarily slower, especially in large programs. And since such package names are generic, they are more likely to collide with other packages imported by client code, forcing clients to invent names to distinguish them.
Source: https://blog.golang.org/package-names

certification/internal/utils/file/helper.go
```diff
@@ -0,0 +1,50 @@
...
+		return err
+	}
+	_, err = io.Copy(out, in)
+
+	if err != nil {
+		return err
+	}
+	out.Close()
+	return nil
+}
```
komish, line 50: Can you write unit tests here?

certification/internal/shell/has_minimal_vulns.go
```diff
@@ -1,16 +1,96 @@
...
+	ovalFileUrl := fmt.Sprintf("%s%s", ovalUrl, ovalFilename)
+
+	tempdir, err := os.MkdirTemp(os.TempDir(), "oval-*")
+	if err != nil {
+		logger.Error("Unable to create temp dir", err)
+		return false, err
+	}
+	logger.Debugf("Oval file dir: %s", tempdir)
+	defer os.RemoveAll(filepath.Dir(tempdir))
```
komish, line 34: Two things here:

1) I don't think you need `file path.Dir(...)` here. This is deleting my `/tmp` which is causing the test to fail in subsequent executions. See [1] below for more details on that.

2) `os.RemoveAll` can throw an error, which makes it not a great candidate for defer. We probably want to let the user know that we couldn't Clea up in this case.

---
Thinking on this - no action needed for this PR as a result of this, just food for thought/discussion:
I wonder if we could figure out a way to elevate our permissions _specifically_ when calling out to oscap-podman. Just seems like a safer approach than calling preflight itself with sudo. As a small scale example, removing `/tmp` probably wouldn't have succeeded if I hadn't called `sudo preflight ...`. 

Not all checks require elevated permissions. So either this check is executed on its own to ensure that all _other_ checks aren't running with more privileges than they need, or we run them all with elevated privileges (what I'm thinking we should avoid doing).

---
[1]

I updated `Validate` to just print what file we were targeting like so:

```go
	logger.Debugf("Oval file dir: %s", tempdir)
	logger.Debug("debug!!!!", filepath.Dir(tempdir)) // printline debugging here, woohoo!
	defer os.RemoveAll(filepath.Dir(tempdir))
```

I got this in the log:

```
time="2021-07-07T10:18:12-05:00" level=debug msg="debug!!!!/tmp"
```

Here's the error that tipped me off that we were removing more than we should have

```
time="2021-07-07T10:18:04-05:00" level=info msg="check completed: HasMinimalVulnerabilities" error="stat /tmp: no such file or directory" result=ERROR
```

komish (changes requested):
Still testing, but I've already added a few comments/change requests for review.

certification/internal/shell/has_minimal_vulns.go
```diff
@@ -1,16 +1,96 @@
...
+
+	// TODO claen up the names
+	ovalFilePath := filepath.Join(tempdir, ovalFilename)
+	logger.Debugf("Oval file path: %s", ovalFilePath)
```
komish, line 38: Still need this TODO?

certification/internal/utils/utils.go
tonytcampbell: Thanks for the blurb @komish.  That explains it well.  @samira-barouti can we just get rid of the utils package altogether?  I think what @komish recommends above would work.

.gitignore
tonytcampbell: +1 to @komish  comment. The binary call should just be `preflight <subcommand>`.  We still need to .gitignore `preflight`

samira-barouti: Not sure why this is happening, but running `go build` on my local (via Vagrant) creates a binary named `openshift-preflight`, and not `preflight`. I have not changed the Makefile and the `-o preflight` is set properly.

certification/internal/utils/utils.go
samira-barouti: I was planning to use this utils package as a parent package which contains multiple sub packages. Right now, we have `file` as a sub-package; I wanted to migrate the `util.go` in the shell package, and create a `container` sub-package. 
I do not have any strong opinion on keeping this; my intention was to introduce a sort of structure in this directory, and enable the future extension of this package to include other utility functions as subpackages. If the consensus is to remove this file, I am good with it :)

komish: @samira-barouti You can leave the directory structure as is even without utils.go file. You would keep the organizational structure that way, but we would have no importable `utils` package. 

You'd still see something like `import .../internal/utils/file` for your import statements. 

Does that make sense?

.gitignore
komish: Are you running `go build` or `make build`? I think the former takes the parent directory name into consideration but doesn't use the make file. The latter does whatever is in the make file (which should spit out `preflight` as the binary).

But either way - keep both in, just in case. 👍

certification/internal/shell/has_minimal_vulns.go
samira-barouti: hmm... it could be.. I see it more of a helper function, specific to this check.That's why I contained it to this type. Do you see it as a generic function?

certification/internal/utils/file/helper.go
samira-barouti: Do you mind if I do unit testing in another PR?

certification/internal/shell/has_minimal_vulns.go
samira-barouti: Thank you Jose for the through investigation... Appreciate your time and effort :) Will fix the issue of removing `/tmp`.

I'll take a look into the `sudo` thing and keep you posted if I found anything.

certification/internal/utils/file/helper.go
komish: @samira-barouti sure thing! I'll mark this resolved.

certification/internal/shell/has_minimal_vulns.go
komish: I'm currently parsing this as a generic helper function _because_ it's not bound to the struct. That is - I would expect any other check to be able to call this function the way it's written currently, so long as the caller is in the same package. 

I would bind it as an internal struct method (i.e. unexported, but still a method on `HasMinimalVulnerabilitiesCheck`) in a case where other checks won't have to use it (e.g. it's specific to this check).

Was your goal to write it so that only this check can use this function? Write now, I'm reading this as anything in the `shell` package can use this function. Could be wrong!

certification/internal/utils/utils.go
samira-barouti: Beautiful! Thank you @komish. That makes sense

.gitignore
samira-barouti: Nice! Did not realize we are using `make`. I need to start a developer guide :)

certification/internal/shell/has_minimal_vulns.go
samira-barouti: Understood. Yes in this case, this must be bound to this struct. Thanks again :)

certification/internal/shell/has_minimal_vulns.go
```diff
@@ -1,15 +1,96 @@
...
+
+	if err != nil {
+		log.Error("unable to execute oscap-podman on the image: ", cmd.Stderr, reportFile)
+		return 0, err
+	}
+
+	path, err := os.Getwd()
+	if err != nil {
+		log.Error("unable to get the current directory: ", err)
+	}
```
komish, line 85: Should we be returning this error here, or is it safe to continue?

certification/internal/shell/has_minimal_vulns.go
```diff
@@ -1,15 +1,96 @@
...
+
+	log.Debugf("The number of found vulnerabilities: %d", numOfVulns)
+	if numOfVulns > 0 {
+		return false, nil
+	}
+
+	return true, nil
 }
+
+func (p *HasMinimalVulnerabilitiesCheck) numberOfVulnerabilities(image string, ovalFilePathDecompressed string) (int, error) {
```
komish, line 68: Can you add a comment here detailing the high level logic? Not strictly required, but a bulk of the work is here, so I think it'll help us when we come back to it in the future.

komish (changes requested):
So sorry to add two more change requests! I completely missed one in the initial review, and the other item is just a quick text blurb.

certification/internal/shell/has_minimal_vulns.go
samira-barouti: It is safe to continue in my opinion. The vulnerability report will be written to the current directory, but the check does not use the report to find out the number of vulnerabilities. The report provides a human readable HTML report to see the scanning result.

komish:
Confirming "works on my machine"

```json
{
    "image": "registry.access.redhat.com/ubi8/ubi:latest",
    "validation_lib_version": {
        "version": "0.0.0",
        "commit": "ef072ff4a141b5322d6517cbf7b4f316c6884023"
    },
    "results": {
        "Passed": [
            {
                "description": "Checking for critical or important security vulnerabilites.",
                "level": "good",
                "knowledge_base_url": "https://connect.redhat.com/zones/containers/container-certification-policy-guide",
                "check_url": "https://connect.redhat.com/zones/containers/container-certification-policy-guide"
            }
        ],
        "Failed": [],
        "Errors": []
    }
}
```

certification/internal/shell/has_minimal_vulns.go
samira-barouti: Please let me know if it is enough details

komish: @samira-barouti just a nit, but one more thing here - it's idiomatic in Go to start function comments with the function name, example:

```go
// numberofVulnerabilities runs the ...
func (p *HasMinimalVulnerabilitiesCheck) numberOfVulnerabilities(image string, ovalFilePathDecompressed string) (int, error) {
...
}
```

Could you swap out `This method` for the method name?

For unexported methods, I don't believe it matters much, but for exported methods - this gets translated into the godoc for that library (which then gets published to pkg.go.dev, etc).

reference: https://blog.golang.org/godoc

samira-barouti: Of course! Will update the method's comment.

tonytcampbell:
Testing now

tonytcampbell (approved):
Confirmed that it fails when expected as also. 👍🏾 

 "Failed": [
            {
                "metadata": {
                    "description": "Checking for critical or important security vulnerabilites.",
                    "level": "good",
                    "knowledge_base_url": "https://connect.redhat.com/zones/containers/container-certification-policy-guide",
                    "check_url": "https://connect.redhat.com/zones/containers/container-certification-policy-guide"
                },
                "helptext": {
                    "message": "Components in the container image cannot contain any critical or important vulnerabilities, as defined at https://access.redhat.com/security/updates/classification",
                    "suggestion": "Update your UBI image to the latest version or update the packages in your image to the latest versions distrubuted by Red Hat."
                }
            }
        ],

PR #7, opened 2021-07-06
Add Dockerfile for Containerizing Preflight

jomkz:
This PR adds the initial Dockerfile to containerize preflight.

## Overview

The preflight binary is built first and the required system packages are then installed, followed by the required external tools (oc, operator-sdk and operator-courier).

In addition, two make targets have been added to build and push the container image. The git hash is used as the image tag, similar to the preflight application version.

Further testing and changes may be required for the podman shell commands to work in all cases but it appears to be working from the limited testing that I have done locally.

## Usage

Running the container without any parameters will display the help.

```
$ podman run -it quay.io/opdev/preflight:f82afd8a7f96c1fcdf1d65a6daed3eeadaba5dbc
A utility that allows you to pre-test your bundles, operators, and container before submitting for Red Hat Certification.
Choose from any of the following checks:
	HasLicense, ValidateOperatorBundle, HasUniqueTag, HasNoProhibitedPackages, RunAsNonRoot, LayerCountAcceptable, HasRequiredLabel, BasedOnUbi, HasMinimalVulnerabilities
Choose from any of the following output formats:
	json, xml, junitxml

Usage:
  preflight <container-image> [flags]

Flags:
  -c, --enabled-checks string   Which checks to apply to the image to ensure compliance.
                                (Env) PREFLIGHT_ENABLED_CHECKS
  -h, --help                    help for preflight
  -o, --output-format string    The format for the check test results.
                                (Env) PREFLIGHT_OUTPUT_FORMAT (Default) json
```

Signed-off-by: John McKenzie <john@mkz.io>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,40 @@
+# golang:1.16 image created 2021-06-24T00:31:06.02014601Z 
+FROM docker.io/library/golang@sha256:be99fa59acd78bb22a41bbc1e15ebfab2262498ee0c2e28c3d09bc44d51d1774 AS builder
+
+# Build the preflight binary
+COPY . /go/src/preflight
+WORKDIR /go/src/preflight
+RUN make build
+
+# ubi-minimal:latest image created 2021-06-01T12:12:46.922866Z
+FROM registry.access.redhat.com/ubi8/ubi-minimal@sha256:0ccb9988abbc72d383258d58a7f519a10b637d472f28fbca6eb5fab79ba82a6b
+
+# Define versions for dependencies
+ARG OPENSCAP_VERSION=1.3.5
+ARG OPENSHIFT_CLIENT_VERSION=4.7.19
+ARG OPERATOR_SDK_VERSION=1.9.0
+
+# Add preflight binary
+COPY --from=builder /go/src/preflight/preflight /usr/local/bin/preflight
+
+# Install dependencies
+RUN microdnf install \
+    buildah \
+    bzip2 \
+    gzip \
+    iptables \
+    podman \
+    skopeo
+
+# Install oscap-podman binary
+RUN curl -L https://github.com/OpenSCAP/openscap/releases/download/${OPENSCAP_VERSION}/openscap-${OPENSCAP_VERSION}.tar.gz | tar -xzv -C /usr/local/bin openscap-${OPENSCAP_VERSION}/utils/oscap-podman
+
+# Install OpenShift client binary
+RUN curl -L https://mirror.openshift.com/pub/openshift-v4/clients/ocp/${OPENSHIFT_CLIENT_VERSION}/openshift-client-linux-${OPENSHIFT_CLIENT_VERSION}.tar.gz | tar -xzv -C /usr/local/bin oc
+
+# Install Operator SDK binray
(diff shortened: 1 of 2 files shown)


jomkz:
This PR did not come across in the repo move for some reason. Since the last update, I have included the `oscap-podman` binary in the image and it appears to be working from my limited testing.

Dockerfile
```diff
@@ -0,0 +1,38 @@
+# golang:1.16 image created 2021-06-24T00:31:06.02014601Z 
+FROM docker.io/library/golang@sha256:be99fa59acd78bb22a41bbc1e15ebfab2262498ee0c2e28c3d09bc44d51d1774 AS builder
+
+# Build the preflight binary
+WORKDIR /go/src/preflight
+COPY . .
```
OchiengEd, line 6: Looks good. Slight suggestion. Could we invert line 5 and 6 and make the copy more prescriptive. Feels a little puzzling in the current state.

```
COPY . /go/src/preflight
WORKDIR /go/src/preflight
```

tonytcampbell (approved):
/lgtm

Looks like a good first iteration. Need to test it but this should get us started. I'll be curious about whether we can do docker in docker from this or podman in podman.

Dockerfile
jomkz: Reordered the commands to be more explicit.

Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+RUN microdnf install \
+    buildah \
+    bzip2 \
+    gzip \
+    iptables \
+    podman \
+    skopeo
+
+# Install OpenShift client binary
+RUN curl -L https://mirror.openshift.com/pub/openshift-v4/clients/ocp/4.7.18/openshift-client-linux-4.7.18.tar.gz | tar -xzv -C /usr/local/bin --exclude=README.md --exclude=kubectl
```
samira-barouti, line 31: same comment here

Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+
+# Install Operator SDK binray
+RUN curl -Lo /usr/local/bin/operator-sdk https://github.com/operator-framework/operator-sdk/releases/download/v1.9.0/operator-sdk_linux_amd64
```
samira-barouti, line 34: same comment here

Dockerfile
```diff
@@ -0,0 +1,38 @@
+# golang:1.16 image created 2021-06-24T00:31:06.02014601Z 
+FROM docker.io/library/golang@sha256:be99fa59acd78bb22a41bbc1e15ebfab2262498ee0c2e28c3d09bc44d51d1774 AS builder
+
+# Build the preflight binary
+COPY . /go/src/preflight
+WORKDIR /go/src/preflight
+RUN make build
+
+# Download and extract openscap 
+RUN curl -L https://github.com/OpenSCAP/openscap/releases/download/1.3.5/openscap-1.3.5.tar.gz | tar -xzv -C /tmp
```
samira-barouti, line 10: It would be great to define the version in an ENV block. Make it easier to change the version in the case of the upgrade.

samira-barouti (approved):

Dockerfile
jomkz: This was something that I was going to do in a follow up PR but it makes sense to just add them now. I added ARG instructions for these, as they are only needed during the build stage and are not needed in the running image.

PR #8, opened 2021-07-06
Add unit tests for config precedence parsing function

komish:
This adds a quick test to ensure that our configuration precedence parsing function works as intended, ensuring flag values > environment values > default values.

### DIFF
--- a/cmd/parse_test.go
+++ b/cmd/parse_test.go
@@ -1,0 +1,39 @@
+package cmd
+
+import (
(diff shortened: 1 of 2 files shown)


tonytcampbell (approved):
lgtm

We should consider using [Viper](https://github.com/spf13/viper) for configs as it has this built in.  Maybe a future PR if we think it's a good idea.

jomkz (approved):
/lgtm

PR #9, opened 2021-07-06
adding separate policy maps for Operators and Containers

tonytcampbell:
In preparation for adding separate sub-commands for Operator checks and Container checks, I've broken up the checks acoordingly.

### DIFF
--- a/certification/engine/engine.go
+++ b/certification/engine/engine.go
@@ -88,6 +88,21 @@ var nameToChecksMap = map[string]certification.Check{
 	validateOperatorBundle.Name():         validateOperatorBundle,
 }
 
+var containerPolicyChecks = map[string]certification.Check{
+	runAsNonRootCheck.Name():              runAsNonRootCheck,
+	underLayerMaxCheck.Name():             underLayerMaxCheck,
+	hasRequiredLabelCheck.Name():          hasRequiredLabelCheck,
+	basedOnUbiCheck.Name():                basedOnUbiCheck,
+	hasLicenseCheck.Name():                hasLicenseCheck,
+	hasMinimalVulnerabilitiesCheck.Name(): hasMinimalVulnerabilitiesCheck,
+	hasUniqueTagCheck.Name():              hasUniqueTagCheck,
+	hasNoProhibitedCheck.Name():           hasNoProhibitedCheck,
+}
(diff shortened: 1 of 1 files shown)


certification/engine/engine.go
```diff
@@ -98,3 +113,25 @@ func AllChecks() []string {
 	}
 	return all
 }
+
+func OperatorPolicy() []string {
```
komish, line 117: I was originally going to ask to replace the word Policy with Check, BUT - do we want to leave "Policy" as the term here, under the previously discussed concept that our _policy_ is a collection of checks?

certification/engine/engine.go
```diff
@@ -88,6 +88,21 @@ var nameToChecksMap = map[string]certification.Check{
...
+	hasMinimalVulnerabilitiesCheck.Name(): hasMinimalVulnerabilitiesCheck,
+	hasUniqueTagCheck.Name():              hasUniqueTagCheck,
+	hasNoProhibitedCheck.Name():           hasNoProhibitedCheck,
+}
+
+var operatorPolicyChecks = map[string]certification.Check{
+	validateOperatorBundle.Name(): validateOperatorBundle,
+}
+
 func AllChecks() []string {
```
komish, line 106: I'm assuming that functions like this one remain because the cobra command split hasn't happened yet. Do you intend on removing this code when the split happens?

If not, I'll assume that the `nameToChecksMap` (higher up in this file) _will_ go away regardless of whether we get rid of this function. With that in mind, do we want to go ahead and have you change this logic to sum return a sum of the values from `containerPolicyChecks` ard `operatorPolicyChecks` instead?

certification/engine/engine.go
tonytcampbell: Exactly.  Policy = a collection of checks.

certification/engine/engine.go
tonytcampbell: Yeah.  I was trying to make small changes but agreed.  In a future PR I was thinking about just creating a `func getChecks(policy string)` Where we could pass in the policy and it would just return the checks.

jomkz (approved):
/lgtm

komish (approved):
lgtm

PR #13, opened 2021-07-07
Remove logger as a parameter

bcrochet:
Originally, the logger was being passed around, and was a part of the
main interfaces. This removes that dependency, and allows code to just
import logrus as 'log' and be able to log without setup or passing the
logger around. This patch maintains backwards compatibility with the
functionality that existed for the logger.

### DIFF
--- a/certification/certification.go
+++ b/certification/certification.go
@@ -1,12 +1,10 @@
 package certification
 
-import "github.com/sirupsen/logrus"
-
 // Check as an interface containing all methods necessary
 // to use and identify a given check.
 type Check interface {
 	// Validate checks whether the asset enforces the check.
-	Validate(image string, logger *logrus.Logger) (result bool, err error)
+	Validate(image string) (result bool, err error)
 	// Name returns the name of the check.
 	Name() string
 	// Metadata returns the check's metadata.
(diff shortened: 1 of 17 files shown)


samira-barouti (approved):

komish:
We lose the ability to globally configure our logger in this PR, don't we?

rocrisp (approved):
Looks good!

komish:
Expanding a bit:

I'm 100% in for not having loggers passed into the function signature of the interfaces.

That said (and I might be reading into this incorrectly), I'm understanding that importing logrus allows us to call it without having it be a part of the signature, but each imported location also has its own configuration.

If that's not the case, and we only have to configure the logger in one place - then I'm good to go! Otherwise, I think there's value in being able to configure the logger globally. Maybe embedding the logger into our struct definitions all the way down is an option to remove the logger from the signature, but retain the single-point-of-configuration.

jomkz:
@komish Does [this](https://stackoverflow.com/a/30261304) answer the question that you have about global configuration? My understanding from reading that explanation is that when using `log` as imported, it uses the global configuration and it's only when you create a new package logger instance that the configuration is at the package level.

komish:
> @komish Does [this](https://stackoverflow.com/a/30261304) answer the question that you have about global configuration? My understanding from reading that explanation is that when using `log` as imported, it uses the global configuration and it's only when you create a new package logger instance that the configuration is at the package level.

Got it. Seems like it should work. I'm good to go, then. Thanks!

komish (approved):
/lgtm

Issue #15, opened 2021-07-07
The output file for preflight itself should be configurable (default: preflight.log)


komish:
We currently write a logfile to preflight.log as well as STDOUT. We should:

* Accept an environment variable/flag that indicates where we should write this file, and default to preflight.log if none is provided.
* Accept an environment variable/flag that indicates whether we should NOT also write to STDOUT/ERR (e.g. a --quiet flag).
There is also a current bug that should be fixed by this enhancement. Currently. all calls to preflight will write a log file on disk. This includes calls to -help or -version which we don't need.

This is because the initial implementation is called in an init() function[1] which is always executed. Instead, we want to explicitly initialize logging functionality once we've parsed the user provided configuration. 

[1]https://github.com/redhat-openshift-ecosystem/openshift-preflight/blob/ae9b2c7a5e78c092d574de7425392ce1a6faea89/cmd/log.go

komish:
This issue has been rescoped and this is no longer configurable for the moment. Closing this to preserve historical context, and re-opening a new issue for the init bug.

Issue #14, opened 2021-07-07
Consider writing check results to a file


komish:
As a user, I may prefer for the results of my preflight execution to be written to a file instead of to STDOUT. Note that this refers to the test output that is passed through the implemented formatters itself, and not the log output the that preflight provides to the user at execution time.

**Implementation Note:**

We write our formatted text in one place: https://github.com/redhat-openshift-ecosystem/openshift-preflight/blob/ae9b2c7a5e78c092d574de7425392ce1a6faea89/cmd/root.go#L56 

This should just be a matter of choosing where to write.

komish:
@tonytcampbell Should we go ahead and write to something like `results.xml` in addition to STDOUT?

PR #17, opened 2021-07-07
Implement configurable log writing for cli output

komish:
Fixes: #15 

## Summary
* By default, we write a log file to preflight-<timestamp>.log wherever you run `preflight`. Because this was done using the `init()` function in the `cmd` package, the tool would _always_ create this file, even in cases where the user only called `--help` or `--version`. This is a bug.

To fix this, I moved the logic out of the `init()` function which is called on package initialization, and instead moved it to an on-demand initializer function. We call this explicitly in `root.go`, and so we no longer have extra log file written to disk if the user doesn't actually intend to run the tool.

With that done, we control when we call this initializer, so I went ahead and added configuration allowing for the user to define exactly what log file they want to write. There are other configurables I could have worked in here ( think format, verbosity, etc), but I think we can do that in a later PR.

* The log file format now changes from `filename-<timestamp>.log` to `filename.log-timestamp`. I've updated the .gitignore to reflect this, but I've left the original format in place as well as a precaution. I've also added a larger catch-all, given that users can specify their log file name.

* I've introduced a new `PREFLIGHT_CLI_LOG_FILE ` environment variable in additional to the `--log-file` flag. Users can specify their output log file either way, with the traditional precedence enforced. Precedence is implemented using the same logic as already implemented.

* I updated the README to reflect the additional CLI flags. 

## Examples

By flag (highest precedence)
```
../preflight -c HasMinimalVulnerabilities  \
    registry.access.redhat.com/ubi8/ubi:latest \
    --log-file foo.log
```

By environment (next precedence)
```
PREFLIGHT_CLI_LOG_FILE=env.log ../preflight \
    -c HasMinimalVulnerabilities \
    registry.access.redhat.com/ubi8/ubi:latest 
```


cmd/root.go
```diff
@@ -34,6 +34,11 @@ var rootCmd = &cobra.Command{
 			Image:          containerImage,
 			EnabledChecks:  parseEnabledChecksValue(),
 			ResponseFormat: parseOutputFormat(),
+			CLILogFile:     parseCLILogFile(),
+		}
+
+		if err := initializeLoggerForConfig(cfg); err != nil {
```
tonytcampbell, line 40: nit - could we name this `initLogger` or even `initLog'?  Under the premise that the ForConfig suffix doesn't make it more readable.

cmd/root.go
```diff
@@ -77,6 +83,10 @@ func Execute() {
 		"output-format", "o", "", fmt.Sprintf(
 			"The format for the check test results.\n(Env) %s (Default) %s",
 			EnvOutputFormat, defaultOutputFormat))
+	rootCmd.Flags().StringVarP(&userCLILogFile,
+		"log-file", "l", "", fmt.Sprintf(
```
tonytcampbell, line 87: What if we didn't give the user a choice of where the log file goes?  If we make sure users are using our container we can be prescriptive and the log location and remove a flag making the codebase simpler.  I do see the need for this if the primary user uses the Go binary directly without the container.

cmd/log.go
```diff
@@ -4,21 +4,33 @@ import (
...
 )
 
-func init() {
+// initializeLoggerForConfig will configure the logger so that it matches the
+// user's specified configuration, such as configuring output files.
+func initializeLoggerForConfig(config runtime.Config) error {
 	time := time.Now().Unix()
-	logname := fmt.Sprintf("preflight-%d.log", time)
+	logname := strings.Join([]string{
+		config.CLILogFile, fmt.Sprint(time)}, "-")
```
tonytcampbell, line 20: What do you think about having just 1 log file that gets rewritten with each run of preflight?  I'm not sure users need all the log files.  I think they will just need the log file from the last run of the tool. They can always copy it somewhere else manually if they want to hold on to it.

bcrochet: That works for me. If someone wants to have timestamps, they can always pass a preflight-$(date).log as env/cli param.

cmd/root.go
bcrochet: As long as the default is sane, I'm good with having the option.

bcrochet:
Should we include a --log-level in this PR? Or future?

cmd/root.go
komish: @tonytcampbell Fair - I think my goal here with the naming was to make it clear that we were configuring our logger based on the user configuration. In other places, the `NewForConfig(runtime.Config)` semantic is used (e.g. [engine](https://github.com/redhat-openshift-ecosystem/openshift-preflight/blob/6a626e2306c2eb6d25600d57611dc333b425f2cd/certification/engine/engine.go#L37)).

I couldn't do that here because the logger configuration isn't in its own package, and putting in its own package when there is a minimal amount of action taking place seemed silly at the time.

Summarizing: I can change it, but it would differ a bit from the semantic. It doesn't match it exactly anyway! If you think moving it to its own package would make it more readable while maintaining the "<Action>ForConfig" semantic (something like `logger.ProvisionForConfig(runtime.Config)` or similar), I could do that as well.

Let me know!

cmd/root.go
komish: > If we make sure users are using our container we can be prescriptive

Isn't this true with or without the flag?

cmd/log.go
komish: I like this. I'll work on it tomorrow.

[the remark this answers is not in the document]
komish: > This this will end up as preflight.log-?

Indeed. Conscious choice. I also tried <timestamp>-<filename> and almost immediately hated that 😆 

> Do we want preflight-.log instead?

This is what we had before, but it's less reliable to do if we accept relatively arbitrary inputs for the filename from the user. I did the above as a result.

> Either way... I want to ask again up front if we can format the time to be '20210707135401' so that at least it is human-readable?
 
Absolutely. I erred on the side of incremental changes and didn't include it in this PR, despite it being effectively a one-line difference. I'll fix it and re-push.

cmd/log.go
```diff
@@ -4,21 +4,33 @@ import (
...
+
 	logFile, err := os.OpenFile(logname, os.O_CREATE|os.O_WRONLY|os.O_APPEND, 0700)
 	if err == nil {
+		// if we could open the file for writing, write to both stdout and the file
+		// otherwise, just use stdout
 		mw := io.MultiWriter(os.Stdout, logFile)
 		log.SetOutput(mw)
 	} else {
-		log.Info("Failed to log to file, using default stderr")
+		log.Info("Failed to log to file, using default stdout")
```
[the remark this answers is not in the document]
komish: We can. For clarity, this change only made a change to the text output itself. We were already logging to stdout from what I could tell.

Caveat, here - if we log command line output (read: everything EXCEPT the results) to stderr, would we be violating the proper use of stderr? They technically aren't error lines (or I guess they could be).

I have another issue open to address writing results to a separate location as well: https://github.com/redhat-openshift-ecosystem/openshift-preflight/issues/14 if that helps with the comfort level of writing both to stdout.

komish:
> Should we include a --log-level in this PR? Or future?

I'd say a future PR. I can draft the issue up right after this merges.

cmd/log.go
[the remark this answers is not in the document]
komish: Actually, if I'm no longer including a timestamp (and leaving that up to the user), then I think we scratch the need for the human-readable format.

cmd/root.go
komish: I've changed this to `initLogger` in the interim. Happy to move forward with that or change it to whatever works best!

cmd/log.go
komish: This is done - we no longer have timestamped log files, and instead just write wherever the user requests, or `preflight.log` if no user input was provided, truncating if the file already exists with data.

cmd/log.go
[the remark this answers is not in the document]
bcrochet: I don't think we would be violating it. It's pretty typical to separate log output to stderr (which might include actual errors) and the proper output to stdout. Think pipes... If we output json or xml, it would not be a stretch to send that output to a redirect or a pipe. If you put the log messages on stdout, you would no longer have the ability to send the output to jq for example.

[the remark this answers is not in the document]
komish: Understood - I'm with you now. This shoudl be fixed.

komish:
This has been rebased and now we're doing the same initialization in both the `check operator` and `check bundle` subcommands.

bcrochet (approved):

cmd/root.go
komish: @tonytcampbell per our discussion offline, we'll look to solidify a single output location for the log file. I can do it in a subsequent PR.

komish:
Force pushed to rebase, sorry @bcrochet

komish:
If #26 merges before this, I can rebase and remove the bits from this PR that are no longer in scope.

komish:
Closing this PR to redraft the message, change what's fixed, and rebase. I'll open a new PR with the bug fix for #15.

PR #26, opened 2021-07-08
Disconnect flags and rely on sane defaults

### LINKED ISSUE #24: Rely on sane defaults for initial release version of preflight tooling
We have a few configurables such as enabled checks and output formats currently implemented. For initial releases and integration with the pipelines as a whole, we should disable those configurations and look to enable them as we need them down the line.

komish:
This PR disconnects existing flags/configurables,~but leaves the ability to override things via environment variable (as an administrative configurable) intact~.

~Now that we have two subcommands with different checks, I've adjusted the function `parseEnabledChecksValue` that determines what tests should be executed such that it accepts a default value so that we can re-use it for both the operator and container tests.~

Lastly, we also set the default result format to junitxml.

FIxes #24 

~Existing functionality to determine precedence of configurables is still in place for future use cases, whenever they're needed.~


cmd/check_operator.go
```diff
@@ -13,7 +13,7 @@ import (
 
 var checkOperatorCmd = &cobra.Command{
 	Use:   "operator",
-	Short: "Run checks for an Operator",
+	Short: "Run checks for an operator",
```
tonytcampbell, line 16: Hehe....shouldn't "Operator" be upper case?  Operator is a proper noun no?

komish: Haha! I just wanted it to be consistent, and container is lower case.

That said... I Tend To Capitalize Way Too Much and so that's why it caught my eye 😆 😆

tonytcampbell (commented):
Can we remove the entire parse.go file?  If it turns out we need it later I think we can recreate it.  It's sole purpose was parsing the flags right?

cmd/parse.go
```diff
@@ -21,15 +27,19 @@ func getConfigStringValueFrom(flagValue, envValue, defaultValue string) string {
 	return defaultValue
 }
 
-func parseEnabledChecksValue() []string {
+// parseEnabledChecksValue returns the enabled checks requested by the user, or the defaultChecks
+// provided if the user has not provided any input.
+func parseEnabledChecksValue(defaultChecks []string) []string {
```
samira-barouti, line 32: I don't know if I am being too picky: I find it a bit confusing to pass around the default values as the arg. I believe we better define defaultChecks as a global var, and set it in the `RunE` function, and access this var here. Looking at the method name, I was expecting that the function is parsing the arg.

The other workaround is to pass both `userEnabledChecks` and `defaultChecks`.

cmd/check_operator.go
tonytcampbell: I think Operator is actually capitalized and container is lower case. I looked at how they do it on RedHat.com and that seems to be their implementation.

cmd/parse.go
tonytcampbell: And I'm voting for removing parse.go altogether if possible.

komish:
> Can we remove the entire parse.go file? If it turns out we need it later I think we can recreate it. It's sole purpose was parsing the flags right?

@tonytcampbell It parsed flags, environment, and defaults and returns the correct value based on precedence. We still have environment variables and defaults in scope. If you remove environment variables, then we don't need "defaults", right? If you want to go that route, we could remove parse.go.

cmd/parse.go
komish: @samira-barouti We have two subcommands that need to parse enabled checks, but they have different default values. The alternative here was to write two functions, one for each, but that seemed inefficient.

cmd/check_operator.go
komish: Fixed.

tonytcampbell:
@komish I think @bcrochet was advocating for still supporting Env variables.  I'd rather implement Envs when partners ask for it, but I can be swayed.  I'd also suggest that we look at [Viper](https://github.com/spf13/viper) if we want to handle Flags, Env Vars, and other Config.

tonytcampbell:
@komish I'll still lend my approval to the PR if we want to talk about this more and make changes in a future PR.

tonytcampbell (approved):
/lgtm

komish:
@tonytcampbell I'm still avoiding Viper given the minimal configuration we have/had - just seems like a large dependency. That said, I've gone ahead and pulled out the envs. We can revisit when requested.

cmd/parse.go
komish: Shoudl be sorted. `parse.go` is removed, which, I think, resolves this thread 👍

komish:
Apologies for the extra push - had to fix the commit message that I botched in the squash.

samira-barouti (approved):

PR #27, opened 2021-07-08
display any warnings from validation

### LINKED ISSUE #19: Need to catch Warnings in bundle validate output
https://github.com/redhat-openshift-ecosystem/openshift-preflight/blob/6a626e2306c2eb6d25600d57611dc333b425f2cd/certification/internal/shell/validate_operator_bundle.go#L23

The operator-sdk bundle validate command outputs a string when it finishes "All validation tests have completed successfully".  However, we should catch the Warnings too and pass those on to the user.

edreyer-1:
While running operator-sdk bundle validate, there could be warnings that would not cause the validation to fail. We want the user to still be able to see these warnings even if it doesn't cause the validation to fail.

fixes #19


certification/internal/shell/validate_operator_bundle.go
```diff
@@ -21,6 +22,11 @@ func (p ValidateOperatorBundlePolicy) Validate(bundle string) (bool, error) {
 	lines := strings.Split(string(stdouterr), "time=")
 
 	if strings.Contains(lines[len(lines)-1], "All validation tests have completed successfully") {
+		for _, line := range lines {
+			if strings.Contains(line, "level=warning") {
+				fmt.Println("time= ", line)
```
tonytcampbell, line 27: @edreyer-1 can we log.Warn() here instead of fmt.Println()?

komish: I think this is all we need to merge!

komish (approved):
/lgtm

PR #32, opened 2021-07-09
Explicitly initialize logging when running subcommands

### LINKED ISSUE #30: The preflight.log file is written even when simply asking for `--help`.
This is because the initial implementation is called in an init() function[1] which is always executed. Instead, we want to explicitly initialize logging functionality once we've parsed the user provided configuration.

Was previously a part of #15 and has been separated into its own issue due to descoping the rest of the work in that issue.

komish:
Fixes #30 

This PR replaces the initialization of logger using the package `init()` function and does so explicitly only when we run subcommands of interest (`check container`, `check operator`). This prevents us from writing empty log files no matter what command was executed.

See below example where two commands would have previously written a log file:

```
$ ../preflight --version; ../preflight check container --help; ls -lah 
preflight version 0.0.0 <commit: 38cde99aac08294794c56ef6a325dce72c49a265>
This command will run the Certification checks for a container image.

Usage:
  preflight check container <url to container image> [flags]

Flags:
  -h, --help   help for container
total 4.0K
drwxrwxr-x.  2 komish komish    6 Jul  9 07:56 .
drwx------. 13 komish komish 4.0K Jul  9 07:57 ..
```


bcrochet (approved):
/lgtm

cmd/check_operator.go
```diff
@@ -31,6 +31,11 @@ var checkOperatorCmd = &cobra.Command{
 			Image:          operatorImage,
 			EnabledChecks:  engine.OperatorPolicy(),
 			ResponseFormat: defaultOutputFormat,
+			LogFile:        defaultLogFile,
+		}
+
+		if err := initLogger(cfg); err != nil {
+			return fmt.Errorf("unable to initialize the logger for the configuration: %e", err)
```
samira-barouti, line 38: Do we want to create an error in the `errors.go`?

samira-barouti (approved):
Added a small comment, no road blocker though

cmd/check_operator.go
komish: Oh! I created one and then forgot to use it. 🤦 I'll fix that!

komish: @samira-barouti fixed

komish:
@samira-barouti @bcrochet I see the approval, but since I pushed a change I'll wait for an lgtm to come through (or you can just push the merge button). Just as a precaution.

komish:
Also resolved a conflict in certification/errors/errors.go as a result of recent merge.

samira-barouti (approved):

PR #45, opened 2021-07-09
Extract common test logic into helper functions

### LINKED ISSUE #44: Extract common unit test logic into helper functions
There is a lot of boilerplate in the tests for each of the checks. This can be extracted into helper functions located on the suite.

bcrochet:
The unit tests for each of the checks follow a basic pattern.

Set up the fake engine
Verify that is passes Validate
Modify the fake engine for a failure mode
Verify that it does not pass Validate
Verify that Podman errors are handled properly

This extracts the verifications into helper functions that can be reused
to cut down on the boilerplate.

Fixes #44

### DIFF
--- a/certification/internal/shell/base_on_ubi_test.go
+++ b/certification/internal/shell/base_on_ubi_test.go
@@ -2,9 +2,7 @@ package shell
 
 import (
 	. "github.com/onsi/ginkgo"
-	. "github.com/onsi/gomega"
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/cli"
-	log "github.com/sirupsen/logrus"
 )
 
 // podmanEngine is a package-level variable. In some tests, we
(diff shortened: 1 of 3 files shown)


komish:
This could just be me, so I'd definitely like to see another opinion on the subject, but this feels less readable to the developer. The developer will dig into this code in a situation where they run the tests and the tests fail even though they expect a success case as a result of the inputs/outputs of their work (i.e. "while debugging tests"). With these changes, I'd have to go hunt down the appropriate functions elsewhere, and though those functions as still using Gomega assertions, it feels a bit more disjointed when I only see the assertion separate from the context.

Key piece, again, is that this is just my opinion, and I'd be interested in other perspectives.

bcrochet:
> This could just be me, so I'd definitely like to see another opinion on the subject, but this feels less readable to the developer. The developer will dig into this code in a situation where they run the tests and the tests fail even though they expect a success case as a result of the inputs/outputs of their work (i.e. "while debugging tests"). With these changes, I'd have to go hunt down the appropriate functions elsewhere, and though those functions as still using Gomega assertions, it feels a bit more disjointed when I only see the assertion separate from the context.
> 
> Key piece, again, is that this is just my opinion, and I'd be interested in other perspectives.

I can see your perspective. It's might be more appropriate to extract any setup logic into shared code. I'm not sure I've seen an example of this type of refactor. 

My motivation here was to speed up the creation of unit tests, and to cut down on the amount of boilerplate.

I'm interested in having further discussion on it.

samira-barouti:
As Jose said, there is no right or wrong way to do it. I agree that it is not too readable. I wanted to suggest to instead refactor the test data generation into functions to simplify `BeforeEach`  block and keep the assertions as is. Giving a meaningful name to the data generation methods could help with the test readability as well.

bcrochet:
Will maybe revisit later. Marking as hold.

/hold

Issue #46, opened 2021-07-09
Add initial behavior testing


komish:
I think we can programmatically test behaviors of the application as a whole without necessarily calling out to the built binary by performing similar functions to what's done in the cobra `Run` commands.

## The Plan
* I'll find a "known good" and "known bad" container image and operator bundle
* build a `runtime.Config` using that image and our known defaults
* get the default engine
* run the checks
* parse the results

Similar to what we do here: https://github.com/redhat-openshift-ecosystem/openshift-preflight/blob/main/cmd/check_container.go#L24-L64

## The Test Case
The "known good" configuration should have no failed/errored entries in our results
The "known bad" configuration should have no successful/errored entries in our results

Down the road, we could expand this to be more granular, where we point to images that have known, specific failures and confirm that the results reflect that failure.

Logistically, I'd probably put this in a top-level `test` directory.

komish:
cc: @bcrochet

samira-barouti:
The UniqueTagCheck needs two good containers as well: one with only the `latest` tag, and one with `latest` AND other tags. So I believe we need at least 3 containers.

bcrochet:
So, this sounds to me light e2e/integration testing, as opposed to unit testing. So, for this, I would definitely concur with the top-level test directory, and it being a separate package.

I think I'd like to see automation around building the known good/known bad images, pushing them to a public (quay.io) repository, then running the full tests based on those built images. Or, have a periodic build action (Does GH Actions support periodics? I would hope so...) that would build the images and push them to the registry. Then, the tests would just pull those down from a known location.

Overall, I like the concept.

PR #48, opened 2021-07-09
Turn on github action using existing make targets

komish:
We have low test coverage for now, but there's an active effort working on addressing that. This turns on a GitHub Action that just runs `make build` and `make test` (i.e. what we're likely already doing in code reviews)

EDIT: full disclosure, I'm not sure of a way to test that make test make build functions in the GH action environment right now (it's not doing anything in my fork for some reason), so this may need some tweaking after merged. I can send a dummy PR to confirm.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -1,0 +1,22 @@
+name: Go
+
+on:
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Set up Go
+      uses: actions/setup-go@v2
+      with:
+        go-version: 1.16
+
+    - name: Build
+      run: make build
+
+    - name: Test
+      run: make test
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,7 @@ image-push:
 
 .PHONY: test
 test:
-	go test -v ./... 
+	go test -v $(go list ./... | grep -v e2e)
 
 .PHONY: vet
 vet:


.github/workflows/go.yml
```diff
@@ -0,0 +1,22 @@
+name: Go
+
+on:
+  pull_request:
+    branches: [ main ]
```
samira-barouti, line 5: can we turn this on for all branches?

komish: @samira-barouti I don't think we need it for the moment, as we have no other branches we maintain and no new branches we're anticipating. In addition, In practice, `main` is the only protected branch. When we add branches, we'll have to determine how we handle those semantics (this, and protection - assuming we can protect multiples).

bcrochet: Agreed. Also, it only applies to this particular repo. It would not, for instance, apply to a branch on your personal fork, if that's what you were thinking.

samira-barouti: Exactly! My intention was to turn this on for all branches to ensure these actions are running on the feature branches that the contributors push to the remote repo. 

Out of curiosity: is this doable at all with GH Actions, or we need a full-fledged CI like prow?

bcrochet: I think as long as the PR is targeting main on this repo, it will run.

bcrochet (approved):
For as I can tell, this looks good.

/lgtm

komish:
Added do not merge. In building behavioral tests for #46, I'm realizing that we may have an issue running this in GHA if we need root privileges for the vulnerability test.

bcrochet:
> Added do not merge. In building behavioral tests for #46, I'm realizing that we may have an issue running this in GHA if we need root privileges for the vulnerability test.

Should we limit (or change) make test to just unit tests? Create a 'test/unit' target in the makefile? And then 'test/e2e' for the other? Then limit make test/unit to specific dirs?

komish:
@bcrochet Yep, I think that's what I need here. At least, we need it until we finish plumbing the infrastructure necessary to run those tests that need it.

komish:
@bcrochet @samira-barouti this now changes the `make test` target to run tests not listed as part of the e2e package, which I'm submitting a PR for in relation to #46.

tonytcampbell (approved):
/lgtm

PR #50, opened 2021-07-09
add golang bindings for v2 implementation of preflight

OchiengEd:
This PR is a work in progress and aims to be used in v2 of preflight.

It checks for the existence of both podman and docker sockets and uses the available socket to interact with containers and images.

The current implementation only supports a handful of methods. More methods need to be implemented for instance to allow exporting container filesystems and running commands in a container

More design and work needs to be done to integrate with the existing code and to provide a mechanism to run operator-sdk / scorecard tests.

See example.go file in the bindings package for an idea how to interact with the tools.

Comments and contributions are welcome.

**TODO:**
- write unit tests

**EXAMPLE:**

```
import (
 	"fmt"
)

func TestListImages() {
	client, err := Client()
	if err != nil {
		fmt.Println("error:", err.Error())
	}
	images, err := client.ListImages()
	if err != nil {
		fmt.Println(err.Error())
	}

	fmt.Printf("%+v/n", images)
}
```


OchiengEd:
let me know if a rebase is needed

certification/internal/bindings/types.go
```diff
@@ -0,0 +1,54 @@
...
+	ID          string
+	Labels      map[string]string
+	RepoTags    []string
+	RepoDigests []string
+}
+
+// The ContainerTool interface provides a unified way to interact with
+// both the podman and docker methods below
+type ContainerTool interface {
+	// PullImage takes an image name and pulls the image to the local cache
```
bcrochet, line 42: Most of these methods should have a return type of a struct specific to that operation (PullImageReport, ListImageReport, etc.). It would "future proof" the return type a bit, so that as more structured data is returned, existing code wouldn't necessarily need to change.

certification/internal/bindings/podman.go
```diff
@@ -0,0 +1,110 @@
...
+	_, err := images.Pull(p.Context, nameOrID, &images.PullOptions{})
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (p *PodmanClient) SaveImage(nameOrID string) error {
+	outfile, err := os.Create("image.tar.gz")
```
bcrochet, line 21: Can we at least name this with the nameOrID?

bcrochet (changes requested):
Couple of comments to start.

certification/internal/bindings/podman.go
OchiengEd: What are you referring to as "this"? I have no reference point.

certification/internal/bindings/types.go
OchiengEd: roger that

certification/internal/bindings/podman.go
bcrochet: image.tar.gz is an extremely generic name. That filename should be created based on nameOrID.

{nameOrID}.tar.gz

OchiengEd:
@bcrochet recommendations done.

certification/internal/bindings/example.go
```diff
@@ -0,0 +1,18 @@
+package bindings
+
+// import (
+// 	"fmt"
+// )
+
+// func TestListImages() {
```
komish, line 7: Can you go ahead and remove this file, and instead just add the example to a comment on this PR?

certification/internal/bindings/helpers.go
```diff
@@ -0,0 +1,162 @@
...
+
+func parseTime(stringTime string) *time.Time {
+	t, err := time.Parse(time.RFC3339, stringTime)
+	if err != nil {
+		return nil
+	}
+	return &t
+}
+
+func buildInspectImageReport(imageData interface{}) *InspectImageReport {
```
komish, line 61: Add a comment describing the behavior

certification/internal/bindings/helpers.go
```diff
@@ -0,0 +1,162 @@
...
+		return dockerInspectImageReport(imageData.(types.ImageInspect))
+	case "*entities.ImageInspectReport": // podman client
+		return podmanInspectImageReport(imageData.(*entities.ImageInspectReport))
+	}
+
+	return nil
+}
+
+// TODO: convert over exposed ports: report.ContainerConfig.ExposedPorts
+func dockerInspectImageReport(report types.ImageInspect) *InspectImageReport {
```
komish, line 73: Add a comment describing the behavior

komish (changes requested):
A couple of changes in-line.

This PR gets us closer to solving the problem of supporting both docker and podman environments. It does, however, require that we rewrite some of our checks in order to make use of it.

I also think we'll have to introduce a few client methods that we're needing for bundle checks that are not represented here.

Finally, this handles containerCLI calls, but doesn't handle skopeo/operator-sdk based calls. We'll need to add those for the checks that currently utilize them in a future PR.

I approve merging based on the fact that this:
1) doesn't impact the existing code execution
2) moves us closer to supporting more developer workstations

From that perspective, we'll need to create additional issues that explores utilizing these bindings as a replacement for our current container CLI invocations, and writing the tests to support these new bindings.

@OchiengEd if you can create these issues, that'd be great. Otherwise, I'll create them after this gets merged.

certification/internal/bindings/helpers.go
komish: The reason I ask for this is because the input and output types appear similar in their description (the parameter name, and the return type), so the comment would help clarify to us the exact reason for this function existing without having to understand the code's behavior. This also applies to the next location.

komish:
@OchiengEd I also see a test failing for what appears to be a missing header file for a dependency? Not sure if we need to tidy up the gomod or adjust the test environment.

certification/internal/bindings/helpers.go
OchiengEd: this function has been removed.

certification/internal/bindings/example.go
OchiengEd: this has been done. Been added to the body of the PR.

certification/internal/bindings/helpers.go
OchiengEd: done.

OchiengEd:
> @OchiengEd I also see a test failing for what appears to be a missing header file for a dependency? Not sure if we need to tidy up the gomod or adjust the test environment.

a rebase was needed to fix this.

bcrochet (commented):
A quick comment. If you don't agree, I'm fine with it.

bcrochet (approved):

certification/internal/bindings/docker.go
```diff
@@ -0,0 +1,148 @@
...
+
+	return &PullImageReport{
+		Output: buffer.String(),
+		Tool:   reflect.TypeOf(d).String(),
+	}, nil
+}
+
+// TODO: allow user to change directory to store archives via env variable.
+// Also, add "/tmp" as default image path
+func (d *DockerClient) SaveImage(nameOrID string) (*SaveImageReport, error) {
```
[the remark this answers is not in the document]
OchiengEd: I think this approach is better as opposed to every author of a test using this function choosing a different location to save the archive. With that also comes the ease to override.

However, if we determine there is a good reason to pass the parameter later, I would be open to that. This approach ensures all our tarballs are in the same location.

komish:
I'm looking into the failing tests. I think it's a dependency problem in the environment used for GH Actions. I'll re-review and approve once I've sorted that out.

komish:
I believe the fix in PR https://github.com/redhat-openshift-ecosystem/openshift-preflight/pull/135 should resolve the failing tests here. After that merges, we should see passing tests here.

Issue #49, opened 2021-07-09
Reduce elevated privileges required to run preflight successfully when testing against a container image


komish:
When testing against a container image, the `HasMinimalVulnerabilitiesCheck` check makes a call out to oscap-podman, and it requires the ability to run a container as root. This test will fail if the runtime user cannot run a pod as root from what I can see.

We see this error if we run without elevated privileges

```
time="2021-07-09T13:36:57-05:00" level=error msg="unable to execute oscap-podman on the image: This script cannot run in rootless mode.\nvuln.html"
```

We're currently unable to control what checks are executed when calling preflight, which means we have to run the entire policy called with `sudo`. 

We need to identify a way to satisfy the HasMinimalVulnerabilitiesCheck's need for elevated privileges while still maintaining the ability to run the rest of the checks without elevated privileges.

PR #52, opened 2021-07-09
Add initial e2e/behavioral tests against test assets

### LINKED ISSUE #46: Add initial behavior testing

komish:
This PR adds an `e2e` package that contains 3 simple tests.
* Run `check operator` against a generated test asset that should pass
* Run `check operator` against a generated test asset that should fail
* Run `check container` against a generated test asset that should pass

Pass/Fail here is defined as "the currently implemented tests", which means that this test will automatically fail if a new test is added and the assets are not modified and uploaded.

Checking if a container fails is currently disabled to reduce the likelihood of flakiness. Apparently locating a container image that both has vulnerabilities AND can fail all the other tests is difficult. Taking any recommendations there.

Right now, this PR leaves us with an action item of identifying a common namespace on a container registry where everyone in the team can upload the test assets. Alternatively, we could wire up automation to rebuild them periodically (on PR, daily, etc).

Lastly, there's now a Makefile target `test-e2e` that will run these tests. Note that passing the vulnerability test requires root privileges. I've opened issue #49 with regards to the issues I see around the required elevated permissions.

closes #46


test/e2e/e2e_test.go
```diff
@@ -0,0 +1,107 @@
...
+			It("should pass all checks", func() {
+				Expect(len(results.Passed)).To(Equal(len(cfg.EnabledChecks)))
+			})
+		})
+
+		// check temporarily disabled: currently unable to find a container image to use
+		// for the container-fails.Dockerfile that would fail both the HasMinimalVulnerabilitiesCheck
+		// check in addition to all other container checks.
+
+		// Context("with a known-bad image", func() {
```
bcrochet, line 89: If you make this XContext or PContext, it will not run, without having to leave commented code in.

See https://onsi.github.io/ginkgo/#pending-specs

test/e2e/e2e_test.go
```diff
@@ -0,0 +1,107 @@
...
+			})
+		})
+	})
+
+	Describe("When enforcing container policy", func() {
+
+		var (
+			// TODO: replace this with a project-accessible namespace once identified.
+			goodImage = "quay.io/komish/preflight-test-container-passes:latest"
+			// badImage  = "quay.io/komish/preflight-test-container-fails:latest"
```
bcrochet, line 64: If you do the below about XContext, then you would not have to comment this due to non-usage.

bcrochet (changes requested):
I would also like to see some Makefile entries to both building and pushing the test images.

test/e2e/e2e_test.go
komish: I'll give this a shot.

test/e2e/e2e_test.go
komish: > I would also like to see some Makefile entries to both building and pushing the test images.

I would have built this here, but we dont' have a quay namespace built for these assets, and currently the assets reside in my namespace. I've opened up a subsequent issue for this here: https://github.com/redhat-openshift-ecosystem/openshift-preflight/issues/59

test/containerfiles/bundle-passes copy.Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+LABEL operators.operatorframework.io.metrics.project_layout=go.kubebuilder.io/v3
+
+# Labels for testing.
+LABEL operators.operatorframework.io.test.mediatype.v1=scorecard+v1
+LABEL operators.operatorframework.io.test.config.v1=tests/scorecard/
+
+# Copy files to locations specified by labels.
+COPY successful-bundle-assets/manifests /manifests/
+COPY successful-bundle-assets/metadata /metadata/
+COPY successful-bundle-assets/tests/scorecard /tests/scorecard/
```
bcrochet, line 20: I don't see a reference to /licenses?

bcrochet (changes requested):
Example-license.txt does not seem to be referenced anywhere.

test/e2e/e2e_test.go
komish: This worked and is pushed!

test/containerfiles/bundle-passes copy.Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM scratch
```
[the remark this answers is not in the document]
komish: Fixed! Thanks for the catch

test/containerfiles/bundle-passes copy.Dockerfile
komish: AFAIK, Licenses are required for container checks and not for bundles. The container-passes.Dockerfile puts this in place.

komish:
> Example-license.txt does not seem to be referenced anywhere.

@bcrochet see https://github.com/redhat-openshift-ecosystem/openshift-preflight/blob/b784f7d7abb71124b956c852a1036e6a1e107d21/test/containerfiles/container-passes.Dockerfile#L5

test/containerfiles/bundle-passes copy.Dockerfile
bcrochet: Cool. WFM.

test/e2e/e2e_test.go
bcrochet: Ok. Will add more in the issue.

bcrochet (commented):
Needs a rebase, but otherwise lgtm.

komish:
> Needs a rebase, but otherwise lgtm.

Done!

bcrochet (approved):

bcrochet (approved):
/lgtm

test/e2e/e2e_test.go
```diff
@@ -0,0 +1,106 @@
...
+			}
+
+			engine, err := engine.NewForConfig(cfg)
+			Expect(err).To(BeNil())
+
+			engine.ExecuteChecks()
+			results := engine.Results()
+
+			It("should pass all checks", func() {
+				Expect(len(results.Passed)).To(Equal(len(cfg.EnabledChecks)))
```
samira-barouti, line 36: for the sake of assurance, could you please assert that the number of failed/errored tests are 0?

test/e2e/e2e_test.go
```diff
@@ -0,0 +1,106 @@
...
+
+			engine, err := engine.NewForConfig(cfg)
+			Expect(err).To(BeNil())
+
+			engine.ExecuteChecks()
+			results := engine.Results()
+
+			// TODO: Replace this check so that you test for individual check failures
+			It("should not pass any checks", func() {
+				Expect(len(results.Passed)).To(BeZero())
```
samira-barouti, line 54: Probably ensuring that the length of passed tests are less than number of enabled checks is a bit more generic?

komish: What would be the goal here?

test/e2e/e2e_test.go
komish: This should be resolved!

test/e2e/e2e_test.go
samira-barouti: The goal is to have a more generic assertion; for example if we add a test image in the future where only one check fails and the remaining 10 checks pass, this test still fails.

komish: Ah, understod. I would expect this test to always fail. When we add images where a single check might fail, then I would expect a corresponding `Context` block, and a unique `It` block for that case. I believe this check is correct as written until we start building images that check for specific failures.

samira-barouti: Fair enough :)

PR #55, opened 2021-07-10
Adding viper for config mgmt and refactoring logging config

### LINKED ISSUE #51: preflight.log is created with an executable permission
The default log file is created with 0700 permissions and probably just needs 0600 (RW-) https://github.com/redhat-openshift-ecosystem/openshift-preflight/blob/main/cmd/log.go#L14

### LINKED ISSUE #54: Introduce viper for configuration and refactor logging init
Introduce viper to handle our configuration  and refactor the logging configuration. 

Modified cmd/root.go to initialize viper
Supports Env Vars
```
export PFLT_LOGFILE="foo.log"
export PFLT_LOGLEVEL="debug"
```

Supports an optional config.yaml file in the same directory as the binary
```
---
 logfile: "foo.log"
 loglevel: "Trace"
```

tonytcampbell:
This adds viper which fixes #54 and refactors our logging config which fixes #51

Modified cmd/root.go to initialize viper
Supports Env Vars
```
export PFLT_LOGFILE="foo.log"
export PFLT_LOGLEVEL="debug"
```

Supports an optional config.yaml file in the same directory as the binary
```
---
 logfile: "foo.log"
 loglevel: "Trace"
```


cmd/root.go
```diff
@@ -13,5 +19,47 @@ var rootCmd = &cobra.Command{
...
+	viper.AddConfigPath(".")
+	if err := viper.ReadInConfig(); err != nil {
+		if _, ok := err.(viper.ConfigFileNotFoundError); ok {
+			log.Info("Config file not found.  Proceeding without it.")
+		}
+	}
+
+	// Set up logging config defaults
+	viper.SetDefault("logfile", "preflight.log")
+	viper.SetDefault("loglevel", "trace")
```
bcrochet, line 49: I feel like the default log level should actually be something much lower, like warning. Trace is the most verbose. I feel like only a developer would need trace.

bcrochet (changes requested):
Question about log level default

bcrochet (commented):
I'd be fine with merging as is, and addressing log level default in a separate PR.

bcrochet (approved):

bcrochet (approved):
Address default log level in future PR.

cmd/root.go
tonytcampbell: Agreed.  Changed the default to warn.

cmd/constants.go
```diff
@@ -2,5 +2,4 @@ package cmd
 
 var (
 	defaultOutputFormat = "junitxml"
-	defaultLogFile      = "preflight.log"
```
komish, line 5: _see review-level comment_

This default should probably remain to remove hard-coded values elsewhere.

cmd/check_operator.go
```diff
@@ -31,11 +31,6 @@ var checkOperatorCmd = &cobra.Command{
 			Image:          operatorImage,
 			EnabledChecks:  engine.OperatorPolicy(),
 			ResponseFormat: defaultOutputFormat,
-			LogFile:        defaultLogFile,
```
komish, line 34: _see review-level comment_

cmd/check_container.go
```diff
@@ -31,11 +31,6 @@ var checkContainerCmd = &cobra.Command{
 			Image:          containerImage,
 			EnabledChecks:  engine.ContainerPolicy(),
 			ResponseFormat: defaultOutputFormat,
-			LogFile:        defaultLogFile,
```
komish, line 34: _see review-level comment_

cmd/root.go
```diff
@@ -13,5 +19,47 @@ var rootCmd = &cobra.Command{
...
+	viper.AddConfigPath(".")
+	if err := viper.ReadInConfig(); err != nil {
+		if _, ok := err.(viper.ConfigFileNotFoundError); ok {
+			log.Info("Config file not found.  Proceeding without it.")
+		}
+	}
+
+	// Set up logging config defaults
+	viper.SetDefault("logfile", "preflight.log")
+	viper.SetDefault("loglevel", "warn")
```
komish, line 49: _see review-level comment_

This also contains a hard-coded default `preflight.log` that probably should have made reference to our globally defined constant. Same for the log level, but we didn't have that configurable defined at all previously.

cmd/root.go
```diff
@@ -13,5 +19,47 @@ var rootCmd = &cobra.Command{
...
+}
+
+func initConfig() {
+	// set up ENV var support
+	viper.SetEnvPrefix("pflt")
+	viper.AutomaticEnv()
+
+	// set up optional config file support
+	viper.SetConfigName("config")
+	viper.SetConfigType("yaml")
```
komish, line 39: 1. Do we have a use case that demands a configuration file?
2. If not, can we proceed without enabling config-file use cases? cc @bcrochet as I know you mentioned that a set-and-forget approach might be valuable
3. Do we prefer a configuration file to CLI flags?

komish (commented):
This PR deviates quite a bit from how we were determining user configuration at runtime.

In fairness - we're not currently parsing user configuration at runtime because we wanted to be more prescriptive (reference #26). However, all of our _possible_ configuration is/was stored in a runtime.Config that we then feed to the various constructs (e.g. Engine) to determine "how we should run". The various components may not care about all configuration items, but nevertheless, they can if they want to.

This PR pulls the log file out of the runtime.Config while making it configurable in another way. For the most part, this seems fine, but there's a precedence being set here that future "configurables" might live in either of two places.

If we're going to use Viper, does it make more sense to have it render out the final value of the configurations (i.e. values after precedence takes place), use that to inform runtime.Config, and then move forward in the logic from there as we did previously? Alternatively, are we considering moving away from runtime.Config, or comfortable with the multiple places where configuration might be defined?

For what it's worth, I do see that the former option makes running initConfig in the `init()` more complicated, but the previous implementation handles this without issue.

tonytcampbell:
I'm an advocate for using Viper as it is a well tested broadly used solution for config. I'm of the opinion that it should be our default choice for config unless there is a use case where it doesn't work. I think we need a good reason to roll our own solution. 

Viper handles precedence, config files, env vars and more. It's a known lib that would be familiar to new devs joining the project. 

I think not needing to pass the config around makes the code simpler and easier to develop but that's just my perspective. 

I may be anchored on Viper so I'm open to being overruled.

samira-barouti (approved):

cmd/root.go
komish: Utilize variables defined in constants.go for the default value of loglevel and logfile so we don't hard-code it here.

komish (changes requested):
Just 2 changes to maintain our default configuration values in a single place, then we can merge.

EDIT: The GitHub UI isn't showing comments, but I'm asking for the constants.go file to maintain our default values for the logfile and loglevel parameters, and then use those in your Viper `SetDefault` calls.

I.e. 
- leave `defaultOutputFile` set to `preflight.log`
- create `defaultLogLevel` with value of `warn`
- Use those in your `SetDefault` calls for the corresponding flags.

komish (approved):
I went ahead and rebased to resolve conflicts since Tony is OOO. Also added a subsequent commit for a test that was added in the rebase which was broken. 

I'll address the changed requests in a future commit.

Thanks @tonytcampbell!

PR #56, opened 2021-07-10
Make tests run

tonytcampbell:
While the GH check is working nicely, the make test command isn't finding any tests.  I set this back to the value it was before  PR #48. @komish you will likely want to take another look at `go test -v $(go list ./... | grep -v e2e)`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,8 @@ image-push:
 
 .PHONY: test
 test:
-	go test -v $(go list ./... | grep -v e2e)
+	# go test -v $(go list ./... | grep -v e2e)
+	go test -v  ./...
 
 .PHONY: vet
 vet:


Makefile
```diff
@@ -24,7 +24,8 @@ image-push:
 
 .PHONY: test
 test:
-	go test -v $(go list ./... | grep -v e2e)
+	# go test -v $(go list ./... | grep -v e2e)
+	go test -v  ./...
```
bcrochet, line 28: If I'm not mistaken, I think we only have tests in certification atm. Can we just make this ./certification/...?

bcrochet: This also fixes the problem:

	go test -v $(shell go list ./... | grep -v e2e)

bcrochet (changes requested):
See comment on possible fix.

komish:
@bcrochet see #57 which should fix this behavior without running e2e tests (if/when they merge)

bcrochet:
Superseded by #57

PR #74, opened 2021-07-12
Olmsuitetest

rocrisp:
Add OlmSuiteTest and fix error trapping when k8s cluster is not configured.


bcrochet:
/hold
Revisit when #81 lands

rocrisp:
Brad refactored this so I see no reason not to close this PR.

Issue #77, opened 2021-07-13
Add [make image-build] and [make image-push] into Workflow


itroyano:
[proposed]  create them as a new yaml e.g. build.yml,   to separate build steps from go/unit steps.

PR #79, opened 2021-07-13
Adding an image build event for pushes to main and for releases

itroyano:
Added a Github Action for building and pushing Preflight images.
Trigger is for pushes to main ("snapshot") and release variant builds.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,39 @@
+name: Build
+
+on:
+  push:
+    branches: [ main ]
+  release:
+    types: [ published, created, edited ]
+
+jobs:
+  build:
+    name: Build and push image
+    runs-on: ubuntu-latest
+    env:
+      IMAGE_REGISTRY: quay.io/opdev
+      RELEASE_TAG: "0.0.0"
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Build Image
+      id: build-image
+      uses: redhat-actions/buildah-build@v2
+      with:
+        image: ${{ env.IMAGE_REGISTRY }}/preflight
+        tags: ${{ github.sha }} ${{ env.RELEASE_TAG }}
+        dockerfiles: |
+          ./Dockerfile
+
+    - name: Push Image
+      id: push-image
+      uses: redhat-actions/push-to-registry@v2
+      with:
+        image: ${{ steps.build-image.outputs.image }}
+        tags: ${{ steps.build-image.outputs.tags }}
+        registry: ${{ env.IMAGE_REGISTRY }}
+        username: ${{ secrets.REGISTRY_USER }}
+        password: ${{ secrets.REGISTRY_PASSWORD }}
+
+    - name: Print image url
+      run: echo "Image pushed to ${{ steps.push-to-quay.outputs.registry-paths }}"


itroyano:
#77

.github/workflows/build.yml
```diff
@@ -0,0 +1,19 @@
...
+
+on:
+  push:
+    branches: [ main ]
+  release:
+    types: [published, created, edited]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
```
itroyano, line 11: needs some more research here...  I'm not sure "ubuntu-latest" contains the podman binary for running the make image-* targets.
and I don't see an "actions/podman" or similar available.
@jomkz @bcrochet please hold on. don't review yet.

itroyano (commented):
do-not-merge/hold

.github/workflows/build.yml
itroyano: @komish maybe you can help?
I need either a Runner for "runs-on" which contains a podman binary,  or the podman variant of docker/build-push-action@v2

itroyano: Thanks @komish   looks like the ubuntu runner has both docker and podman 
https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md

.github/workflows/build.yml
```diff
@@ -0,0 +1,19 @@
...
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Image Build
+      run: make image-build
+
+    - name: Image Push
+      run: make image-push
```
bcrochet, line 19: Nit: I wouldn't hate a newline here.

bcrochet (commented):
Small nit, otherwise looks good.

.github/workflows/build.yml
itroyano: fixed nit

komish, line 19: I don't think these will work as written, right? We would need to login to Quay.io for this to work. Does it make more sense to use the push-to-registry? action? https://github.com/marketplace/actions/push-to-registry

.github/workflows/build.yml
```diff
@@ -0,0 +1,19 @@
+name: Build
+
+on:
+  push:
+    branches: [ main ]
```
komish, line 5: I'm not sure if we want to be doing this on push, or need to be doing this only on release. If we do it on push, I think we need to be managing our tag more closely.

komish (changes requested):

.github/workflows/build.yml
itroyano: makes sense.

one thing I'm missing is
username: ${{ secrets.OPDEV_REGISTRY_USER }} 
password: ${{ secrets.OPDEV_REGISTRY_PASSWORD }}

can you or @bcrochet  add these credentials ?   looks like I don't have permission.

.github/workflows/build.yml
itroyano: I'm planning to use
`tags: ${{ github.sha }}` at first, to mimic the  `VERSION=$(shell git rev-parse HEAD)` logic currently present in the Makefile.

Long term what I want to be doing is, in case this is a release build take the semver of the release as the image tag.
pending a bit more discussion with @jomkz about how a release looks like (I'm for git tag).

.github/workflows/build.yml
itroyano: changes made

komish: Doesn't look like I can add secrets to this repository, so it might have to be @jomkz or @tonytcampbell.

komish (approved):
Aside from the confirmation that the necessary secrets have been provisioned on this repository, this looks good. Thanks @itroyano 

/lgtm

.github/workflows/build.yml
jomkz: I have added a new robot account to the Quay [repo](https://quay.io/organization/opdev?tab=robots). Can you see it now @itroyno?

.github/workflows/build.yml
jomkz: I am for git tag as well. However, I'm unsure if there is a reason for us to assign a semver for official relaseas?

jomkz (approved):
This looks like a good start, pending getting the robot account credentials sorted out.

PR #82, opened 2021-07-13
Add log level default definition and use it in logging init

komish:
Quick fix up after #55 - this PR just adds the log level to the default configurations and uses those in our logg initialization logic.


samira-barouti (approved):

bcrochet (approved):
/lgtm

PR #85, opened 2021-07-13
Write results to disk and to stdout

### LINKED ISSUE #14: Consider writing check results to a file

komish:
Fixes #14 

## Summary

This PR configures preflight to write the check Results to a file as well as stdout using a MultiWriter.

This PR recognizes that different formatter types might have different file extensions, so this extends to ResponseFormatter interface to allow for the definition of the desired file extension [3ad9a1a22c757491d93d20dc345e70424732df4c](https://github.com/redhat-openshift-ecosystem/openshift-preflight/commit/3ad9a1a22c757491d93d20dc345e70424732df4c).

The writers then take this into account to write the file. [48140895b3ae9d6fa17a9da0f9ddcddd2082376b](https://github.com/redhat-openshift-ecosystem/openshift-preflight/commit/48140895b3ae9d6fa17a9da0f9ddcddd2082376b)

## Highlights
- Removed AllFormats() function as it was not in use.
- Made the formatters_test suite and internal test. This was previously an external test suite due to a dependency cycle with the cmd package which has now been removed. An equivalent test should live in the `cmd` package.
- Introduced a new interface method ResponseFormatter.FileExtension() which returns a string containing the file extension someone might use when writing the data from that formatter to disk.
- Expended the genericFormatter to continue implementing the ResponseFormatter interface
- Reduced the public surface of the genericFormatter
- Added comments to the interface definition.
- Added a function that takes the extension defined by the formatter and concatenates it with our expected results file.
- Modified the `check container` and `check operator` subcommands to write to a default file.

### DIFF
--- a/cmd/check_container.go
+++ b/cmd/check_container.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"fmt"
+	"io"
 	"os"
 
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/engine"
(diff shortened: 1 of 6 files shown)


samira-barouti (approved):

cmd/check_container.go
```diff
@@ -43,16 +44,46 @@ var checkContainerCmd = &cobra.Command{
...
 			return err
 		}
 
-		fmt.Fprint(os.Stdout, string(formattedResults))
+		fmt.Fprint(resultsOutputTarget, string(formattedResults))
+		if err := resultsFile.Close(); err != nil {
+			return err
+		}
+
+		// engine.ExecuteChecks()
```
bcrochet, line 77: Big block of commented code.

bcrochet (changes requested):

cmd/check_container.go
komish: 🤦 I'll fix and push tomorrow. Thanks for the catch!

komish: @bcrochet this should be fixed.

bcrochet (approved):
/lgtm

PR #89, opened 2021-07-14
Fixing a name bug in push-image step

itroyano:
Combining image name "quay.io/opdev/preflight" and registry "quay.io/opdev" to form registry path "quay.io/opdev/quay.io/opdev/preflight"
Warning: "quay.io/opdev/quay.io/opdev/preflight" does not seem to be a valid registry path. The registry path should not contain more than 2 slashes. Refer to the Inputs section of the readme for naming image and registry.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,11 +29,11 @@ jobs:
       id: push-image
       uses: redhat-actions/push-to-registry@v2
       with:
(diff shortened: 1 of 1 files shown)


jomkz (approved):
/lgtm

komish (approved):
/lgtm

PR #90, opened 2021-07-14
Add trace and error log lines to osdk Scorecard func

komish:
I was getting a failure early on in the OperatorSDK Scorecard-related that was not providing any logs. I added a log line that should catch it, as well as a trace line that shows what cmd invocation would get executed.


bcrochet (approved):
/lgtm

samira-barouti (approved):

PR #92, opened 2021-07-14
workaround for operator-sdk returning non-zero even when functional

komish:
OperatorSDK scorecard checks will fall into the `Errors` check list if they parse a bundle and the bundle does not pass a scorecard test. This is because `operator-sdk scorecard` returns a non-zero exit code in those cases.

The non-zero exit code is always 1, regardless of whether `operator-sdk scorecard` encountered a fatal error preventing it from testing the assets, or the assets failed a check.

The only workaround I've found so far is to check the length of `stderr` and if it's empty, we assume the command completed without a problem and take `stdout` as the scorecard results.

I also added a large comment explaining the situation. I've reached out to OperatorSDK to see if they can implement a different return code in cases where the test has failed vs. operator-sdk itself has failed so that users can distinguish these cases.

### DIFF
--- a/certification/internal/shell/operatorsdk.go
+++ b/certification/internal/shell/operatorsdk.go
@@ -42,8 +42,18 @@ func (o OperatorSdkCLIEngine) Scorecard(image string, opts cli.OperatorSdkScorec
 	cmd.Stderr = &stderr
 	err = cmd.Run()
 	if err != nil {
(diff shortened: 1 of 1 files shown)


bcrochet (approved):
/lgtm

samira-barouti (approved):

PR #95, opened 2021-07-14
Move user-facing results definitions to formatters

komish:
This moves our user-facing results definition over the formatters, and adjusts the way it renders a little bit. This also adds the ElapsedTime into the output, as well as the name (both were previously hidden).

I didn't want to spend too much time tweaking this because the schema is still being defined, but something a little better than what we had previously would help while we're working on the new checks.

```
{
    "image": "quay.io/komish/preflight-test-bundle-passes:latest",
    "validation_lib_version": {
        "version": "0.0.0",
        "commit": "51d7b962e776eb142d8700957ec3f24fe22c73b2"
    },
    "results": {
        "Passed": [
            {
                "name": "ValidateOperatorBundle",
                "elapsed_time": "4.721589672s",
                "description": "Validating Bundle image"
            },
            {
                "name": "ScorecardBasicSpecCheck",
                "elapsed_time": "7.720154257s",
                "description": "Check to make sure that all CRs have a spec block."
            }
        ],
        "Failed": [
            {
                "name": "ScorecardOlmSuiteCheck",
                "elapsed_time": "8.858909391s",
                "description": "OLM Test Suite Check",
                "help": "Operator-sdk scorecard OLM Test Suite. One or more checks failed.",
                "suggestion": "See scorecard output for details, artifacts/operator_bundle_scorecard_OlmSuiteCheck.json"
            }
        ],
        "Errors": []
    }
}
```

### DIFF
--- a/certification/formatters/util.go
+++ b/certification/formatters/util.go
@@ -1,43 +1,57 @@
 package formatters
 
 import (
-	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification"
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/runtime"
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/version"
 )
 
 // getResponse will extract the runtime's results and format it to fit the
 // UserResponse definition in a way that can then be formatted.
-func getResponse(r runtime.Results) runtime.UserResponse {
-	passedChecks := make([]certification.Metadata, len(r.Passed))
-	failedChecks := make([]certification.CheckInfo, len(r.Failed))
-	erroredChecks := make([]certification.HelpText, len(r.Errors))
+func getResponse(r runtime.Results) UserResponse {
+	passedChecks := make([]checkExecutionInfo, len(r.Passed))
+	failedChecks := make([]checkExecutionInfo, len(r.Failed))
+	erroredChecks := make([]checkExecutionInfo, len(r.Errors))
 
 	if len(r.Passed) > 0 {
 		for i, check := range r.Passed {
-			passedChecks[i] = check.Metadata()
+			passedChecks[i] = checkExecutionInfo{
+				Name:        check.Name(),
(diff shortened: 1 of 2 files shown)


certification/formatters/util.go
```diff
@@ -1,43 +1,55 @@
...
 	}
 
 	if len(r.Failed) > 0 {
 		for i, check := range r.Failed {
-			failedChecks[i] = certification.CheckInfo{
-				Metadata: check.Metadata(),
-				HelpText: check.Help(),
+			failedChecks[i] = checkExecutionInfo{
+				Name:        check.Name(),
+				ElapsedTime: check.ElapsedTime.String(),
```
samira-barouti, line 29: I suggest we limit the number of digits after the decimal point to two and round it to the nearest number.

samira-barouti (approved):
A suggestion; otherwise it is good to be merged.

certification/formatters/util.go
komish: @samira-barouti I 100% agree, but I'm going to leave it as is for now and we can sort this out when we have our schema definition sorted. The reason here being that the Round and Truncate methods on time.Duration don't seem to do what I want when I read through their description, which means I'd have to convert to common unit (ms) and type (float64), round/truncate mathematically, and then reconvert back to the original unit. For now I'm thinking it's not worth the effort, but I definitely agree, and actually would prefer to have the same unit across the board in the final results anyway. 

We can definitely revist in the future.

samira-barouti: Fair enough :) Thanks for your insight!

komish:
I've force pushed two things:
- Added KnowledgeBaseURL and CheckURL which were in the previous version of this but were removed in this PR (so I put them back)
- Added struct tags for the checks to force the JSON representation to have a lower cased value for the `resultsText.Passed/Failed/Errors`

bcrochet (approved):
/lgtm

PR #101, opened 2021-07-15
Write distinct message for error

### LINKED ISSUE #12: Print distinct message for check errors
Right now if a check errors out  we print the help text but we should print a different message to distinguish between a Check Failure and an error.  We can see say something as simple as, "This check encountered an error.  Please review the logs for more information."

edreyer-1:
Instead of the help text being displayed when there is an error with a check we want to display a message that directs the user to look at the logs to figure out the issue. We want there to be a distinction between a check failure and error.

Fixes #12

### DIFF
--- a/certification/internal/shell/has_prohibited_packages.go
+++ b/certification/internal/shell/has_prohibited_packages.go
@@ -62,7 +62,7 @@ func (p *HasNoProhibitedPackagesCheck) Name() string {
 }
 func (p *HasNoProhibitedPackagesCheck) Metadata() certification.Metadata {
 	return certification.Metadata{
-		Description:      "Checks to ensure that the image in use does not contain prohibited packages.",
+		Description:      "Checks to ensure that the image in use does not include prohibited packages, such as Red Hat Enterprise Linux (RHEL) kernel packages.",
 		Level:            "best",
 		KnowledgeBaseURL: "https://connect.redhat.com/zones/containers/container-certification-policy-guide",
 		CheckURL:         "https://connect.redhat.com/zones/containers/container-certification-policy-guide",
@@ -71,7 +71,7 @@ func (p *HasNoProhibitedPackagesCheck) Metadata() certification.Metadata {
 
 func (p *HasNoProhibitedPackagesCheck) Help() certification.HelpText {
 	return certification.HelpText{
-		Message:    "The container image should not include Red Hat Enterprise Linux (RHEL) kernel packages.",
+		Message:    "Check HasNoProhibitedPackages encountered an error. Please review the preflight.log file for more information.",
 		Suggestion: "Remove any RHEL packages that are not distributable outside of UBI",
 	}
 }
--- a/certification/internal/shell/has_required_labels.go
+++ b/certification/internal/shell/has_required_labels.go
@@ -50,7 +50,7 @@ func (p *HasRequiredLabelsCheck) Name() string {
 
 func (p *HasRequiredLabelsCheck) Metadata() certification.Metadata {
 	return certification.Metadata{
-		Description:      "Checking if the required labels are present in the container metadata.",
+		Description:      "Checking if the required labels (name, vendor, version, release, summary, description) are present in the container metadata.",
 		Level:            "good",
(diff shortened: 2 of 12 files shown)


komish (commented):
Thanks for taking care of this @edreyer-1!

Instead of adding `ErrorMessage` to our resultsText, what about repurposing our per-check Metadata and Help to fit the need? Happy to get more opinions on this.

Each check has a [Description](https://github.com/redhat-openshift-ecosystem/openshift-preflight/blob/main/certification/internal/shell/has_unique_tag.go#L50), [Suggestion](https://github.com/redhat-openshift-ecosystem/openshift-preflight/blob/main/certification/internal/shell/has_unique_tag.go#L59), and [Message](https://github.com/redhat-openshift-ecosystem/openshift-preflight/blob/main/certification/internal/shell/has_unique_tag.go#L59) which contains text indicating what the user should do in various scenarios. I'm seeing the Message and Description to be similar enough that we might not need both.

What do you think about capturing the gist of Message into the Description text, and then repurpose Message to answer the question of "where do I look when this check shows up in the error list"?

edreyer-1:
@komish yes I agree with you that we could combine description and message as they seem to be very similar purposes. Seems Message often just repeats the description and says it doesn't meet that requirement. Though I think that is already indicated by the check being in the failed part of the results. 
I definitely agree with your suggestion

samira-barouti:
I think we should probably involve the tool consumers to ensure the backward incompatible changes that we are making, is not breaking the pipeline. Other than that, I am also in favor of removing unnecessary data.

bcrochet (approved):
/lgtm

komish:
> I think we should probably involve the tool consumers to ensure the backward incompatible changes that we are making, is not breaking the pipeline. Other than that, I am also in favor of removing unnecessary data.

@samira-barouti I don't think the suggested path causes backwards incompatible changes, given that it preserves the existing fields and doesn't add or remove any fields. In the suggested case, we just repurpose the existing fields so we don't have to make any changes to the schema.

samira-barouti:
IMO this is still a backward incompatible change; it is true that the syntax does not change, but the semantic (i.e. behavior) of the API is changing.

komish:
I'll send this PR over to those folks for review!

komish:
Per conversation with @amisstea this shouldn't cause any issues on their end, so we should be good to change the meaning of `help` to guide the user in how to troubleshoot a check error. 

@edreyer-1 you should be set to implement!
@samira-barouti good catch!

certification/internal/shell/has_prohibited_packages.go
```diff
@@ -62,7 +62,7 @@ func (p *HasNoProhibitedPackagesCheck) Name() string {
 }
 func (p *HasNoProhibitedPackagesCheck) Metadata() certification.Metadata {
 	return certification.Metadata{
-		Description:      "Checks to ensure that the image in use does not contain prohibited packages.",
+		Description:      "Checks to ensure that the image in use does not include Red Hat Enterprise Linux (RHEL) kernel packages.",
```
komish, line 65: I _think_ the packages that we don't want to see include more than just kernel-related packages. Can you instead say something like `does not include prohibited packages, such as ...`

certification/internal/shell/has_required_labels.go
```diff
@@ -59,7 +59,7 @@ func (p *HasRequiredLabelsCheck) Metadata() certification.Metadata {
 
 func (p *HasRequiredLabelsCheck) Help() certification.HelpText {
 	return certification.HelpText{
-		Message:    "Container images must include the following metadata: name, vendor, version, release, summary, description",
+		Message:    "Check Check HasRequiredLabel encountered an error. Please review the preflight.log file for more information. encountered an error. Please review the preflight.log file for more information.",
```
komish, line 62: This looks like a duplicate paste

komish (changes requested):
Just a couple of updates!

certification/internal/shell/has_required_labels.go
edreyer-1: oops looks like I did double paste good catch

komish (approved):
/lgtm

Thanks @edreyer-1!

PR #103, opened 2021-07-15
Fix skopeo list-tags to support image digest

### LINKED ISSUE #98: Skopeo tag listing implementation does not properly strip digests from images
The skopeo command does not want tags or digests when doing inspections. Our current implementation will strip a tag, but does not handle digest stripping which causes the check to fail.

The line that needs an update is here: https://github.com/redhat-openshift-ecosystem/openshift-preflight/blob/main/certification/internal/shell/skopeo.go#L16

We need to handle cases where the user provided an image that looks like this:

registry.access.redhat.com/ubi8/ubi-micro@sha256:d389fa0cfee2091b3dd75a446fb982714db67bfb0916369f48985923890e2597

cc @samira-barouti (to make sure I parsed the implementation details correctly)

samira-barouti:
This PR leverages regular expression to strip the tag/digest from the image name. Skopeo requires the image name (without tag and digest) to list the available tags.

The current implementation strips a tag successfully, but fails to strip the image digest.

Fixes #98


certification/internal/shell/skopeo.go
```diff
@@ -13,13 +14,24 @@ import (
...
-	imageName := strings.Split(image, ":")[0]
+
+	re, err := regexp.Compile(`(?P<Image>[^@:]+)@*:*.+`)
+
+	if err != nil {
+		log.Error("unable to parse the image name", err)
+		log.Debug("error parsing the image name: ", err)
+		log.Trace("failure in attempt to parse the image name to strip tag/digest")
+		return nil, err
+	}
```
komish, line 25: Much prefer this over MustCompile, thanks!

certification/internal/shell/skopeo.go
```diff
@@ -13,13 +14,24 @@ import (
...
 	cmd := exec.Command("skopeo", "list-tags", "docker://"+imageName)
 
 	var stdout, stderr bytes.Buffer
 	cmd.Stdout = &stdout
 	cmd.Stderr = &stderr
-	err := cmd.Run()
+	err = cmd.Run()
 
 	if err != nil {
 		return nil, err
```
komish, line 37: Can you log stdout and stderr here at some higher-verbosity log level? 

E.g. 
```
		log.Debug("stdout: ", stdout.String())
		log.Debug("stderr: ", stderr.String())
```

certification/internal/shell/skopeo.go
```diff
@@ -13,13 +14,24 @@ import (
...
+	if err != nil {
+		log.Error("unable to parse the image name", err)
+		log.Debug("error parsing the image name: ", err)
+		log.Trace("failure in attempt to parse the image name to strip tag/digest")
+		return nil, err
+	}
+
+	imageName := re.FindStringSubmatch(image)[1]
+
 	cmd := exec.Command("skopeo", "list-tags", "docker://"+imageName)
```
komish, line 29: Can you log the command at a higher-verbosity log level? E.g.

```
	log.Trace("running skopeo with the following invocation ", cmd.Args)
```

certification/internal/shell/skopeo.go
komish: Do you think the regex logic is worth extracting into a function that we can then unit test against a series of different image formats?

komish (changes requested):
Couple of changes inline.

Also, this doesn't quite work how I expect if the user provides an image _without_ a tag/digest. 

I'm running this:

```
[root@lab container-test]# ../preflight check container registry.access.redhat.com/ubi8/ubi
```

I added some debug lines (the same/similar as to what's in the requested changes) and I see this is happening:

```
time="2021-07-15T15:48:49-05:00" level=info msg="running check: HasUniqueTag"
time="2021-07-15T15:48:49-05:00" level=trace msg="running skopeo with the following invocation [skopeo list-tags docker://registry.access.redhat.com/ubi8/ub]"      <<<< This Line
time="2021-07-15T15:48:50-05:00" level=debug msg="stdout: "
time="2021-07-15T15:48:50-05:00" level=debug msg="stdout: time=\"2021-07-15T15:48:50-05:00\" level=fatal msg=\"Error listing repository tags: Error fetching tags list: invalid status code from registry 404 (Not Found)\"\n"
time="2021-07-15T15:48:50-05:00" level=error msg="unable to execute skopeo on the image: exit status 1"
time="2021-07-15T15:48:50-05:00" level=info msg="check completed: HasUniqueTag" error="exit status 1" result=ERROR
```

The image that's being passed to `skopeo` is missing a character!

We'll likely want to address this, but I'm also wondering if we should reject the execution of `preflight` altogether if the user didn't provide a tag or digest. What do you think?

certification/internal/shell/skopeo.go
samira-barouti: This is a very good point! I'll isolate the regex stuffs into a function and add the unit tests.

certification/internal/shell/skopeo_test.go
```diff
@@ -0,0 +1,36 @@
...
+				Expect(err).ToNot(HaveOccurred())
+				Expect(image).To(Equal("quay.io/opdev/ubi-micro"))
+			})
+		})
+		Context("When it does not have neither a digest, nor a tag", func() {
+			It("should return error", func() {
+				image, err := skopeoEngine.ImageName("quay.io/opdev/ubi-micro")
+				Expect(err).To(HaveOccurred())
+				Expect(image).To(Equal(""))
+			})
```
bcrochet, line 32: Can you add test cases for a bare image (no registry)?

certification/internal/shell/skopeo.go
```diff
@@ -3,39 +3,71 @@ package shell
...
 	}
 	report.Stdout = stdout.String()
 	report.Stderr = stderr.String()
 
+	log.Debugf(fmt.Sprintf("detected these tags for %s: %s", imageName, report.Tags))
+
 	return &report, nil
 }
+
+func (e SkopeoCLIEngine) ImageName(image string) (string, error) {
```
bcrochet, line 59: Is there a reason for this to be exported?

bcrochet (changes requested):

certification/internal/shell/skopeo_test.go
samira-barouti: Done

certification/internal/shell/skopeo.go
samira-barouti: Ahh🤦🏻‍♀️ Fixed!

komish (approved):
/lgtm

bcrochet (approved):
/lgtm

PR #105, opened 2021-07-15
Improvements to HasLicenseCheck

### LINKED ISSUE #87: HasLicenseCheck should be more robust
Right now, when the /licenses directory exists, it is actually passing not because of the "No such file or directory" check. It's because the podman run exits with an error code. It never actually gets to the Validate method. Even if it did, "No such file or directory" would be on stderr, not stdout. So, stderr should in fact be checked for that string, rather than stdout.

If might make more sense to change this into a couple of different tests, that might yield more granular results, and allow us to be more prescriptive in the help suggestion.

First step would be to run 'stat /licenses'. And rather than parse the output, we should be checking the return code. Then, only if that succeeds (exit code of 0) should we then check the output of 'ls -1 /licenses | wc -l'. This will yield a number, that can then be checked against 0, rather than relying on the empty string.

skattoju:
- Fixes #87
- Added ExitCode to imageRunReport

Signed-off-by: Sid Kattoju <skattoju@redhat.com>

### DIFF
--- a/certification/internal/shell/has_license.go
+++ b/certification/internal/shell/has_license.go
@@ -1,6 +1,7 @@
 package shell
 
 import (
+	"fmt"
 	"strings"
 
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification"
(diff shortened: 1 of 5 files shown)


certification/internal/shell/has_license.go
```diff
@@ -8,40 +8,52 @@ import (
...
 	if err != nil {
-		ok, _ := p.validate(runReport.Stdout)
-		if ok {
-			log.Error("some error attempting to identify if /licenses container the license: ", err)
-			log.Debugf("Stdout: %s", runReport.Stdout)
-			log.Debugf("Stderr: %s", runReport.Stderr)
-			return "", err
-		}
-
+		log.Error("Error when checking for "+licensePath+" : ", err)
```
samira-barouti, line 36: I think using fmt.Sprintf for the first arg improves the readability.

samira-barouti (approved):
Nit... ;)

certification/internal/shell/has_license.go
bcrochet: Agreed

certification/internal/shell/shell_suite_test.go
```diff
@@ -88,6 +90,7 @@ func (bpe BadPodmanEngine) Run(cli.ImageRunOptions) (*cli.ImageRunReport, error)
 	runReport := cli.ImageRunReport{
 		Stdout: "Bad stadout",
 		Stderr: "Bad stderr",
+		ErrorCode: -1
```
bcrochet, line 93: Needs a comma.

bcrochet (changes requested):
Couple of fixes and a clarification.

certification/internal/shell/has_license_test.go
```diff
@@ -18,21 +18,22 @@ var _ = Describe("HasLicense", func() {
...
 				Expect(err).ToNot(HaveOccurred())
 				Expect(ok).To(BeTrue())
 			})
 		})
-		Context("When license is not found", func() {
+		Context("When licenses directory is not found", func() {
 			BeforeEach(func() {
 				engine := fakeEngine.(FakePodmanEngine)
-				engine.RunReportStdout = "No such file or directory"
+				engine.RunReportStdout = "stat: cannot stat '/licenses': No such file or directory"
```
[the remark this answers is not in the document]
skattoju: I get "exit status ..." on stderr and the "cannot stat ..." on stdout..

certification/internal/shell/has_license.go
skattoju: Agree as well

skattoju:
oops missed a thing will fix.

bcrochet (approved):
/lgtm

certification/internal/shell/has_license_test.go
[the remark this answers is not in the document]
komish: @bcrochet I get what you're seeing if I run this on my system, but I get what @skattoju gets if I run it through podman:

```
$ podman run -it --rm --entrypoint stat registry.access.redhat.com/ubi8/ubi:latest foo.txt -- 2>/dev/null
stat: cannot stat 'foo.txt': No such file or directory
```

(I should see nothing above if "cannot stat" was sent to stderr)

komish (changes requested):
NVM - merged before I could finish review! We'll address the rest in a future PR.

certification/internal/shell/has_license.go
```diff
@@ -1,47 +1,60 @@
...
-		if ok {
-			log.Error("some error attempting to identify if /licenses container the license: ", err)
-			log.Debugf("Stdout: %s", runReport.Stdout)
-			log.Debugf("Stderr: %s", runReport.Stderr)
-			return "", err
-		}
-
+		log.Error(fmt.Sprintf("Error when checking for %s : ", licensePath), err)
+		log.Errorf("Stdout: %s", runReport.Stdout)
+		log.Debugf("Stderr: %s", runReport.Stderr)
```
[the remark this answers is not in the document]
skattoju: Yeah thought of doing that .. but the stdout seemed more "human readable" the stderr was usually something like "exit status 1" so left it as is..

PR #110, opened 2021-07-19
Update Operator Bundle Validate to use OperatorSdkEngine

### LINKED ISSUE #33: [Check] Operator Bundle Data Linting
Run `operator-sdk bundle validate -b podman --select-optional name=operatorhub --select-optional name=community registry.redhat.io/quay/quay-operator-bundle`  
Check should fail on errors and print to stdout and the log
Check should pass on warnings but print warnings to stdout and the log

bcrochet:
* Rename Policy to Check
* Add unit test for Check
* Fix a typo in BundleValidateOptions
* Add Verbose flag to BundleValidateOptions
* Add support for Verbose to OperatorSdkEngine implementation

Fixes #33

### DIFF
--- a/certification/engine/engine.go
+++ b/certification/engine/engine.go
@@ -94,7 +94,7 @@ var hasLicenseCheck certification.Check = &shell.HasLicenseCheck{}
 var hasMinimalVulnerabilitiesCheck certification.Check = &shell.HasMinimalVulnerabilitiesCheck{}
 var hasUniqueTagCheck certification.Check = &shell.HasUniqueTagCheck{}
 var hasNoProhibitedCheck certification.Check = &shell.HasNoProhibitedPackagesCheck{}
-var validateOperatorBundle certification.Check = &shell.ValidateOperatorBundlePolicy{}
+var validateOperatorBundle certification.Check = &shell.ValidateOperatorBundleCheck{}
 var scorecardBasicSpecCheck certification.Check = &shell.ScorecardBasicSpecCheck{}
 var scorecardOlmSuiteCheck certification.Check = &shell.ScorecardOlmSuiteCheck{}
 
--- a/certification/internal/shell/validate_operator_bundle.go
+++ b/certification/internal/shell/validate_operator_bundle.go
@@ -1,42 +1,59 @@
 package shell
 
 import (
-	"os/exec"
-	"strings"
-
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification"
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/cli"
 	log "github.com/sirupsen/logrus"
 )
 
-type ValidateOperatorBundlePolicy struct {
+type ValidateOperatorBundleCheck struct {
 }
 
-func (p ValidateOperatorBundlePolicy) Validate(bundle string) (bool, error) {
-	stdouterr, err := exec.Command("operator-sdk", "bundle", "validate", "-b", "podman", "--verbose", bundle).CombinedOutput()
+func (p ValidateOperatorBundleCheck) Validate(bundle string) (bool, error) {
+	report, err := p.getDataToValidate(bundle)
 	if err != nil {
 		log.Error("Error will executing operator-sdk validate bundle: ", err)
 		return false, err
 	}
 
-	lines := strings.Split(string(stdouterr), "time=")
-
-	if strings.Contains(lines[len(lines)-1], "All validation tests have completed successfully") {
-		for _, line := range lines {
-			if strings.Contains(line, "level=warning") {
-				log.Warn("time= ", line)
-			}
-		}
-		return true, nil
-	}
-	log.Warn("The bundle image did not pass all of the validation tests")
-	return false, nil
+	return p.validate(report)
 }
 
-func (p ValidateOperatorBundlePolicy) Name() string {
+func (p ValidateOperatorBundleCheck) getDataToValidate(bundle string) (*cli.OperatorSdkBundleValidateReport, error) {
+	selector := []string{"community", "operatorhub"}
+	opts := cli.OperatorSdkBundleValidateOptions{
+		Selector:        selector,
+		Verbose:         true,
+		ContainerEngine: "podman",
+		OutputFormat:    "json-alpha1",
+	}
+
+	return operatorSdkEngine.BundleValidate(bundle, opts)
+}
+
+func (p ValidateOperatorBundleCheck) validate(report *cli.OperatorSdkBundleValidateReport) (bool, error) {
+	if !report.Passed || len(report.Outputs) > 0 {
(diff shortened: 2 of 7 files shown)


samira-barouti (approved):

certification/internal/shell/validate_operator_bundle.go
```diff
@@ -1,42 +1,59 @@
...
-	"os/exec"
-	"strings"
-
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification"
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/cli"
 	log "github.com/sirupsen/logrus"
 )
 
-type ValidateOperatorBundlePolicy struct {
+type ValidateOperatorBundleCheck struct {
```
komish, line 9: Nit: Same here re: check naming

certification/internal/shell/validate_operator_bundle.go
```diff
@@ -1,42 +1,59 @@
...
-		return true, nil
 	}
-	log.Warn("The bundle image did not pass all of the validation tests")
-	return false, nil
+	return report.Passed, nil
 }
 
-func (p ValidateOperatorBundlePolicy) Name() string {
+func (p ValidateOperatorBundleCheck) Name() string {
 	return "ValidateOperatorBundle"
```
komish, line 53: This should end with the `Check` suffix

certification/internal/shell/validate_operator_bundle.go
```diff
@@ -1,42 +1,59 @@
...
+			switch output.Type {
+			case "warning":
+				logFn = log.Warn
+			case "error":
+				logFn = log.Error
+			default:
+				logFn = log.Debug
 			}
+			logFn(output.Message)
 		}
```
komish, line 47: This is taking the output directly from `operator-sdk bundle validate` and mapping to our logging facilities in preflight, correct? This is a tricky approach for me; it directly ties the logging structure of operator-sdk to our logging structure. If their logging structure changes, we're unlikely to see it prior to shipping.

I know it's probably less pretty, but from the above perspective, would it be more stable to just pretty-print the entire log output at logging verbosity that we control so that the user has visibility into why something may have triggered a failure case? 

I do think we have this on other places as well. I'm not sure we need to solve it in this PR, but I did want to mention it to see what we could to give the user the needed visibility into things without hooking too deeply into logging behaviors which may not be reliable for our use case.

Also, if I misunderstood what this is doing, let me know/resolve away!

cli/operatorsdk.go
```diff
@@ -28,10 +28,11 @@ type OperatorSdkScorecardResult struct {
 }
 
 type OperatorSdkBundleValidateOptions struct {
-	LogLevel         string
-	ContainerEnginer string
-	Selector         []string
-	OutputFormat     string
+	LogLevel        string
+	ContainerEngine string
```
komish, line 32: Just want to confirm: this is here as a pass-through flag for operator-sdk, correct? We aren't exposing this to the user in any way, so this is _technically_ not configurable today, but may be something we can configure in the future?

komish (commented):
Couple of questions/nits - no explicit change required for the moment.

certification/engine/engine.go
```diff
@@ -94,7 +94,7 @@ var hasLicenseCheck certification.Check = &shell.HasLicenseCheck{}
 var hasMinimalVulnerabilitiesCheck certification.Check = &shell.HasMinimalVulnerabilitiesCheck{}
 var hasUniqueTagCheck certification.Check = &shell.HasUniqueTagCheck{}
 var hasNoProhibitedCheck certification.Check = &shell.HasNoProhibitedPackagesCheck{}
-var validateOperatorBundle certification.Check = &shell.ValidateOperatorBundlePolicy{}
+var validateOperatorBundle certification.Check = &shell.ValidateOperatorBundleCheck{}
```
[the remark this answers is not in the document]
komish: The goal here is keeping in line with existing checks which implicitly explain what "true" and "false" mean when reading their variable names..

certification/internal/shell/validate_operator_bundle.go
komish: Er, nevermind. I think I misread this.

certification/internal/shell/validate_operator_bundle.go
bcrochet: It really it just for reporting. We are in no way relying on this output to validate the check. I just wanted to log the messages at the appropriate level. It could just be log.Info for any of them.

cli/operatorsdk.go
bcrochet: Correct. Not configurable by the user. It may not be necessary if we never actually support docker instead of podman. Also, if we move to API calls instead of Exec calls, this would probably be moot.

certification/internal/shell/validate_operator_bundle.go
bcrochet: That's what I did?

certification/internal/shell/validate_operator_bundle.go
bcrochet: Isn't this the correct name? I am changing it to Check from Policy.

certification/engine/engine.go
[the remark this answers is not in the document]
bcrochet: That might be a little too verbose for my tastes... But I definitely see the reason for the comment. This one in particular is not very prescriptive. Let me think on it a bit.

certification/internal/shell/validate_operator_bundle.go
komish: By naming here, I meant the adjustment from ValidateOperatorBundle to something more prescriptive. That said, I'll mark this resolved because changing this in the previous comment would require it to fix this variable reference as well. There's no reason for me to have commented twice here. My apologies!

certification/internal/shell/validate_operator_bundle.go
komish: Definitely agree that the reporting is necessary. Just wary about associating the operator-sdk log levels to ours. I think I would prefer the entire log output from invoked tool just be a single log level on our side (some higher log level). 

Starting to feel like a nit, so I think we can table this and disregard. I'll leave it to your discretion 👍

jomkz (approved):
/lgtm pending any further comments from @komish

certification/internal/shell/validate_operator_bundle.go
bcrochet: As it isn't there log levels, but actual json output from operator-sdk, I'm ok with this one for the time being.

certification/engine/engine.go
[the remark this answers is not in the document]
bcrochet: I'll stick with this for the moment. Let's at least get the last Policy changed to Check. :)

PR #112, opened 2021-07-19
update HasRequiredLabels check to fall back to failing

### LINKED ISSUE #16: Ensure checks fall back to "failing" to avoid inadvertently passing a check
In a given check's `Validate()` function, we should ensure that the logic checks for explicit cases indicating we've passed (and falling back to failing) instead of explicit cases indicating we've failed (and falling back to passing).

So for example, say we have this simplified method (pseudo-code) 

```go
func (s *SomeCheck) Validate() (bool, error) {
	data := s.runSomeCheckAndGetSomeData()
	
	hasFailed := s.evaluateDataForFailure(data) // returnings a bool
	
	if hasFailed {
		return false, nil
	}

	return true, nil
}
```

In this case, we check to see if we've failed. If we have, then we return `false` and fail the check. If, for whatever reason, `s.evaluateDataForFailure(data)` returns a false positive due to some error in our parsing (i.e. `hasFailed == false), then we'd pass the check by default.

Instead, I'm thinking we should encourage the inverse:

```go
func (s *SomeCheck) Validate() (bool, error) {
	data := s.runSomeCheckAndGetSomeData()
	
	hasPassed := s.evaluateDataForFailure(data) // returnings a bool
	
	if hasPassed {
		return true, nil
	}

	return false, nil
}
```

**Implementation Notes:**
Ex. where this is done according to the above example:
https://github.com/redhat-openshift-ecosystem/openshift-preflight/blob/ae9b2c7a5e78c092d574de7425392ce1a6faea89/certification/internal/shell/base_on_ubi.go#L47https://github.com/redhat-openshift-ecosystem/openshift-preflight/blob/main/certification/internal/shell/base_on_ubi.go#L47

We would want to replace checks that return true as their final line, as seen here:
https://github.com/redhat-openshift-ecosystem/openshift-preflight/blob/ae9b2c7a5e78c092d574de7425392ce1a6faea89/certification/internal/shell/has_license.go#L32

To fix this, we would have to go through and update existing checks to make sure that we look for explicit indicators of success and then return true in those cases (as opposed to looking for explicit indicators of failure and then returning false.

edreyer-1:
We want the checks to fall back to failing instead of falling back on passing. We want to be checking for specific cases in which the check passes instead of looking for cases the check fails. This is because we want to prevent getting false positive checks and the default be failure instead of passing.

Currently the only check that this needs to be addressed in is HasRequiredLabels

fixes #16

### DIFF
--- a/certification/internal/shell/has_required_labels.go
+++ b/certification/internal/shell/has_required_labels.go
@@ -6,6 +6,8 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
+var requiredLabels = []string{"name", "vendor", "version", "release", "summary", "description"}
+
 type HasRequiredLabelsCheck struct{}
 
 func (p *HasRequiredLabelsCheck) Validate(image string) (bool, error) {
@@ -28,7 +30,6 @@ func (p *HasRequiredLabelsCheck) getDataForValidate(image string) (map[string]st
 }
 
 func (p *HasRequiredLabelsCheck) validate(labels map[string]string) (bool, error) {
-	requiredLabels := []string{"name", "vendor", "version", "release", "summary", "description"}
 	missingLabels := []string{}
 	for _, label := range requiredLabels {
 		if labels[label] == "" {
@@ -38,10 +39,9 @@ func (p *HasRequiredLabelsCheck) validate(labels map[string]string) (bool, error
 
 	if len(missingLabels) > 0 {
 		log.Warn("expected labels are missing:", missingLabels)
-		return false, nil
 	}
 
-	return true, nil
+	return len(missingLabels) == 0, nil
 }
 
 func (p *HasRequiredLabelsCheck) Name() string {


bcrochet (approved):
The code looks good. The question I do have is whether we want to keep the issue open or not... You state that this is the only one affected. Are we sure? If not, can you change the 'Fixes #16' to 'Related #16'? Otherwise, lgtm

edreyer-1:
I guess I should have said that only one of the checks needed to be updated for now. I think that the issue would apply to how we wanted to implement checks moving forward as well. I can update it to 'Related' if we want to keep that issue open.

bcrochet:
Just leave it as is. We can always reopen for tracking.

/lgtm

certification/internal/shell/has_required_labels.go
```diff
@@ -38,10 +38,10 @@ func (p *HasRequiredLabelsCheck) validate(labels map[string]string) (bool, error
 
 	if len(missingLabels) > 0 {
 		log.Warn("expected labels are missing:", missingLabels)
-		return false, nil
 	}
 
-	return true, nil
+	log.Warn("expected labels are missing:", missingLabels)
```
samira-barouti, line 43: Thank you Emily. Now that you are touching this file, could you please address the following improvements?
1) Please define requiredLabels as a const.
2) Are we printing the warning message twice? I would like to keep the previous logic where we check for a condition before printing it.

samira-barouti (approved):
/lgtm

PR #118, opened 2021-07-20
Add overall status of all checks

### LINKED ISSUE #106: Add an overall status of a policy execution to the results output
Yes, users need to parse 3 check lists to evaluate whether their preflight execution was successful.

Add a key in the results that reflects the overall state of the policy execution

N/A

## Additional context.
Implementation detail note - though all formatters get the Results in order to format them, we probably don't want them to do this evaluation at formatting time to reduce the likelihood of inconsistency.

edreyer-1:
Wanted to add an overall status for the policy execution. Right now the user has to parse 3 lists to see if their preflight was successful. The Status label will represent the overall status of the preflight policy execution. There are 3 possible status codes ("PASSED", "FAILED", "ERROR").

Passed - All checks were successful without any errors
Failed - At least one check failed but there were no errors
Error - An error occurred in at least one of the checks

Fixes #106

### DIFF
--- a/certification/internal/shell/engine.go
+++ b/certification/internal/shell/engine.go
@@ -11,6 +11,12 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
+const (
+	passed string = "PASSED"
+	failed string = "FAILED"
+	err    string = "ERROR"
+)
+
 // CheckEngine implements a CheckRunner.
 type CheckEngine struct {
 	Image  string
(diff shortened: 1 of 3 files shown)


certification/internal/shell/engine.go
```diff
@@ -57,12 +63,16 @@ func (e *CheckEngine) ExecuteChecks() error {
 		if err != nil {
 			log.WithFields(log.Fields{"result": "ERROR", "error": err.Error()}).Info("check completed: ", check.Name())
 			e.results.Errors = append(e.results.Errors, runtime.Result{Check: check, ElapsedTime: checkElapsedTime})
+			e.results.Status = "ERROR"
```
samira-barouti, line 66: Thank you Emily!
IMO, the status should be one of passed or failed. 

In any case, I think if you manage setting the `status` field outside of the for loop, the code is easier to read/maintain; for example the first time I saw lines 73-75, it took me a while to understand why there is such a condition for setting the status to FAILED.

samira-barouti (changes requested):

certification/internal/shell/engine.go
```diff
@@ -57,12 +63,16 @@ func (e *CheckEngine) ExecuteChecks() error {
...
 			continue
 		}
 
 		if !passed {
 			log.WithFields(log.Fields{"result": "FAILED"}).Info("check completed: ", check.Name())
 			e.results.Failed = append(e.results.Failed, runtime.Result{Check: check, ElapsedTime: checkElapsedTime})
+			if e.results.Status == "PASSED" {
```
jomkz, line 73: There is a lot of usage of the string literals `PASSED`, `FAILED` and `ERROR` in the status and the log messages. Can we make constants for these values? This way if we need to change that for whatever reason, we can just change it in one place. My general rule of thumb is if I am using the value more than once, I make a constant for it.

certification/internal/shell/engine.go
edreyer-1: Yeah I agree I will move setting the status outside the loop.

I guess I was wondering what exactly we want the status to indicate. It can either be an indicator of if it was successful or failed for whatever reason and the user can figure out what went wrong by looking at the check lists (Passed/Failed). Or it could represent the state of the policy execution (Passed/Failed/Error). Both are similar and I am fine either way.

certification/internal/shell/engine.go
```diff
@@ -11,6 +11,10 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
+const passed string = "PASSED"
```
samira-barouti, line 14: please use a const block:
const (
  passed string = "PASSED"
  failed string = "FAILED"
 // ...
)

samira-barouti (approved):
/lgtm pending confirmation from the team for having a binary status.

certification/internal/shell/engine.go
komish: +1 to changes mentioned by @samira-barouti. They look implemented me so I'm resolving this thread. Just closing the loop with this message.

komish (approved):
/lgtm

Thanks @edreyer-1

PR #122, opened 2021-07-21
Catch error in operator policy execution

komish:
Quick bugfix, When running `check operator`, we were not catching errors when calling `engine.ExecuteChecks()`, so we were getting results that reflected no checks were executed because the engine was throwing an error before any validations took place.

The `check container` command was updated to account for this, and I must have just ovelrooked `check container`. My apologies! 

### DIFF
--- a/cmd/check_operator.go
+++ b/cmd/check_operator.go
@@ -61,7 +61,9 @@ var checkOperatorCmd = &cobra.Command{
 		resultsOutputTarget := io.MultiWriter(os.Stdout, resultsFile)
 
 		// execute the checks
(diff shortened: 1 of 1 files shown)


komish:
Given `check operator` and `check container` basically share their execution workflow, there's certainly an opportunity to move the function to a single place, and abstract the distinctions (which is basically the runtime.Config), and then call the same function for both use cases. Perhaps that's a PR for another day.

bcrochet (approved):
/lgtm

samira-barouti (approved):

Issue #35, opened 2021-07-09
[Check] Operator Images Schema Version


tonytcampbell:
Need to confirm that all the related images for this Operator are using FAT manifest (schemaVersion 2)

Possible implementation
`podman pull` the bundle image
`podman image mount` the bundle image
Load <mount point>/manifest/<crd.yaml> into a variable
Use gojq to search for any values in the relatedImages field
Iterate the list of relatedImages and run `podman manifest inspect <link to related image>`
confirm that the JSOn returned has "schemaVersion": 2 for each related image

If all related images have a manifest schemaVersion: 2 then this check should pass
If any related image does not have schemaVersion: 2 then the check should fail and print the offending images to the log

PR #123, opened 2021-07-21
Add podman create, remove, copyFrom, and skopeo inspect

komish:
## Context

In implementing #35 (Operator Related Image Schema Version Check), I'll need to temporarily "run" a container so that I can copy files out of the running container.

In addition, I need to utilize `skopeo inspect --raw` in order to parse the schemaVersion value from the rawManifest.

This PR extends both the Podman and Skopeo engines and interfaces, adding the necessary functionality that I can leverage for to implement the check.

## Change Summary

- Defines `Create(...)` for the Podman Engine which allows us to run `podman create` in order to create a running container. I've also defined options for this method, and only included the specific option I need for my implementation. The FakePodmanEngine and BadPodmanEngine have corresponding methods to ensure they implement the interfaces.

- Defines `CopyFrom(...)` for the Podman Engine  which allows us to run `podman cp` with implicit direction (copy from container to host). FakePodmanEngine and BadPodmanEngine are updated with corresponding methods.

- Defines `Remove(...)` for the Podman Engine which allows us to run `podman rm` to clean up the container we're creating above. FakePodmanEngine and BadPodmanEngine are updated with corresponding methods.

- Defines `Inspect(...)` for the Skopeo Engine which allows us to run `skopeo inspect` to inspect images as we need. Podman inspect was already implemented, but I don't believe it has the ability to inspect the raw manifest as skopeo does. FakeSkopeoEngine has been updated to ensure it continues to implement the interface.

### DIFF
--- a/certification/internal/shell/podman.go
+++ b/certification/internal/shell/podman.go
@@ -9,6 +9,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"regexp"
+	"strings"
 
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/errors"
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/internal/utils/file"
@@ -84,7 +85,6 @@ func (pe PodmanCLIEngine) InspectImage(rawImage string, opts cli.ImageInspectOpt
 // ScanImage takes the `image` and runs `oscap-podman` against the image.
 // It also saves results to the vuln.html file in the current directory,
 // for the end users' reference.
-
 func (pe PodmanCLIEngine) ScanImage(image string) (*cli.ImageScanReport, error) {
 	ovalFileUrl := fmt.Sprintf("%s%s", ovalUrl, ovalFilename)
 	dir, err := ioutil.TempDir(".", "oval-")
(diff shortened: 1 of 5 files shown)


certification/internal/shell/podman.go
```diff
@@ -136,3 +136,89 @@ func (pe PodmanCLIEngine) ScanImage(image string) (*cli.ImageScanReport, error)
...
+		return &cli.PodmanCreateReport{
+			Stdout: stdout.String(),
+			Stderr: stderr.String(),
+		}, nil
+	}
+
+	return &cli.PodmanCreateReport{
+		Stdout:      stdout.String(),
+		Stderr:      stderr.String(),
+		ContainerID: strings.TrimSuffix(stdout.String(), "\n"),
```
bcrochet, line 169: Nit: TrimSpace seems to work better here. It encompasses more whitespace, and is more platform independent.

bcrochet (commented):
Small nit, otherwise looks good.

certification/internal/shell/skopeo.go
```diff
@@ -71,3 +71,35 @@ func (e SkopeoCLIEngine) imageName(image string) (string, error) {
...
+		return &cli.SkopeoInspectReport{Stdout: stdout.String(), Stderr: stderr.String()}, err
+	}
+
+	report := cli.SkopeoInspectReport{
+		Stdout: stdout.String(),
+		Stderr: stderr.String(),
+		Blob:   stdout.Bytes(),
+	}
+
+	return &report, nil
```
samira-barouti, line 104: Since we are not reusing report, it would be good to create it ad-hoc in the return statement.

samira-barouti (approved):
A small nit

certification/internal/shell/podman.go
komish: Fixed!

certification/internal/shell/skopeo.go
komish: Fixed!

samira-barouti (approved):
/lgtm

bcrochet (approved):
/lgtm

PR #124, opened 2021-07-22
Implements bundle check ensuring CSV related images are schema version 2

### LINKED ISSUE #35: [Check] Operator Images Schema Version

komish:
Fixes #35 

## Context
This check ensures that all related images found in a ClusterServiceVersion reflect the image manifest schema version 2.

In effect, we need to:

- Grab the bundle from the image
- Grab the CSV from the bundle
- Grab the related images section
- Inspect the image manifest for each related image
- Pass the check if all images are schema version 2

PR #123 added the required functionality in the skopeo and podman engines which enable this check.

## Change Notes

- First, we `create` the container image so that it exists as a stopped container in our container engine. This is required in order to copy data outside of the running container
- With the stopped container created, we copy `/`, which is where our bundle contents are expected to be, to our runtime directory (Note that this isn't great if someone puts their controller image, but I'm not sure we have an alternate course).
- Next, we use a library (from operator-sdk) to parse the bundle on disk and store it into a struct. With that struct in hand, we have an easy reference to the ClusterServiceVersion
- We grab the CSV's related images; if they have no related images, they auto pass the check.
- We iterate over each related image with `skopeo inspect --raw`, which returns the raw manifest for the image. This returns a json blob that _should_ have `schemaVersion` as a top-level key.
- We store the images with their schema versions in a map
- We evaluate the map, and ensure that all images have the desired schema version.

Regarding tests: the functions were written modularly to try to offload the amount of tests that need to rely on our fake engines. With that said, I'm seeing two calls which don't have explicit tests. The reason they do not is because they both make calls out to the library from operator-sdk that reads the bundles on disks. I think we could look at storing fixtures like these at some point in the future.


certification/internal/shell/has_v2_related_image_schema_version.go
```diff
@@ -0,0 +1,229 @@
...
+
+// extractBundleFromImage will extract the bundle contents from the specified image, and return the
+// path to the bundle on the filesystem. This is done by creating the container, copying the contents
+// out of the container, and then cleaning up the container.
+func (p *RelatedImagesAreSchemaVersion2Check) extractBundleFromImage(image string) (string, error) {
+	localBundlePath := "extracted-bundle"
+
+	// create the container. Bundle containers are non-runnable, so we need to
+	// add a dummy entrypoint so that we can create the container.
+	createReport, err := podmanEngine.Create(image, &cli.PodmanCreateOptions{Entrypoint: "false"})
```
tonytcampbell, line 69: As an alternative to running a stopped container would the following work?
`podman image mount registry.redhat.io/quay/quay-operator-bundle` for example.
then pass `<value from last command>/manifests/` to readBundle(...

komish: @tonytcampbell I wish! Unfortunately, it doesn't play out that way - 

```
(nonroot-user)$ podman image mount registry.redhat.io/quay/quay-operator-bundle
Error: cannot run command "podman image mount" in rootless mode, must execute `podman unshare` first
```

I discussed this with @bcrochet recently, and this is what led to the implementation submitted in #126. 

#126 solves this by executing a single check of preflight in an rootless mode obtained exactly how that error text states.

This PR does utilize the alternate approach (create a stopped container and pulling data from it).

certification/internal/shell/has_v2_related_image_schema_version.go
```diff
@@ -0,0 +1,229 @@
...
+)
+
+// RelatedImagesAreSchemaVersion2Check is part of the Operator policy and implements
+// the Check interface.
+type RelatedImagesAreSchemaVersion2Check struct{}
+
+// Validate checks the image manifest for each related image referenced in a
+// ClusterServiceVersion and ensures that the schema version used is version 2.
+func (p *RelatedImagesAreSchemaVersion2Check) Validate(image string) (bool, error) {
+
```
bcrochet, line 22: Nit: Extraneous line

certification/internal/shell/has_v2_related_image_schema_version.go
```diff
@@ -0,0 +1,229 @@
...
+	}
+
+	copyReport, err := podmanEngine.CopyFrom(createReport.ContainerID, "/", localBundlePath)
+	if err != nil {
+		log.Error("stdout: ", copyReport.Stdout)
+		log.Error("stderr: ", copyReport.Stderr)
+		return "", err
+	}
+
+	deleteReport, err := podmanEngine.Remove(createReport.ContainerID)
```
bcrochet, line 83: How do you feel about changing this to defer podmanEngine.Remove() and ignoring the return? We'd want the container cleaned up if CopyFrom fails.

bcrochet (changes requested):
Question and a nit.

certification/internal/shell/has_v2_related_image_schema_version.go
komish: @bcrochet Hmm, great point. I changed this defer the removal, and take any errors that return from deleting the container into warning log lines so that the log should reflect it at the warn log level. Check it, let me know if that works for you!

certification/internal/shell/has_v2_related_image_schema_version.go
komish: Fixed

certification/engine/engine.go
```diff
@@ -97,6 +97,7 @@ var hasNoProhibitedCheck certification.Check = &shell.HasNoProhibitedPackagesChe
 var validateOperatorBundle certification.Check = &shell.ValidateOperatorBundleCheck{}
 var scorecardBasicSpecCheck certification.Check = &shell.ScorecardBasicSpecCheck{}
 var scorecardOlmSuiteCheck certification.Check = &shell.ScorecardOlmSuiteCheck{}
+var relatedImageManifestSchemaVersionCheck certification.Check = &shell.RelatedImagesAreSchemaVersion2Check{}
```
samira-barouti, line 100: How about `csvImageSchemaVersionCheck`? The check name is a bit ambiguous.

certification/internal/shell/has_v2_related_image_schema_version.go
```diff
@@ -0,0 +1,230 @@
...
+	return p.validate(imageToSchemaVersion)
+}
+
+// getDataToValidate pulls a ClusterServiceVersion from the input operator bundle,
+// checks the ClusterServiceVersion's related images declaration, and assembles a list of
+// images and their image manifest schema version values.
+func (p *RelatedImagesAreSchemaVersion2Check) getDataToValidate(image string) (map[string]int, error) {
+	bundlePath, err := p.extractBundleFromImage(image)
+	if err != nil {
+		log.Error(err)
```
samira-barouti, line 36: Can we have an error message which describes what goes wrong?

certification/internal/shell/has_v2_related_image_schema_version.go
```diff
@@ -0,0 +1,230 @@
...
+
+	return imageStrings, nil
+}
+
+// inspectSchemaVersionForImages will inspect each input image using skopeo and returns a map of each image
+// with its corresponding schemaVersion. It is expeted that len(images) > 0.
+func (p *RelatedImagesAreSchemaVersion2Check) inspectSchemaVersionForImage(images []string) (map[string]int, error) {
+	inspectOpts := cli.SkopeoInspectOptions{Raw: true}
+
+	mapping := map[string]int{}
```
samira-barouti, line 128: How about imageToSchemaVersion? This makes it easier to understand what is the purpose of this variable

certification/engine/engine.go
komish: The CSV contains many locations where an image may be referenced (e.g. related images vs. deployment images).

certification/internal/shell/has_v2_related_image_schema_version.go
komish: The `extractBundleFromImage` method logs stdout and stderr if it encounters any errors along the way (three total tasks happening in that method, each being checked for errors - although one is now deferred and only emits warnings).

The next line of this `if` block returns the error message within the context of the calling function which also logs. 

Actually, I feel like I might be double logging...! What do you think?

certification/internal/shell/has_v2_related_image_schema_version.go
komish: Fixed!

bcrochet (approved):
/lgtm

certification/internal/shell/has_v2_related_image_schema_version.go
samira-barouti: I see. From a developer perspective, it might be good to see what is causing the issue. It might facilitate the troubleshooting.

komish:
Most recent force-push rebases to resolve conflicts.

samira-barouti (approved):

komish:
Originally, I intended to refactor this commit using the Unshare changes in #126, but those changes don't account for images and only account for cases where we want to mount a container. For now, I'll merge this as it is working without those changes, and when we add the `image mount` functionality, mirroring that of the container mounting functionality that was merged, we'll refactor this.

I'll create an issue associated with this work.

PR #126, opened 2021-07-22
Add a new command to run preflight in an unshare namespace

### LINKED ISSUE #104: HasNoProhibitedPackagesCheck throws an error when the image does not have the `rpm` binary
When running `check container` against UBI Micro (`registry.access.redhat.com/ubi8/ubi-micro@sha256:d389fa0cfee2091b3dd75a446fb982714db67bfb0916369f48985923890e2597`), the HasProhibitedPackages will fail because it cannot find `rpm`. We probably need to support cases where UBI Micro is used, and potentially need to fail gracefully in cases where an image that isn't RH*-based is used (though an error in those cases may be appropriate).

The error log was longer, but here's the gist:

```
level=error msg=\"container_linux.go:370: starting container process caused: exec: \\\"rpm\\\": executable file not found in $PATH\": OCI not found\n
```

For context, I believe this is because Micro UBI doesn't have any package tooling installed at all in an effort to make it smaller, but I wouldn't quote me on that fact.

bcrochet:
Introduces a new command: 'preflight check run'. This command will inspect
the environment for specific values, and exit if they are not found.
Values are:
PREFLIGHT_EXEC_CHECK - The name of a check as given by check.Name()
PREFLIGHT_EXEC_IMAGE - The image under test

What this gives us is the ability to run a test within a podman
namespace, via the unshare command.

This will allow a container to be created, mounted, have the filesystem
directly operated upon, then unmounted and removed, all within the context
of podman unshare.

HasNoProhibitedPackages has been modified to utilize this behavior, in order
to operator on the RPM db directly, without the need for the 'rpm' executable,
which happens to not be available in either a scratch or microubi-based
image.

Fixes #104 

### DIFF
--- a/certification/engine/engine.go
+++ b/certification/engine/engine.go
@@ -40,11 +40,17 @@ func NewForConfig(config runtime.Config) (CheckEngine, error) {
 		checks[i] = check
 	}
 
-	engine := &shell.CheckEngine{
+	var engine CheckEngine
+	engine = &shell.CheckEngine{
 		Image:  config.Image,
 		Checks: checks,
 	}
-
+	if config.Mounted {
+		engine = &shell.MountedCheckEngine{
+			Image: config.Image,
+			Check: checks[0],
+		}
+	}
 	return engine, nil
 }
 
@@ -52,13 +58,15 @@ func NewForConfig(config runtime.Config) (CheckEngine, error) {
 // if found; nil otherwise
 func queryChecks(checkName string) certification.Check {
 	// query Operator checks
-	check, exists := operatorPolicy[checkName]
-	if exists {
+	if check, exists := operatorPolicy[checkName]; exists {
 		return check
 	}
 	// if not found in Operator Policy, query container policy
-	check, exists = containerPolicy[checkName]
-	if exists {
+	if check, exists := containerPolicy[checkName]; exists {
+		return check
+	}
+	// Lastly, look at the mounted checks
+	if check, exists := mountedChecks[checkName]; exists {
 		return check
 	}
 
@@ -77,6 +85,7 @@ var hasNoProhibitedCheck certification.Check = &shell.HasNoProhibitedPackagesChe
 var validateOperatorBundle certification.Check = &shell.ValidateOperatorBundleCheck{}
 var scorecardBasicSpecCheck certification.Check = &shell.ScorecardBasicSpecCheck{}
 var scorecardOlmSuiteCheck certification.Check = &shell.ScorecardOlmSuiteCheck{}
+var hasNoProhibitedMountedCheck certification.Check = &shell.HasNoProhibitedPackagesMountedCheck{}
 
 var containerPolicy = map[string]certification.Check{
 	runAsNonRootCheck.Name():              runAsNonRootCheck,
@@ -93,6 +102,10 @@ var operatorPolicy = map[string]certification.Check{
 	validateOperatorBundle.Name():  validateOperatorBundle,
 	scorecardBasicSpecCheck.Name(): scorecardBasicSpecCheck,
 	scorecardOlmSuiteCheck.Name():  scorecardOlmSuiteCheck,
+}
+
+var mountedChecks = map[string]certification.Check{
+	hasNoProhibitedMountedCheck.Name(): hasNoProhibitedMountedCheck,
 }
 
 func makeCheckList(checkMap map[string]certification.Check) []string {
--- a/certification/errors/errors.go
+++ b/certification/errors/errors.go
@@ -22,3 +22,4 @@ var ErrOperatorSdkScorecardFailed = errors.New("failed to run operator-sdk score
 var ErrOperatorSdkBundleValidateFailed = errors.New("failed to run operator-sdk bundle validate")
 var ErrInvalidImageName = errors.New("failed to validate the image name")
 var ErrNoSocketFound = errors.New("neither podman or docker socket found")
+var ErrAlreadyInUnshare = errors.New("unshare is already active")
--- a/certification/internal/shell/has_prohibited_packages.go
+++ b/certification/internal/shell/has_prohibited_packages.go
@@ -1,65 +1,32 @@
 package shell
 
 import (
-	"bufio"
-	"strings"
+	"os"
 
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification"
-	"github.com/redhat-openshift-ecosystem/openshift-preflight/cli"
 	log "github.com/sirupsen/logrus"
 )
 
+// HasProhibitedPackages evaluates that the image does not contain prohibited packages,
+// which refers to packages that are not redistributable without an appropriate license.
 type HasNoProhibitedPackagesCheck struct{}
 
 func (p *HasNoProhibitedPackagesCheck) Validate(image string) (bool, error) {
-	line, err := p.getDataToValidate(image)
+	result, err := podmanEngine.UnshareWithCheck("HasNoProhibitedPackagesMounted", image, os.Args[0], "check", "run")
 	if err != nil {
-		log.Error("unable to get a list of all packages in the image")
+		log.Trace("unable to execute preflight in the unshare env")
+		log.Debugf("Stdout: %s", result.Stdout)
+		log.Debugf("Stderr: %s", result.Stderr)
 		return false, err
 	}
 
-	return p.validate(line)
-
-}
-
-func (p *HasNoProhibitedPackagesCheck) getDataToValidate(image string) (string, error) {
-	runOpts := cli.ImageRunOptions{
-		EntryPoint:     "rpm",
-		EntryPointArgs: []string{"-qa", "--queryformat", "%{NAME}\n"},
-		LogLevel:       "debug",
-		Image:          image,
-	}
-	runReport, err := podmanEngine.Run(runOpts)
-	if err != nil {
-		log.Error("unable to get a list of all packages in the image, error: ", err)
-		log.Debugf("Stdout: %s", runReport.Stdout)
-		log.Debugf("Stderr: %s", runReport.Stderr)
-		return "", err
-	}
-	return runReport.Stdout, nil
-}
-
-func (p *HasNoProhibitedPackagesCheck) validate(line string) (bool, error) {
-	scanner := bufio.NewScanner(strings.NewReader(line))
-	var prohibitedPackages []string
-	for scanner.Scan() {
-		for _, pkg := range prohibitedPackageList {
-			if pkg == scanner.Text() {
-				prohibitedPackages = append(prohibitedPackages, pkg)
-			}
-		}
-	}
-	log.Warn("The number of prohibited package found in the container image: ", len(prohibitedPackages))
-	if len(prohibitedPackages) > 0 {
-		log.Warn("found the following prohibited packages: ", prohibitedPackages)
-	}
-
-	return len(prohibitedPackages) == 0, nil
+	return result.PassedOverall, nil
 }
 
 func (p *HasNoProhibitedPackagesCheck) Name() string {
 	return "HasNoProhibitedPackages"
 }
+
 func (p *HasNoProhibitedPackagesCheck) Metadata() certification.Metadata {
 	return certification.Metadata{
 		Description:      "Checks to ensure that the image in use does not include prohibited packages, such as Red Hat Enterprise Linux (RHEL) kernel packages.",
@@ -74,28 +41,4 @@ func (p *HasNoProhibitedPackagesCheck) Help() certification.HelpText {
 		Message:    "Check HasNoProhibitedPackages encountered an error. Please review the preflight.log file for more information.",
 		Suggestion: "Remove any RHEL packages that are not distributable outside of UBI",
 	}
-}
-
-// prohibitedPackageList is a list of packages commonly present in the RHEL contianer images that are not redistributable
-// without proper licensing (i.e. packages that are not under the same availability as those found in UBI).
-// TODO: Confirm these packages are the only packages in immediate scope.
-var prohibitedPackageList = []string{
-	"grub",
-	"grub2",
-	"kernel",
-	"kernel-core",
-	"kernel-debug",
-	"kernel-debug-core",
-	"kernel-debug-modules",
-	"kernel-debug-modules-extra",
-	"kernel-debug-devel",
-	"kernel-devel",
-	"kernel-doc",
-	"kernel-modules",
-	"kernel-modules-extra",
-	"kernel-tools",
-	"kernel-tools-libs",
-	"kmod-kvdo",
-	"kpatch*",
-	"linux-firmware",
 }
--- a/certification/internal/shell/has_prohibited_packages_mounted.go
+++ b/certification/internal/shell/has_prohibited_packages_mounted.go
@@ -1,0 +1,116 @@
+package shell
+
+import (
+	"path/filepath"
+	"strings"
+
+	rpmdb "github.com/knqyf263/go-rpmdb/pkg"
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification"
+	log "github.com/sirupsen/logrus"
+)
+
+// HasProhibitedPackages evaluates that the image does not contain prohibited packages,
+// which refers to packages that are not redistributable without an appropriate license.
+type HasNoProhibitedPackagesMountedCheck struct{}
+
+func (p *HasNoProhibitedPackagesMountedCheck) Validate(image string) (bool, error) {
+	pkgList, err := p.getDataToValidate(image)
+	if err != nil {
+		log.Error("unable to get a list of all packages in the image")
+		return false, err
+	}
+
+	return p.validate(pkgList)
+}
+
+func (p *HasNoProhibitedPackagesMountedCheck) getDataToValidate(dir string) ([]string, error) {
+	log.Debugf("Mounted directory: %s", dir)
+
+	db, err := rpmdb.Open(filepath.Join(dir, "var", "lib", "rpm", "Packages"))
+	if err != nil {
+		return nil, err
+	}
+	pkgList, err := db.ListPackages()
+	if err != nil {
+		return nil, err
+	}
+	pkgs := make([]string, len(pkgList))
+	for i, pkg := range pkgList {
+		pkgs[i] = pkg.Name
+	}
+	return pkgs, nil
+}
+
+func (p *HasNoProhibitedPackagesMountedCheck) validate(pkgList []string) (bool, error) {
+	var prohibitedPackages []string
+	for _, pkg := range pkgList {
+		_, ok := prohibitedPackageList[pkg]
+		if ok {
+			prohibitedPackages = append(prohibitedPackages, pkg)
+			continue
+		}
+		for _, name := range prohibitedPackageGlobList {
+			if strings.HasPrefix(pkg, name) {
+				prohibitedPackages = append(prohibitedPackages, pkg)
+				continue
+			}
+		}
+	}
+
+	log.Warn("The number of prohibited package found in the container image: ", len(prohibitedPackages))
+	if len(prohibitedPackages) == 0 {
+		return true, nil
+	}
+
+	log.Warn("found the following prohibited packages: ", prohibitedPackages)
+	return false, nil
+}
+
+func (p *HasNoProhibitedPackagesMountedCheck) Name() string {
+	return "HasNoProhibitedPackagesMounted"
+}
+
+func (p *HasNoProhibitedPackagesMountedCheck) Metadata() certification.Metadata {
+	return certification.Metadata{
+		Description:      "Checks to ensure that the image in use does not include prohibited packages, such as Red Hat Enterprise Linux (RHEL) kernel packages.",
+		Level:            "best",
+		KnowledgeBaseURL: "https://connect.redhat.com/zones/containers/container-certification-policy-guide",
+		CheckURL:         "https://connect.redhat.com/zones/containers/container-certification-policy-guide",
+	}
+}
+
+func (p *HasNoProhibitedPackagesMountedCheck) Help() certification.HelpText {
+	return certification.HelpText{
+		Message:    "Check HasNoProhibitedPackages encountered an error. Please review the preflight.log file for more information.",
+		Suggestion: "Remove any RHEL packages that are not distributable outside of UBI",
+	}
+}
+
+// prohibitedPackageList is a list of packages commonly present in the RHEL contianer images that are not redistributable
+// without proper licensing (i.e. packages that are not under the same availability as those found in UBI).
+// Implementation detail: Use a map[string]struct{} so that lookups can be done, and determine their existence
+// in the map without having to do nested iteration.
+// TODO: Confirm these packages are the only packages in immediate scope.
+var prohibitedPackageList = map[string]struct{}{
+	"grub":                       {},
+	"grub2":                      {},
+	"kernel":                     {},
+	"kernel-core":                {},
+	"kernel-debug":               {},
+	"kernel-debug-core":          {},
+	"kernel-debug-modules":       {},
+	"kernel-debug-modules-extra": {},
+	"kernel-debug-devel":         {},
+	"kernel-devel":               {},
+	"kernel-doc":                 {},
+	"kernel-modules":             {},
+	"kernel-modules-extra":       {},
+	"kernel-tools":               {},
+	"kernel-tools-libs":          {},
+	"kmod-kvdo":                  {},
+	"linux-firmware":             {},
+}
+
+var prohibitedPackageGlobList = []string{
+	"kpatch",
+}
--- a/certification/internal/shell/mounted_engine.go
+++ b/certification/internal/shell/mounted_engine.go
@@ -1,0 +1,62 @@
+package shell
+
+import (
+	"time"
+
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification"
+	containerutil "github.com/redhat-openshift-ecosystem/openshift-preflight/certification/internal/utils/container"
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/runtime"
+	log "github.com/sirupsen/logrus"
+)
+
+type MountedCheckEngine struct {
+	Image string
+	Check certification.Check
+
+	results runtime.Results
+}
+
+// ExecuteChecks runs all checks stored in the check engine.
+func (e *MountedCheckEngine) ExecuteChecks() error {
+	log.Info("target image: ", e.Image)
+
+	e.results.TestedImage = e.Image
+	targetImage := e.Image
+
+	log.Info("running check: ", e.Check.Name())
+	// We want to know the time just for the check itself, so reset checkStartTime
+	checkStartTime := time.Now()
+
+	// run the validation
+	pass, err := containerutil.RunInsideContainerFS(podmanEngine, targetImage, e.Check.Validate)
+
+	checkElapsedTime := time.Since(checkStartTime)
+
+	switch {
+	case err != nil:
+		log.WithFields(log.Fields{"result": "ERROR", "error": err.Error()}).Info("check completed: ", e.Check.Name())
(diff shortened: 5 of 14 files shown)


tonytcampbell:
I think I understand the problem we are seeking to solve.  It feels like the `preflight check run` subcommand is  a bit clunky.  I think an optional flag might be better if we need to the user to tell us.   Is this something that could be detected automatically without the user needing to tell us anything?  Is ubi-micro the only base container that we need to solve for.  Initial thoughts...still looking this over.

cli/podman.go
```diff
@@ -77,13 +77,21 @@ type PodmanRemoveReport struct {
...
 	InspectImage(rawImage string, opts ImageInspectOptions) (*ImageInspectReport, error)
+	Mount(containerId string) (string, error)
 	Pull(rawImage string, opts ImagePullOptions) (*ImagePullReport, error)
+	Remove(containerID string) (*PodmanRemoveReport, error)
 	Run(opts ImageRunOptions) (*ImageRunReport, error)
 	Save(nameOrID string, tags []string, opts ImageSaveOptions) error
 	ScanImage(image string) (*ImageScanReport, error)
-	Create(rawImage string, opts *PodmanCreateOptions) (*PodmanCreateReport, error)
-	CopyFrom(containerID, sourcePath, destinationPath string) (*PodmanCopyReport, error)
-	Remove(containerID string) (*PodmanRemoveReport, error)
```
komish, line 88: Should we leave these as implemented functionality, even if we prefer to use unmount and unshare?

cmd/check_run.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/runtime"
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+)
+
+var checkRunCmd = &cobra.Command{
+	Use:   "run",
+	Short: "Run a single check inside of podman unshare",
+	Long: `This command will run a check while wrapped inside of podman unshare.
+It is a hidden command, and is not meant to be called by the user.
```
komish, line 19: nit: prefer that this is worded as "internal" vs. "hidden".

cmd/check_run.go
```diff
@@ -0,0 +1,87 @@
...
+It takes its input from environment variables only.`,
+	Hidden: true,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		// Expect an environment variable named PREFLIGHT_CHECK_EXEC
+		// It will specify the one check to execute inside of the podman unshare
+		// environment.
+		enabledCheck, ok := os.LookupEnv("PREFLIGHT_EXEC_CHECK")
+		if !ok {
+			log.Fatal("Enabled check envvar not specified")
+			os.Exit(1)
```
komish, line 29: We're using `RunE` for cobra; should we just return an error?

cmd/check_run.go
```diff
@@ -0,0 +1,87 @@
...
+		}
+		image, ok := os.LookupEnv("PREFLIGHT_EXEC_IMAGE")
+		if !ok {
+			log.Fatal("Operator image envvar not specified")
```
komish, line 33: Should we return an error/exit here?

cmd/check_run.go
```diff
@@ -0,0 +1,87 @@
...
+		}
+
+		cfg := runtime.Config{
+			Image:          image,
+			EnabledChecks:  []string{enabledCheck},
+			ResponseFormat: DefaultOutputFormat,
+			Mounted:        true,
+		}
```
komish, line 41: We don't have this problem today, but we'd have the challenge of figuring out what the original configuration was re: things like ResponseFormat. Just a note, nothing to change for now.

cmd/check_run.go
```diff
@@ -0,0 +1,87 @@
...
+			os.O_WRONLY|os.O_CREATE|os.O_APPEND,
+			0600,
+		)
+
+		if err != nil {
+			return err
+		}
+
+		// // also write to stdout
+		resultsOutputTarget := io.MultiWriter(os.Stdout, resultsFile)
```
komish, line 64: Does this overwrite the preflight.log?

cmd/check_run.go
```diff
@@ -0,0 +1,87 @@
...
+
+		// execute the checks
+		engine.ExecuteChecks()
```
komish, line 67: We should be catching an error here, and evaluating `if err != nil {...}`

cmd/check_run.go
```diff
@@ -0,0 +1,87 @@
...
+		// return results to the user and then close output files
+		formattedResults, err := formatter.Format(results)
+		if err != nil {
+			return err
+		}
+
+		fmt.Fprint(resultsOutputTarget, string(formattedResults))
+		// if err := resultsFile.Close(); err != nil {
+		// 	return err
+		// }
```
komish, line 79: If I'm understanding this correctly, this command would effectively run a single command in a specific runtime environment (i.e. the `podman unshare` namespace). From this perspective, we'd have to figure out exactly how we handle the outputs.

I may be misunderstanding, but it feels like we should be capturing the output of this internal execution and _merging_ it with the rest of our output. I may be misunderstanding the complete picture here, so any guidance would help!

certification/internal/shell/has_prohibited_packages.go
```diff
@@ -39,16 +53,29 @@ func (p *HasNoProhibitedPackagesCheck) getDataToValidate(image string) (string,
 	return runReport.Stdout, nil
```
komish, line 53: Was this supposed to go away?

certification/internal/shell/has_prohibited_packages.go
```diff
@@ -12,17 +13,30 @@ import (
...
+		if err != nil {
+			log.Error("unable to execute preflight in the unshare env")
+			log.Debugf("Stdout: %s", result.Stdout)
+			log.Debugf("Stderr: %s", result.Stderr)
+			return false, err
+		}
+		return true, nil
+	}
+
+	log.Debug("This should be preflight in the unshare env")
```
komish, line 28: nit: remove this, or otherwise reword.

certification/internal/shell/has_prohibited_packages.go
```diff
@@ -39,16 +53,29 @@ func (p *HasNoProhibitedPackagesCheck) getDataToValidate(image string) (string,
...
+		return nil, err
+	}
+	pkgList, err := db.ListPackages()
+	if err != nil {
+		return nil, err
+	}
+	return pkgList, nil
+}
+
+func (p *HasNoProhibitedPackagesCheck) validate(pkgList []*rpmdb.PackageInfo) (bool, error) {
```
komish, line 70: On the whole, I much prefer this more-direct approach of reading the packages in the container. The only caveat I have is the library hasn't seen much maintenance in some time.

certification/internal/shell/has_prohibited_packages.go
```diff
@@ -79,23 +106,23 @@ func (p *HasNoProhibitedPackagesCheck) Help() certification.HelpText {
...
-	"kernel-devel",
-	"kernel-doc",
-	"kernel-modules",
-	"kernel-modules-extra",
-	"kernel-tools",
-	"kernel-tools-libs",
-	"kmod-kvdo",
-	"kpatch*",
-	"linux-firmware",
+var prohibitedPackageList = map[string]struct{}{
```
komish, line 109: I'm not going to pretend like I have a firm grasp on language efficiency semantics, but making this a `map[string]interface{}`, and setting all values to `nil` be a more efficient? I'm understanding that you want this to be a map to facilitate lookups, which is fine, but I"m wondering if empty structs are ... less efficient.

Like I said, not even going to pretend like I fully understand the space complexity concepts as it applies in this particular case. Happy to just resolve and move forward.

certification/internal/shell/has_prohibited_packages.go
```diff
@@ -79,23 +106,23 @@ func (p *HasNoProhibitedPackagesCheck) Help() certification.HelpText {
...
+	"kernel-debug-modules-extra": {},
+	"kernel-debug-devel":         {},
+	"kernel-devel":               {},
+	"kernel-doc":                 {},
+	"kernel-modules":             {},
+	"kernel-modules-extra":       {},
+	"kernel-tools":               {},
+	"kernel-tools-libs":          {},
+	"kmod-kvdo":                  {},
+	"kpatch*":                    {},
```
komish, line 126: This glob likely won't match the package name; we might need to explicitly include `kpatch` packages here.

certification/internal/shell/has_prohibited_packages.go
```diff
@@ -79,23 +106,23 @@ func (p *HasNoProhibitedPackagesCheck) Help() certification.HelpText {
 // prohibitedPackageList is a list of packages commonly present in the RHEL contianer images that are not redistributable
 // without proper licensing (i.e. packages that are not under the same availability as those found in UBI).
 // TODO: Confirm these packages are the only packages in immediate scope.
```
komish, line 108: This likely needs a reword now that this is a map.

certification/internal/shell/mounted_engine.go
```diff
@@ -0,0 +1,63 @@
...
+	// We want to know the time just for the check itself, so reset checkStartTime
+	checkStartTime := time.Now()
+
+	// run the validation
+	pass, er := containerutil.RunInsideContainerFS(podmanEngine, targetImage, e.Check.Validate)
+
+	checkElapsedTime := time.Since(checkStartTime)
+
+	switch {
+	case er != nil:
```
komish, line 37: nit: `er` should probably be `err`

certification/internal/shell/mounted_engine.go
```diff
@@ -0,0 +1,63 @@
...
+	Image string
+	Check certification.Check
+
+	results runtime.Results
+}
+
+// ExecuteChecks runs all checks stored in the check engine.
+func (e *MountedCheckEngine) ExecuteChecks() error {
+	log.Info("target image: ", e.Image)
+	// check if the image needs downloading
```
komish, line 22: can probably remove this comment as we're assuming the image has been downloaded before we ever reach this point.

komish (commented):
Some comments inline. I'll need to do further testing and another round of reviews. This preflight-ception is a bit tricky.

bcrochet:
> I think I understand the problem we are seeking to solve. It feels like the `preflight check run` subcommand is a bit clunky. I think an optional flag might be better if we need to the user to tell us. Is this something that could be detected automatically without the user needing to tell us anything? Is ubi-micro the only base container that we need to solve for. Initial thoughts...still looking this over.

This is a completely hidden command. This is not meant for user consumption. This code does detect when it should be run. While the first use case is for ubi-micro, there are others where having access to the mounted filesystem will benefit. Like in #124 . And for generating the bundle checksum.

cli/podman.go
bcrochet: I just reordered them. Alphabetically. :)

cmd/check_run.go
bcrochet: Done.

cmd/check_run.go
bcrochet: Sure.

cmd/check_run.go
bcrochet: Done.

cmd/check_run.go
bcrochet: I'll have to check on that before we merge.

cmd/check_run.go
bcrochet: Done.

cmd/check_run.go
bcrochet: You are correct. I need to figure out the results processing.

certification/internal/shell/has_prohibited_packages.go
bcrochet: Yes.

certification/internal/shell/has_prohibited_packages.go
bcrochet: Done.

cli/podman.go
komish: Oh, whoops! I can't read. Resolved!

certification/internal/shell/has_prohibited_packages.go
bcrochet: It's pretty simple. And last update was Dec. 2020, so not too long ago. I'm changing the interface to be more generic. Only the getDataToValidate will have a dependency on the library.

certification/internal/shell/has_prohibited_packages.go
bcrochet: I hadn't even noticed that. Let me think on that one.

bcrochet: Ok. I have an ok solution. LMKWYT.

certification/internal/shell/has_prohibited_packages.go
bcrochet: Reworked it slightly.

certification/internal/shell/mounted_engine.go
bcrochet: Done.

certification/internal/shell/mounted_engine.go
bcrochet: Fixed.

certification/internal/shell/has_prohibited_packages.go
komish: Globs should work! Resolving

certification/engine/engine.go
```diff
@@ -40,9 +40,17 @@ func NewForConfig(config runtime.Config) (CheckEngine, error) {
...
+	if config.Mounted {
+		engine = &shell.MountedCheckEngine{
+			Image: config.Image,
+			Check: checks[0],
+		}
+	} else {
+		engine = &shell.CheckEngine{
+			Image:  config.Image,
+			Checks: checks,
+		}
```
komish, line 53: Nit: can we avoid this `else` here? Instead - just assign shell.CheckEngine with the values we want to `engine` in line 43, then your if block will change that if it needs to do so.

certification/internal/shell/has_prohibited_packages.go
```diff
@@ -1,65 +1,28 @@
...
 )
 
 type HasNoProhibitedPackagesCheck struct{}
 
 func (p *HasNoProhibitedPackagesCheck) Validate(image string) (bool, error) {
-	line, err := p.getDataToValidate(image)
+	result, err := podmanEngine.Unshare("HasNoProhibitedPackagesMounted", image, "./preflight", "check", "run")
 	if err != nil {
-		log.Error("unable to get a list of all packages in the image")
+		log.Error("unable to execute preflight in the unshare env")
```
komish, line 13: On one hand, we _need_ to log this. On the other, your average user seeing this would likely have no idea exactly how to "fix this check" without the knowledge that we've recursed (i.e. implementation details of preflight as a tool).

Maybe we need to do this in at a higher verbosity, and log simply that this check has not passed at a lower verbosity?

certification/internal/shell/has_prohibited_packages.go
```diff
@@ -1,65 +1,28 @@
...
 import (
-	"bufio"
-	"strings"
-
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification"
-	"github.com/redhat-openshift-ecosystem/openshift-preflight/cli"
 	log "github.com/sirupsen/logrus"
 )
 
 type HasNoProhibitedPackagesCheck struct{}
```
komish, line 8: Can you document this struct similar to what is done in #131?

certification/internal/shell/has_prohibited_packages_mounted.go
```diff
@@ -0,0 +1,113 @@
...
+import (
+	"path/filepath"
+	"strings"
+
+	rpmdb "github.com/knqyf263/go-rpmdb/pkg"
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification"
+	log "github.com/sirupsen/logrus"
+)
+
+type HasNoProhibitedPackagesMountedCheck struct{}
```
komish, line 12: Can you document this struct similar to what's done in #131?

certification/internal/shell/has_prohibited_packages_mounted.go
```diff
@@ -0,0 +1,113 @@
...
+			}
+		}
+	}
+
+	log.Warn("The number of prohibited package found in the container image: ", len(prohibitedPackages))
+	if len(prohibitedPackages) > 0 {
+		log.Warn("found the following prohibited packages: ", prohibitedPackages)
+	}
+
+	return len(prohibitedPackages) == 0, nil
```
komish, line 63: nit: Can you instead evaluate `len(prohibitedPackages) == 0` and return `true, nil`, and otherwise warn as you've done on 60 outside of the loop, and return `false, nil`? I think on the whole that I prefer the explicit returns over the internally evaluated cases such as line line 63.

certification/internal/shell/podman.go
```diff
@@ -222,3 +222,64 @@ func (pe PodmanCLIEngine) Remove(containerID string) (*cli.PodmanRemoveReport, e
...
+	err := cmd.Run()
+	report := &cli.PodmanUnmountReport{Stdout: stdout.String(), Stderr: stderr.String()}
+	if err != nil {
+		log.Errorf("could not run unmount. Output: %s", stderr.String())
+		return report, err
+	}
+	return report, nil
+}
+
+func (pe PodmanCLIEngine) Unshare(check, image string, command ...string) (*cli.PodmanUnshareReport, error) {
```
komish, line 255: We're directly coupling the check names and the image to thsi function signature of the podman engine. If we need to reuse this for any other reason, we would be bound to always having PREFLIGHT_EXEC_CHECK/IMAGE as environment variables., Should we pass an environment instead?

certification/internal/utils/container/helper.go
```diff
@@ -66,3 +67,56 @@ func SaveContainerToFilesystem(podmanEngine cli.PodmanEngine, imageLog string) (
...
 }
+
+type ContainerFn func(string) (bool, error)
+
+func unmountAndRemove(podmanEngine cli.PodmanEngine, containerId string) {
+	podmanEngine.Unmount(containerId)
+	podmanEngine.Remove(containerId)
+}
+
+func RunInsideContainerFS(podmanEngine cli.PodmanEngine, image string, containerFn ContainerFn) (bool, error) {
```
komish, line 78: Document this function

certification/internal/utils/container/helper.go
```diff
@@ -66,3 +67,56 @@ func SaveContainerToFilesystem(podmanEngine cli.PodmanEngine, imageLog string) (
...
+		log.Error("could not mount filesystem", err)
+		return false, err
+	}
+
+	defer unmountAndRemove(podmanEngine, containerId)
+
+	return containerFn(report.MountDir)
+}
+
+func GenerateBundleHash(podmanEngine cli.PodmanEngine, image string) error {
```
komish, line 98: Document this function, or rather - if this is not related, should we add this in a subsequent PR? I don't see it in use anywhere yet.

certification/internal/shell/has_prohibited_packages.go
```diff
@@ -1,65 +1,28 @@
...
 
 func (p *HasNoProhibitedPackagesCheck) Validate(image string) (bool, error) {
-	line, err := p.getDataToValidate(image)
+	result, err := podmanEngine.Unshare("HasNoProhibitedPackagesMounted", image, "./preflight", "check", "run")
```
komish, line 11: Hmm... can we assume this is where `preflight` is on a given system? Mine, for example, is `../preflight`. Some people might have it in path.

cmd/check_run.go
komish: It does indeed:

```
[komish@lab test]$ cat preflight.log 
time="2021-07-27T15:47:28-05:00" level=warning msg="The number of prohibited package found in the container image: 0"
time="2021-07-27T15:47:29-05:00" level=info msg="check completed: HasNoProhibitedPackages" result=true
```

komish (changes requested):
More review!

certification/engine/engine.go
bcrochet: Will do.

certification/internal/shell/has_prohibited_packages.go
bcrochet: Lowering to trace

certification/internal/shell/has_prohibited_packages_mounted.go
bcrochet: Sure

certification/internal/shell/podman.go
bcrochet: I started there, but didn't like that the caller needed to know that it was being used as environment. I preferred to pass them as discreet variables, and leave the environment part as an implementation detail within Unshare.

certification/internal/utils/container/helper.go
bcrochet: Done.

certification/internal/utils/container/helper.go
bcrochet: Removed.

certification/internal/shell/has_prohibited_packages.go
bcrochet: Good point. I think I have a solution.

komish (changes requested):
A few issues noted in-line

certification/errors/errors.go
```diff
@@ -21,4 +21,8 @@ var ErrImageScanFailed = errors.New("failed to scan image")
 var ErrOperatorSdkScorecardFailed = errors.New("failed to run operator-sdk scorecard")
 var ErrOperatorSdkBundleValidateFailed = errors.New("failed to run operator-sdk bundle validate")
 var ErrInvalidImageName = errors.New("failed to validate the image name")
+<<<<<<< HEAD
 var ErrNoSocketFound = errors.New("neither podman or docker socket found")
+=======
+var ErrAlreadyInUnshare = errors.New("unshare is already active")
+>>>>>>> 6a65a8a (Add a new command to run preflight in an unshare namespace)
```
komish, line 28: Incomplete rebase here

cmd/check_run.go
```diff
@@ -0,0 +1,75 @@
...
+
+		results := engine.Results()
+
+		// return results to the user and then close output files
+		formattedResults, err := formatter.Format(results)
+		if err != nil {
+			return err
+		}
+
+		fmt.Fprint(os.Stdout, string(formattedResults))
```
komish, line 67: I know I highlighted this already, but I do see this changed from what I remember seeing before. This still overwrites preflight.log.

Given the following execution:

```
$ PFLT_LOGLEVEL=trace ../preflight check container quay.io/komish/preflight-test-container-passes:latest
time="2021-07-28T11:11:49-05:00" level=info msg="config file not found, proceeding without it"
time="2021-07-28T11:11:49-05:00" level=info msg="target image: quay.io/komish/preflight-test-container-passes:latest"
time="2021-07-28T11:11:49-05:00" level=info msg="downloading image"
time="2021-07-28T11:11:51-05:00" level=info msg="running check: HasRequiredLabel"
time="2021-07-28T11:11:51-05:00" level=info msg="check completed: HasRequiredLabel" result=true
time="2021-07-28T11:11:51-05:00" level=info msg="running check: BasedOnUbi"
time="2021-07-28T11:11:53-05:00" level=info msg="check completed: BasedOnUbi" result=true
time="2021-07-28T11:11:53-05:00" level=info msg="running check: HasLicense"
time="2021-07-28T11:11:58-05:00" level=info msg="1 Licenses found"
time="2021-07-28T11:11:58-05:00" level=info msg="check completed: HasLicense" result=true
time="2021-07-28T11:11:58-05:00" level=info msg="running check: HasMinimalVulnerabilities"
time="2021-07-28T11:11:58-05:00" level=debug msg="Oval file dir: oval-790686789"
time="2021-07-28T11:11:58-05:00" level=debug msg="Oval file path: oval-790686789/rhel-8.oval.xml.bz2"
time="2021-07-28T11:11:58-05:00" level=error msg="unable to execute oscap-podman on the image: This script cannot run in rootless mode.\nvuln.html"
time="2021-07-28T11:11:58-05:00" level=error msg="unable to scan the image: exit status 1"
time="2021-07-28T11:11:58-05:00" level=debug msg="Stdout: []"
time="2021-07-28T11:11:58-05:00" level=info msg="check completed: HasMinimalVulnerabilities" ERROR="failed to scan image: exit status 1" result="failed to scan image: exit status 1"
time="2021-07-28T11:11:58-05:00" level=info msg="running check: HasUniqueTag"
time="2021-07-28T11:11:58-05:00" level=trace msg="running skopeo with the following invocation[list-tags docker://quay.io/komish/preflight-test-container-passes]"
time="2021-07-28T11:11:59-05:00" level=debug msg="detected these tags for quay.io/komish/preflight-test-container-passes: [latest 0.0.1]"
time="2021-07-28T11:11:59-05:00" level=info msg="check completed: HasUniqueTag" result=true
time="2021-07-28T11:11:59-05:00" level=info msg="running check: HasNoProhibitedPackages"
time="2021-07-28T11:12:01-05:00" level=info msg="check completed: HasNoProhibitedPackages" result=true
time="2021-07-28T11:12:01-05:00" level=info msg="running check: RunAsNonRoot"
time="2021-07-28T11:12:03-05:00" level=debug msg="the runtime user id is 1000"
time="2021-07-28T11:12:03-05:00" level=info msg="check completed: RunAsNonRoot" result=true
time="2021-07-28T11:12:03-05:00" level=info msg="running check: LayerCountAcceptable"
time="2021-07-28T11:12:03-05:00" level=debug msg="detected 4 layers in image"
time="2021-07-28T11:12:03-05:00" level=info msg="check completed: LayerCountAcceptable" result=true
{
    "image": "quay.io/komish/preflight-test-container-passes:latest",
    "Status": "FAILED",
    "validation_lib_version": {
        "version": "0.0.0",
        "commit": "6a65a8a7cd3d558c558d944702a72efe5ef89ec6"
    },
    "results": {
        ... truncated for brevity ...
    }
}
```

My logfile only contained a partial result:

```
$ cat preflight.log 
time="2021-07-28T11:12:00-05:00" level=warning msg="The number of prohibited package found in the container image: 0"
time="2021-07-28T11:12:01-05:00" level=info msg="check completed: HasNoProhibitedPackages" result=true
time="2021-07-28T11:12:01-05:00" level=info msg="running check: RunAsNonRoot"
time="2021-07-28T11:12:03-05:00" level=debug msg="the runtime user id is 1000"
time="2021-07-28T11:12:03-05:00" level=info msg="check completed: RunAsNonRoot" result=true
time="2021-07-28T11:12:03-05:00" level=info msg="running check: LayerCountAcceptable"
time="2021-07-28T11:12:03-05:00" level=debug msg="detected 4 layers in image"
time="2021-07-28T11:12:03-05:00" level=info msg="check completed: LayerCountAcceptable" result=true
```

How much the log file contains is a function of the location of the check that utilizes `check run`. Our checks are run in a random order every execution, so if a check that utilizes `check run` is executed in the beginning, we might see more logs. If it's executed at the end, we see less logs.

I think you're running into this because we initialize our logger in the `init()` function for package `cmd`, and we, by default, write to stdout and preflight.log if the user has not specified a configuration.

https://github.com/redhat-openshift-ecosystem/openshift-preflight/blob/main/cmd/root.go#L28

You'll probably need to change the built-in logger's outputs to work around this.

komish (changes requested):
My bad, more comments that didn't get included in the last review for some reason.

certification/errors/errors.go
bcrochet: Helps if I actually save.

komish (approved):
/lgtm

While I prefer not to use the hidden command chain, I do understand that it solves the problem of access to the container filesystem without having to muck around with running and copying things out.

All that said, this appears to work as I expect. From that perspective, this LGTM.

bcrochet:
> /lgtm
> 
> While I prefer not to use the hidden command chain, I do understand that it solves the problem of access to the container filesystem without having to muck around with running and copying things out.
> 
> All that said, this appears to work as I expect. From that perspective, this LGTM.

I am sympathetic to the aversion of the hidden command chain, in the short-term it solve a problem until we can rely on native golang libraries. Thanks for all the reviews.

jomkz (approved):
/lgtm with some of the same comments already made.

PR #127, opened 2021-07-23
Fix constant misusing the symbol err

komish:
Just a quick bugfix. There was a package-level constant identified as `err` that could have caused issues, as `err` is the idiomatic way to represent an error returned from a function.


bcrochet (approved):
/lgtm

samira-barouti (approved):

PR #128, opened 2021-07-24
WIP: Implement DeployableByOLM check

samira-barouti:
* Add custom Go clients to create CRs (OperatorGroup, CatalogSource, Subscription)
* Add OpenshiftEngine to interact with kube api server


certification/engine/engine.go
```diff
@@ -41,6 +41,38 @@ type CheckRunner interface {
 	Results() runtime.Results
 }
 
+// Register all checks
```
bcrochet, line 44: Something isn't right here... Why did this entire block of code change?

certification/internal/shell/deployable_by_olm.go
```diff
@@ -0,0 +1,63 @@
+package shell
```
bcrochet, line 1: So... This probably shouldn't go in shell? It's using an actual golang client, as opposed to shell commands. I don't really have a suggested package name. @komish thoughts?

certification/internal/shell/openshift.go
```diff
@@ -0,0 +1,248 @@
+package shell
```
bcrochet, line 1: Again, this one probably shouldn't be in shell either.

certification/internal/utils/file/helper.go
```diff
@@ -48,3 +48,10 @@ func Unzip(bzipfile string, destination string) error {
 	out.Close()
 	return nil
 }
+
+func HomeDir() string {
+	if h := os.Getenv("HOME"); h != "" {
+		return h
+	}
+	return os.Getenv("USERPROFILE") // windows
```
bcrochet, line 56: Is this really necessary? Is this something we anticipate having to support?

bcrochet (changes requested):

certification/engine/engine.go
komish: I think it switched positions from below `NewForConfig(...)` to above it in the given file?

samira-barouti: yes, I moved the definition of these global vars to the top.

certification/internal/utils/file/helper.go
komish: I think I would prefer requiring an explicitly defined kubeconfig location. @samira-barouti what do you think?

komish:
Thanks for doing all the leg work here @samira-barouti! Can I have you write up a quick blurb about the overall goal and design here? Any gotchas that we need to be cognizant of for implementing this check? What exactly does the DeployableByOLM do to perform its validation?

certification/internal/shell/deployable_by_olm.go
komish: I liked `client`. which I think @samira-barouti already defined. 

With that said, we're getting dangerously close to blurring the line between our proposed "v1" and "v2" implementations, where here "v1" meant exec-ing out to shell commands, and "v2" meant explicitly doing everything using as close to a pure-go mentality as we could. 

Do we think we need to adapt our internal structure a bit to fit this direction?

certification/internal/shell/openshift.go
samira-barouti: Good point! I created a new package called k8s and moved the openshiftEngine and the check to this new package. Any feedback on the package name is welcome.

certification/internal/utils/file/helper.go
samira-barouti: I removed this new function and leveraged viper to look for kubeconfig under $HOME/.kubee/config as the default location.

samira-barouti:
@komish I updated the PR description to include details on what needs to be done, and what changes were made to the code base.

certification/internal/client/client.go
```diff
@@ -0,0 +1,36 @@
...
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+var (
+	scheme                   = runtime.NewScheme()
+	operatorV1SchemeGV       = schema.GroupVersion{Group: "operators.coreos.com", Version: "v1"}
+	OperatorV1Alpha1SchemeGV = schema.GroupVersion{Group: "operators.coreos.com", Version: "v1alpha1"}
+)
+
+func addKnownTypes(scheme *runtime.Scheme) error {
```
acornett21, line 17: Should this function have a return type since we are only ever returning nil? or is the intent to add more error handling later?

certification/internal/k8s/deployable_by_olm.go
```diff
@@ -0,0 +1,209 @@
...
+
+var (
+	k8sconfig                     *rest.Config
+	err                           error
+	subscriptionName, packageName string
+	channel, ooNamespace          string
+	operatorGroup, catalogSource  string
+	catalogImage                  string
+	targetNamespaces              []string = []string{ooNamespace}
+)
```
komish, line 34: Do these variables need to be package-level variables? 

I see you're setting things up in an initializer function, can we instead make these struct-level variables?

certification/internal/k8s/deployable_by_olm.go
```diff
@@ -0,0 +1,209 @@
...
+			break
+		}
+		// if the context deadline exceeds, fail the check
+		if ctx.Err() != nil {
+			log.Error("failed to fetch .status.installedCSV from Subscription: ", ctx.Err())
+			return false, errors.ErrK8sApiCallFailed
+		}
+		log.Debug("InstalledCSV is not ready yet, retrying...")
+		time.Sleep(2 * time.Second)
+	}
```
komish, line 69: Is the intention here to time out of this for loop after 60 seconds? Should we instead be doing this in a goroutine? 

I know this happens in a couple of places, so I'm wondering if we need to do this differently such that the timeout logic is consistent.

certification/internal/k8s/deployable_by_olm.go
```diff
@@ -0,0 +1,209 @@
...
+func (p *DeployableByOlmCheck) writeToFile(data interface{}, resource string, resourceType string) {
+	yamlData, err := yaml.Marshal(data)
+	if err != nil {
+		log.Error("unable to serialize the subscription")
+	}
+	err = ioutil.WriteFile(fmt.Sprintf("artifacts/%s-%s.yaml", resource, resourceType), yamlData, 0644)
+	if err != nil {
+		log.Error("failed to write the subscription object to the file")
+	}
+}
```
komish, line 189: The error messages indicate that we're only writing a subscription here. Is that the case?

certification/internal/k8s/deployable_by_olm.go
```diff
@@ -0,0 +1,209 @@
...
+		Description:      "Checking if the operator could be deployed by OLM",
+		Level:            "best",
+		KnowledgeBaseURL: "https://connect.redhat.com/zones/containers/container-certification-policy-guide", // Placeholder
+		CheckURL:         "https://connect.redhat.com/zones/containers/container-certification-policy-guide",
+	}
+}
+
+func (p *DeployableByOlmCheck) Help() certification.HelpText {
+	return certification.HelpText{
+		Message:    "It is recommened that your operator could be deployed by OLM",
```
komish, line 206: I think it's _required_ that the operator be deployable by OLM in order to pass certification.

certification/internal/k8s/openshift.go
```diff
@@ -0,0 +1,286 @@
...
+	operatorv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
+	client "github.com/redhat-openshift-ecosystem/openshift-preflight/certification/internal/client"
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/cli"
+)
+
+type OpenshiftCLIEngine struct{}
+
+func (pe OpenshiftCLIEngine) CreateNamespace(name string, opts cli.OpenShiftCliOptions, config *rest.Config) (*cli.OpenshiftCreateReport, error) {
+
+	k8sClientset, err := kubernetes.NewForConfig(config)
```
komish, line 24: Can we instead add an accessor method like `OpenshiftCLIEngine.New()` that accepts the configuration necessary  to build out a client, and then store that client in an internal struct key (or similar)?

cli/openshift.go
```diff
@@ -0,0 +1,57 @@
+package cli
+
+import (
+	operatorv1 "github.com/operator-framework/api/pkg/operators/v1"
+	operatorv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/rest"
+)
+
+type OpenShiftCliOptions struct {
```
komish, line 10: Nit: "Cli" --> "CLI"

certification/internal/k8s/openshift.go
```diff
@@ -0,0 +1,286 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+
+	operatorv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
+	client "github.com/redhat-openshift-ecosystem/openshift-preflight/certification/internal/client"
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/cli"
+)
+
+type OpenshiftCLIEngine struct{}
```
komish, line 20: uber nit!: "Openshift" --> "OpenShift"

cmd/root.go
```diff
@@ -47,9 +48,14 @@ func initConfig() {
 		}
 	}
 
+	if !configFileUsed {
+		log.Info("config file not found, proceeding without it")
+	}
+
```
komish, line 54: I think the placement of this if statement introduces a bug caused by the logger formatter not being configured until later (regression, we've seen this one before IIRC).

certification/internal/client/client.go
samira-barouti: Thank you Adam for taking the time to review my code.

This function is being passed as an argument to XXXXClient in each client file (catalogsource.go, operatorgroup.go, etc). I checked the functions that are being called and they do not have return value (https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#AddToGroupVersion and https://github.com/kubernetes/apimachinery/blob/a644435e2c133d990b624858d9c01985d7f59adf/pkg/runtime/scheme.go#L162).

That being said, none of these function calls return anything and that's why I defined this method to return nothing as well.

komish (changes requested):
A couple of comments, and mostly nits for now. I need to do more review, but before I do - the following points for discussion next week:

I think our interface definitions here are a little too type-specific. I also am not sure we should use the rest client directly if we don't have to. I'd be interested in hearing motives to use it over use something like the dynamic client.

Having CRUD methods for each specific API type can quickly become cumbersome when we _should_ be able to utilize a genericized client that can handle API interactions regardless of the specified type. For context, I'm thinking about the client we use in operator-sdk/kubebuilder projects which can do this, though this might not be the _right_ client buildout plan given our context (https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/client#Client)

certification/engine/engine.go
bcrochet: Carry on. :)

certification/internal/k8s/deployable_by_olm.go
bcrochet: Regardless of it being a goroutine or not, I'd like to see it all moved to a separate function anyway.

certification/internal/k8s/deployable_by_olm.go
```diff
@@ -0,0 +1,209 @@
...
+		time.Sleep(2 * time.Second)
+	}
+
+	log.Trace(fmt.Sprintf("Looking for csv %s in namespace %s", installedCSV, ooNamespace))
+
+	// query API server for CSV by name
+	ctx, cancel = context.WithTimeout(context.Background(), 90*time.Second)
+	defer cancel()
+
+	for {
```
bcrochet, line 77: Same with this block of code. A separate function here would clean things up a bit.

certification/internal/k8s/k8s.go
```diff
@@ -0,0 +1,16 @@
+// package shell contains check implementations that rely on utilizing
+// shell commands directly through the use of cmd.Exec. This implies that the
+// various shell tools are installed.
+package k8s
+
+import "github.com/redhat-openshift-ecosystem/openshift-preflight/cli"
+
+// Create a package-level podmanEngine variable, that can be overridden
```
bcrochet, line 8: openshiftEngine

bcrochet (commented):
A couple of questions...

certification/internal/k8s/deployable_by_olm.go
samira-barouti: Yes. The intention is to dump all created resources in a file before the check exits, and disregard any error.

certification/internal/k8s/deployable_by_olm.go
```diff
@@ -0,0 +1,215 @@
...
+	openshiftEngine.DeleteOperatorGroup(app, cli.OpenShiftOptions{Namespace: ooNamespace}, k8sconfig)
+	openshiftEngine.DeleteNamespace(ooNamespace, cli.OpenShiftOptions{}, k8sconfig)
+}
+
+func (p *DeployableByOlmCheck) writeToFile(data interface{}, resource string, resourceType string) {
+	yamlData, err := yaml.Marshal(data)
+	if err != nil {
+		log.Error("unable to serialize the data")
+	}
+	err = ioutil.WriteFile(fmt.Sprintf("artifacts/%s-%s.yaml", resource, resourceType), yamlData, 0644)
```
bcrochet, line 191: Small nit: Use filepath.Join

certification/internal/k8s/openshift.go
```diff
@@ -0,0 +1,311 @@
...
```
bcrochet, line 20: The CLI part of the name doesn't really fit.

cmd/defaults.go
```diff
@@ -4,4 +4,5 @@ var (
 	DefaultOutputFormat = "json"
 	DefaultLogFile      = "preflight.log"
 	DefaultLogLevel     = "warn"
+	DefaultKubeConfig   = "~/.kube/config"
```
bcrochet, line 7: Is this used anywhere?

bcrochet (changes requested):

cli/openshift.go
```diff
@@ -0,0 +1,57 @@
...
+	Image string
+}
+
+type OperatorGroupData struct {
+	Name             string
+	TargetNamespaces []string
+}
+
+type OpenshiftEngine interface {
+	CreateNamespace(name string, opts OpenShiftOptions, config *rest.Config) (*OpenshiftReport, error)
```
bcrochet, line 40: Here's my feeling on this. Tying the interface to rest.Config makes it less useful as an interface. It seems like that should either be included in OpenShiftOptions, or wrapped in something more "generic".  Otherwise, there will only ever really be a single implementation of this interface, thereby rendering it redundant and unnecessary.

bcrochet, line 40: Also, you have inconsistent case. You are using both Openshift, and OpenShift. I think OpenShift is the proper one to use, as that's the product name as well.

cmd/root.go
```diff
@@ -65,6 +65,7 @@ func initConfig() {
 	}
 
 	log.SetFormatter(&log.TextFormatter{})
+
```
bcrochet, line 68: This is extra. No need to touch this file.

bcrochet (changes requested):

cli/openshift.go
samira-barouti: OpenshiftOptions contains the info for the resource creation (such as namespace and labels). IMO using `opts` to pass along the *rest.Config is dangerous as the client does not need this and the "least privileged principle" will be violated.

Passing `*rest.config` to a method will give the clients the flexibility to communicate with any k8s cluster.

bcrochet: I don't see that we would have a reason to have any other implementation of this than the one that is here. It may be the subject of a refactoring down the road.

bcrochet (approved):

certification/internal/k8s/deployable_by_olm.go
```diff
@@ -0,0 +1,216 @@
...
+		}
+		// if the context deadline exceeds, fail the check
+		if ctx.Err() != nil {
+			log.Error(fmt.Sprintf("failed to fetch the csv: %s ", installedCSV), ctx.Err())
+			return false, errors.ErrK8sAPICallFailed
+		}
+		log.Debug("CSV is not ready yet, retrying...")
+		time.Sleep(2 * time.Second)
+	}
+	return true, nil
```
komish, line 124: Can we restructure this logic such that we explicitly find cases that return `true`, and otherwise return false?

We want to continue to explicitly find success cases, and otherwise return a failure to reduce false positives.

samira-barouti: Fixed...

certification/internal/k8s/deployable_by_olm.go
```diff
@@ -0,0 +1,242 @@
...
+		}
+		// if the context deadline exceeds, fail the check
+		if ctx.Err() != nil {
+			log.Error(fmt.Sprintf("failed to fetch the csv %s: ", installedCSV), ctx.Err())
+			break
+		}
+		log.Debug("CSV is not ready yet, retrying...")
+		time.Sleep(2 * time.Second)
+	}
+	return false, nil
```
komish, line 150: I think we need a select statement in this block, and any others that are leveraging `context.WithTimeout`.

You're effectively running these operations with a 90 second timeout, but you're checking for the timeout evaluation serially, and I think what you wan to do is similar to what is seen in this example, where you run your logic every N seconds using `time.After`, or fail as soon as the context deadline is exceeded.

https://pkg.go.dev/context#example-WithTimeout

samira-barouti: Done.. please let me know if you have any feedback :)

komish (changes requested):
@samira-barouti so sorry for the delay and additional requests for change. Per our discussion, I think there's one more key change here that needs to happen before we can merge.

Right now, I believe this PR expects that the caller provides the index image containing their bundle asset, as well as all the information surrounding the representation of that bundle in the index.

I think instead that we need to take care of that for the user.

The bundle that the user is submitting for test should be ready to add to an index. I would expect that the bundle would be successfully deployable by OLM as the `operator-sdk run bundle` command would do. 

We may even be able to leverage the same container image the `operator-sdk run bundle` command uses to automatically build a quick index image from a given bundle in our tests.

The bundle's `metadata/annotations.yaml` should have all of the package information that the operator registry server should be reading, and that could help inform the subscription that you need.

With this in mind, I think we can reduce the amount of required configuration provided by the user.

The Image Schema V2 validation already extracts the bundle, though the method to do so is bound to the type. This may be a good place for some shared functionality.

https://github.com/redhat-openshift-ecosystem/openshift-preflight/blob/main/certification/internal/shell/has_v2_related_image_schema_version.go#L63-L92

The Package Uniqueness also extracts the bundle, and more specifically, reads the annotations file at the expected location.

https://github.com/redhat-openshift-ecosystem/openshift-preflight/blob/5b4827ac93beb99c8281d3290a541a397a1bea48/certification/internal/shell/operator_pkg_name_uniqueness_mounted.go#L87-L105

Let me know if I can help with anything!

samira-barouti:
@komish I refactored the code to leverage the bundle image to extract package/channel/operator name. The check, however, needs the index image URI which should be passed as PFLT_INDEXIMAGE. I moved some of the common logic to util/container and updated the unique package check. I plan to do the crane stuff in a separate PR as this PR has been pending for so long. Please let me know if you have any feedback.

jomkz (approved):
This looks good to me, pending anything further from @komish.

PR #129, opened 2021-07-26
Fix a couple of linter issues

bcrochet:
The go linter was throwing a couple of warnings. One to simplify a range expression,
and the other had a redundant type specified.

### DIFF
--- a/certification/internal/shell/less_than_max_layers_test.go
+++ b/certification/internal/shell/less_than_max_layers_test.go
@@ -16,7 +16,7 @@ var _ = Describe("LessThanMaxLayers", func() {
 
 	BeforeEach(func() {
 		layers := make([]string, 5)
-		for i, _ := range layers {
+		for i := range layers {
 			layers[i] = fmt.Sprintf("layer%d", i)
 		}
 		fakeEngine = FakePodmanEngine{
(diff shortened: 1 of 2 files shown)


certification/internal/shell/less_than_max_layers_test.go
```diff
@@ -47,7 +47,7 @@ var _ = Describe("LessThanMaxLayers", func() {
 			BeforeEach(func() {
 				engine := fakeEngine.(FakePodmanEngine)
 				layers := make([]string, 50)
-				for i, _ := range layers {
+				for i := range layers {
```
komish, line 50: Below, purely a nit, more a learning opportunity for myself:

I read/heard/learned at some point (from a location that I can't remember now) that this kind of declaration makes it clear that the developers intentionally chose to ignore the value that would otherwise be returned by the right side of the equal sign. I agreed with the premise, though I see it as more valuable in functions that return multiple values where you could potentially not address the value you didn't need.

For this reason, I haven't really made an effort to change these linter complaints. What do you think?

bcrochet: Considering this is in a test, I'm good with fixing this linter complaint. In other code, I might be more explicit.

komish (approved):
/lgtm

samira-barouti (approved):
/lgtm

PR #130, opened 2021-07-26
Remove unused CheckEngine and ContainerFileManager and rename CheckRunner to CheckEngine

komish:
While working through some docs, I noticed that our CheckEngine implementation (Which composed CheckRunner and ContainerFileManager) was unused. I also noticed that ContainerFileManager was not used either; we implement the functionality but we do not bind it to a struct.

Given that ContainerFileManager logic is effectively an implementation detail of the `ExecuteChecks` logic today, and is otherwise not called in packages upstream from its definition (such as `cmd`, where our user-facing logic sits), I concluded that we likely didn't need this abstraction anymore.

As such, this PR removes the CheckEngine and ContainerFileManager interfaces. It renames CheckRunner (which remains) to CheckEngine to be consistent with the term's usage in the package naming.


bcrochet (approved):
/lgtm

samira-barouti (approved):

PR #131, opened 2021-07-26
Increase documentation coverage

komish:
The goal of this PR is just to increase documentation coverage. 

This PR does add a net-new doc on how to implement a check, with a clear note indicating that we may reject a PR for a check given our goal of ensuring this tool accurately reflects the pipeline's validation.

Finally, this adds several comments throughout, including package comments which will show up in pkg.go.dev.

### DIFF
--- a/certification/internal/shell/validate_operator_bundle.go
+++ b/certification/internal/shell/validate_operator_bundle.go
@@ -6,13 +6,15 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
+// ValidateOperatorBundleCheck evaluates the image and ensures that it passes bundle validation
+// as executed by `operator-sdk bundle validate`
 type ValidateOperatorBundleCheck struct {
 }
 
 func (p ValidateOperatorBundleCheck) Validate(bundle string) (bool, error) {
 	report, err := p.getDataToValidate(bundle)
 	if err != nil {
-		log.Error("Error will executing operator-sdk validate bundle: ", err)
+		log.Error("Error while executing operator-sdk bundle validate: ", err)
 		return false, err
 	}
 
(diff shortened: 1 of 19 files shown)


samira-barouti (approved):

komish:
An extra note, this also removes the `CheckInfo` struct which was previously used in user responses, but we've since standardized on that response and so this CheckInfo struct is no longer in use.

certification/internal/shell/validate_operator_bundle.go
```diff
@@ -6,13 +6,15 @@ import (
...
+// as executed by `operator-sdk bundle validate`
 type ValidateOperatorBundleCheck struct {
 }
 
 func (p ValidateOperatorBundleCheck) Validate(bundle string) (bool, error) {
 	report, err := p.getDataToValidate(bundle)
 	if err != nil {
-		log.Error("Error will executing operator-sdk validate bundle: ", err)
+		log.Error("Error will executing operator-sdk bundle validate: ", err)
 		return false, err
```
tonytcampbell, line 18: Typo:  "Error while executing....."

komish: Good catch, fixed!

docs/dev/README.md
```diff
@@ -5,91 +5,27 @@ operator projects to see if they pass Red Hat operator certification
 requirements.
 
 The project will include a commandline interface that will accept your operator
-bundle image as an input, and will run validate that your operator bundle
-complies with a configurable set of checks.
+bundle or container image as an input, and will run validate that your container
+image or operator bundle complies with a series of validations.
```
tonytcampbell, line 9: is validation a synonym of check at the end of Line 9?  I think we should be consistent with out terms and point out if we are talking about something different than checks.

tonytcampbell (commented):
Few nits, but looks good outside of that.

bcrochet (approved):
/lgtm

PR #134, opened 2021-07-27
Use boolean value to indicate overall result

komish:
The endpoint that will receive preflight results would like to use a boolean flag at a top-level `passed` key to indicate the overall result of policy execution.

This PR adjusts the value from a string to a boolean, and also changes the name of the key from `status` to `passed` so that a boolean value makes more sense.

We also align the Result struct to reflect similar values, and update relevant comments to match the new definition.


bcrochet (approved):
/lgtm

samira-barouti (approved):

PR #136, opened 2021-07-28
Implement package uniqueness checks

komish:
Extending original PR #132 (thanks @rocrisp for doing the leg work on this check!)
Fixes: #34 

## Context
This PR refactors the Package Uniqueness check, which has a goal of hitting an external API with the package name stored in the bundle's ClusterServiceVersion as a query string, and confirming that the API does not have an entry for the given package already.

## Summary
- This check uses the mounted engine approach, and mounts the bundle on the filesystem using `podman unshare`. It then  reads the `annotations.yaml` file from disk and extracts the appropriate key/value that contains the package name.
- The package name is then sent to the defined API with the appropriate query string
- The API's response body is parsed/extracted to determine if a package with the name already exists. It's possible this might change if our target API changes.
- This PR also adds an apiClient interface just to simplify testing against a fake client.

PR #139, opened 2021-07-30
Add image mounting to podman engine

### LINKED ISSUE #137: Extend Podman interfaces to allow mounting an image instead of a container in unshare-based checks.
Yes, when running checks for operator bundles, which are non-runnable container, I must fake-create the container in order to extract the bundle and perform my check.

Example: #124

Instead, it would be ideal if we could mount the container image without running the container at all.

#126 implemented similar functionality for containers. A PR resolving this issue would have to expand on that, allowing for "running in a filesystem" to work for either containers or images. 

This would likely mean expanding the podmanEngine interface to allow for mounting an image.

The alternative is implemented in #124 as an example, and described above in the related problem description. We must create the container to extract resources from that filesystem, and then we must clean up after ourselves.

komish:
Fixes #137 
Unblocks #136 

## Context

PR #126 added the ability to mount a container image so that we can directly inspect the container filesystem without having to extract files or exec _into_ the container.

This still requires creating a container, however, which doesn't apply cleanly to operator bundle checks which are non-runnable. We _can_ "run" an operator bundle by faking the entry point, and there is an existing check that does this. With that said, if we could mount the image similarly to how we mount containers, we would no longer need to extract the bundle to disk.

This PR extends the PodmanEngine to have a (Un)MountImage functionality to solve this problem.

## Summary

- Extends `PodmanEngine`, `FakePodmanEngine`, `BadPodmanEngine`, and the related interface to have MountImage and UnmountImage methods.
- ~Introduces an environment variable PREFLIGHT_EXEC_TYPE for the `check run` hidden command that allows callers to identify the _type_ of asset being tested in sub-execution.~ See [design update](https://github.com/redhat-openshift-ecosystem/openshift-preflight/pull/139#issuecomment-891120951).
- This PREFLIGHT_EXEC_TYPE value is only used in the mounted engine, and causes the mounted engine to leverage a new helper `RunInsideImageFS`, similar to `RunInsideContainerFS`, in cases where the asset is not a container and therefore doesn't need to be created.
- Extended the PodmanEngine.UnshareWithCheck definition to also accept the image type so that it exports the new environment variable mentioned above.
- Fixed a small bug in our engine code, caused by overloading the variable name `passed` which caused our log output to spit out result=true instead of result=passed

In my testing, the only check utilizing this passes as expected, and this enables the Package Name Uniqueness check which is in the works.

### DIFF
--- a/certification/internal/shell/mounted_engine.go
+++ b/certification/internal/shell/mounted_engine.go
@@ -26,17 +26,14 @@ func (e *MountedCheckEngine) ExecuteChecks() error {
 	log.Info("running check: ", e.Check.Name())
 	// We want to know the time just for the check itself, so reset checkStartTime
 	checkStartTime := time.Now()
-
-	// run the validation
-	pass, err := containerutil.RunInsideContainerFS(podmanEngine, targetImage, e.Check.Validate)
-
+	checkPassed, err := containerutil.RunInsideImageFS(podmanEngine, targetImage, e.Check.Validate)
 	checkElapsedTime := time.Since(checkStartTime)
 
 	switch {
 	case err != nil:
 		log.WithFields(log.Fields{"result": "ERROR", "error": err.Error()}).Info("check completed: ", e.Check.Name())
 		e.results.Errors = append(e.results.Errors, runtime.Result{Check: e.Check, ElapsedTime: checkElapsedTime})
-	case !pass:
+	case !checkPassed:
 		log.WithFields(log.Fields{"result": "FAILED"}).Info("check completed: ", e.Check.Name())
 		e.results.Failed = append(e.results.Failed, runtime.Result{Check: e.Check, ElapsedTime: checkElapsedTime})
 	default:
--- a/certification/internal/shell/engine.go
+++ b/certification/internal/shell/engine.go
@@ -56,7 +56,7 @@ func (e *CheckEngine) ExecuteChecks() error {
 		checkStartTime := time.Now()
 
 		// run the validation
-		passed, err := check.Validate(targetImage)
+		checkPassed, err := check.Validate(targetImage)
 
 		checkElapsedTime := time.Since(checkStartTime)
 
@@ -66,7 +66,7 @@ func (e *CheckEngine) ExecuteChecks() error {
 			continue
 		}
 
-		if !passed {
+		if !checkPassed {
 			log.WithFields(log.Fields{"result": failed}).Info("check completed: ", check.Name())
 			e.results.Failed = append(e.results.Failed, runtime.Result{Check: check, ElapsedTime: checkElapsedTime})
 			continue
--- a/certification/internal/shell/podman.go
+++ b/certification/internal/shell/podman.go
@@ -223,6 +223,7 @@ func (pe PodmanCLIEngine) Remove(containerID string) (*cli.PodmanRemoveReport, e
 	}, nil
 }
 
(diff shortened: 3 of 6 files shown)


certification/internal/shell/mounted_engine.go
```diff
@@ -27,16 +28,21 @@ func (e *MountedCheckEngine) ExecuteChecks() error {
 	// We want to know the time just for the check itself, so reset checkStartTime
 	checkStartTime := time.Now()
 
-	// run the validation
-	pass, err := containerutil.RunInsideContainerFS(podmanEngine, targetImage, e.Check.Validate)
+	// If the target is a bundle, we need to run the image filesystem mount logic instead
+	// of the container filesystem mount logic.
+	mountFn := containerutil.RunInsideContainerFS
+	if e.TargetIsBundle {
+		mountFn = containerutil.RunInsideImageFS
```
bcrochet, line 35: Is this distinction even necessary? Is there an advantage to using the container mount over the image mount for non-bundles? If we can get away with just using the image mount everywhere, then I would just switch the existing mount logic to use the image mount.

bcrochet (changes requested):
Overall, I don't see any problems at a quick glance, but I have a fundamental question about the direction of the PR.

certification/internal/shell/mounted_engine.go
komish: @bcrochet Okay. I'll make the switch and repush. More to come as I have it.

komish: @bcrochet Done!

komish:
## Update

This PR now replaces the `RunInsideContainerFS` command with only the `RunInsideImageFS`, and replaces any location where the container FS mounting logic was used to utilize image mounting logic instead.

As a result, we no longer need toggles identifying the input asset as a bundle vs. a container.

certification/internal/shell/mounted_engine.go
```diff
@@ -27,16 +27,17 @@ func (e *MountedCheckEngine) ExecuteChecks() error {
 	// We want to know the time just for the check itself, so reset checkStartTime
 	checkStartTime := time.Now()
 
-	// run the validation
-	pass, err := containerutil.RunInsideContainerFS(podmanEngine, targetImage, e.Check.Validate)
+	// If the target is a bundle, we need to run the image filesystem mount logic instead
```
bcrochet, line 30: This comment doesn't seem to make sense anymore.

bcrochet (changes requested):

certification/internal/shell/mounted_engine.go
komish: @bcrochet fixed

bcrochet (approved):
/lgtm

bcrochet:
>     * Introduces an environment variable PREFLIGHT_EXEC_TYPE for the `check run` hidden command that allows callers to identify the _type_ of asset being tested in sub-execution.
> 
>     * This PREFLIGHT_EXEC_TYPE value is only used in the mounted engine, and causes the mounted engine to leverage a new helper `RunInsideImageFS`, similar to `RunInsideContainerFS`, in cases where the asset is not a container and therefore doesn't need to be created.

This is not the case anymore. Up to you if you want to edit this description, but make sure to update the commit message.

komish:
@bcrochet I don't think my commit message has those notes, but I've added a link to the update comment in the original verbiage.

samira-barouti (approved):
/lgtm

PR #141, opened 2021-08-02
Run minimal vulnerabilities check in an unshare environment

### LINKED ISSUE #49: Reduce elevated privileges required to run preflight successfully when testing against a container image

bcrochet:
Besides making the minimal vulnerabilities check to run in an
unshare, this patch also changes the UnshareWithCheck signature to
simplify it, so that the "preflight check run" command does not have
to be specified each time, but it also specifies whether the check
should have the passed image mounted, then run the check, or just
have the check run in the unshare env.

Fixes #49

### DIFF
--- a/certification/internal/shell/has_minimal_vulns_unshare.go
+++ b/certification/internal/shell/has_minimal_vulns_unshare.go
@@ -1,0 +1,75 @@
+package shell
+
+import (
+	"fmt"
+	"regexp"
+	"strings"
+
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification"
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/errors"
+	log "github.com/sirupsen/logrus"
+)
+
+// HasMinimalVulnerabilitiesUnshareCheck evaluates the image to confirm that only vulnerabilities
+// below the Critical or Important severities, using oscap-podman and a corresponding
+// OVAL definition.
+type HasMinimalVulnerabilitiesUnshareCheck struct{}
+
+func (p *HasMinimalVulnerabilitiesUnshareCheck) Validate(image string) (bool, error) {
+	lines, err := p.scanImage(image)
+	if err != nil {
+		log.Debugf("Stdout: %s", lines)
+		return false, fmt.Errorf("%w: %s", errors.ErrImageScanFailed, err)
+	}
+
+	return p.validate(lines)
+}
+
+func (p *HasMinimalVulnerabilitiesUnshareCheck) scanImage(image string) ([]string, error) {
+	imageScanReport, err := podmanEngine.ScanImage(image)
+	if err != nil {
+		log.Error("unable to scan the image: ", err)
+		return nil, err
+	}
+
+	return strings.Split(imageScanReport.Stdout, "\n"), nil
+}
+
+func (p *HasMinimalVulnerabilitiesUnshareCheck) validate(lines []string) (bool, error) {
+
+	// oscap-podman writes `Definition oval:com.redhat.<CVE#>: true` to stdout
+	// if the vulnerability exist in the container, and `Definition oval:com.redhat.<CVE#>: false`
+	// if the vulnerability does not exist
+	r := regexp.MustCompile("Definition oval:com.redhat.*: true")
+
+	numOfVulns := 0
+	for _, line := range lines {
+		if r.MatchString(line) {
+			numOfVulns++
+		}
+	}
+	// count the number of matches
+	log.Debugf("The number of found vulnerabilities: %d", numOfVulns)
+
+	return numOfVulns == 0, nil
+}
+
+func (p *HasMinimalVulnerabilitiesUnshareCheck) Name() string {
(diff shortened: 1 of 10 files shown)


certification/internal/shell/has_minimal_vulns_unshare.go
```diff
@@ -0,0 +1,75 @@
...
+	"fmt"
+	"regexp"
+	"strings"
+
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification"
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/errors"
+	log "github.com/sirupsen/logrus"
+)
+
+// HasMinimalVulnerabilitiesCheck evaluates the image to confirm that only vulnerabilities
```
komish, line 13: `HasMinimalVulnerabilitiesCheck` --> `HasMinimalVulnerabilitiesUnshareCheck`

certification/internal/shell/has_minimal_vulns_unshare.go
```diff
@@ -0,0 +1,75 @@
...
+// below the Critical or Important severities, using oscap-podman and a corresponding
+// OVAL definition.
+type HasMinimalVulnerabilitiesUnshareCheck struct{}
+
+func (p *HasMinimalVulnerabilitiesUnshareCheck) Validate(image string) (bool, error) {
+	lines, err := p.getDataToValidate(image)
```
komish, line 19: Nit: I somewhat dislike `getDataToValidate` as a method name because it isn't descriptive. I don't have a suggestion at the moment, and I know we use it in multiple places, but If you can think of a way to change this so that we can somewhat establish a shared context with the reader/developer reviewing or attempting to follow this code, that'd be great.

No need to change others, we can change them as we come across them.

certification/internal/shell/has_minimal_vulns_unshare.go
```diff
@@ -0,0 +1,75 @@
...
+	if err != nil {
+		log.Debugf("Stdout: %s", lines)
+		return false, fmt.Errorf("%w: %s", errors.ErrImageScanFailed, err)
+	}
+
+	return p.validate(lines)
+}
+
+func (p *HasMinimalVulnerabilitiesUnshareCheck) getDataToValidate(image string) ([]string, error) {
```
komish, line 28: I know it's not strictly required, but I believe we should start commenting these.

certification/internal/shell/podman.go
```diff
@@ -284,14 +284,18 @@ func (pe PodmanCLIEngine) Unshare(env map[string]string, command ...string) (*cl
...
 	env := map[string]string{
 		"PATH":                 os.Getenv("PATH"),
 		"PREFLIGHT_EXEC_CHECK": check,
 		"PREFLIGHT_EXEC_IMAGE": image,
 	}
 
-	unshareReport, err := pe.Unshare(env, command...)
+	if mounted {
+		env["PREFLIGHT_EXEC_MOUNTED"] = fmt.Sprintf("%t", mounted)
+	}
```
komish, line 296: Why are we tracking the mounted state for this function? Is there logic somewhere that utilizes `PREFLIGHT_EXEC_MOUNTED` for some purpose?

I see it was added to the runtime.Config as well; trying to get a more complete picture but I'm missing it reviewing this.

komish (commented):
Couple of questions and nits. LGTM for the most part, but I'll do some testing after I get more context.

EDIT: I hit "Comment" instead of "Request Changes", not sure how to change it so I'm leaving this as is!

certification/internal/shell/has_minimal_vulns_unshare.go
bcrochet: Fixed.

certification/internal/shell/has_minimal_vulns_unshare.go
bcrochet: I'll change this one to "scanImage", as that describes exactly what it is doing.

certification/internal/shell/podman.go
bcrochet: This particular check does not require a mounted image. In fact, it will fail if it has one. So, I've added the ability for a check to designate whether it needs to have a mounted image or not. The environment variable then allows the "check run" command to signify whether it should instantiate a MountedEngine, or just a CheckEngine.

jomkz (approved):
/lgtm as I believe all of the comments have been addressed

komish (approved):
/lgtm

PR #145, opened 2021-08-03
Add generation of a "bundle hash"

### LINKED ISSUE #60: The preflight tool should generate hashes based on the bundle directory
https://docs.google.com/document/d/1AVSIJCdXvBk2LzvnkXZ1PTtJudVDxYQuJ-YmT7cF2_U/edit describes how a set of hashes can be generated to identify a specific bundle.

bcrochet:
This creates a hashfile containing the md5sum output of the files contained
in the bundle image. This is not meant to be a secure hash. It is meant
to be used for identifying a particular bundle within the pipeline.

Fixes #60

### DIFF
--- a/certification/internal/utils/container/helper.go
+++ b/certification/internal/utils/container/helper.go
@@ -1,6 +1,7 @@
 package shell
 
 import (
+	"crypto/md5"
 	"fmt"
 	"os"
 	"os/exec"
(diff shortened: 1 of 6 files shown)


komish:
@bcrochet was there a requirement that enforces that the bundle hash must live in its own file? I'm wondering if this is something we should include in the Result. I realize that this may be a tricky task to accomplish given the current implementation of things.

If it needs to be a standalone file should we put this in the artifacts directory? I believe the system receiving our data is expecting any additional files, outside of the preflight log and the results JSON, to exist in the artifacts directory.

bcrochet:
> @bcrochet was there a requirement that enforces that the bundle hash must live in its own file? I'm wondering if this is something we should include in the Result. I realize that this may be a tricky task to accomplish given the current implementation of things.
> 
> If it needs to be a standalone file should we put this in the artifacts directory? I believe the system receiving our data is expecting any additional files, outside of the preflight log and the results JSON, to exist in the artifacts directory.

The requirement that was asked for was actually a hash of the hash file. We could just return that hash, as that is what is expected. As for artifacts, sure. I can move it there.

certification/internal/utils/container/helper.go
```diff
@@ -97,3 +98,24 @@ func RunInsideContainerFS(podmanEngine cli.PodmanEngine, image string, container
 
 	return containerFn(report.MountDir)
 }
+
+func GenerateBundleHash(podmanEngine cli.PodmanEngine, image string) (string, error) {
+	hashCmd := `find . -not -name "Dockerfile" -type f -printf '%f\t%p\n' | sort -V -k1 | cut -d$'\t' -f2 | tr '\n' '\0' | xargs -r0 -I {} md5sum "{}"` // >> $HOME/hashes.txt`
```
samira-barouti, line 103: It would be great to define this as a const on top and not recreate it every time this function is called.

bcrochet: It's only called one time. And since it is basically a constant, the compiler probably optimizes for that anyway.

bcrochet:
> > The requirement that was asked for was actually a hash of the hash file. We could just return that hash, as that is what is expected. As for artifacts, sure. I can move it there.
> 
> I think we've got some logistics to work out here, but it may be a function of the `certify` subcommand which is still undefined. 🤔 For now, though, having it placed in `artifacts/` should suffice.

Yes. There is a good chance that this ends up just being called in certify.

komish (approved):
/lgtm

jomkz (approved):
/lgtm

PR #148, opened 2021-08-04
Represent elapse time as a float

### LINKED ISSUE #138: Adjust elapsed_time type in UserResponse struct to be a float representing milliseconds
This is a `time.Duration` in our `Results` struct(1), but we represent this as a string to the user(2). We've received a request to represent this as a JSON number, and more specifically, milliseconds as a float. 

(1): https://github.com/redhat-openshift-ecosystem/openshift-preflight/blob/main/certification/runtime/runtime.go#L20
(2): https://github.com/redhat-openshift-ecosystem/openshift-preflight/blob/ff73edd14f29ba48379feb8ce760886fafda7183/certification/formatters/util.go#L81-L89

edreyer-1:
We want elapsed time to represented as a float in milliseconds instead of a string.

Fixes #138

### DIFF
--- a/certification/formatters/util.go
+++ b/certification/formatters/util.go
@@ -16,7 +16,7 @@ func getResponse(r runtime.Results) UserResponse {
 		for i, check := range r.Passed {
 			passedChecks[i] = checkExecutionInfo{
 				Name:        check.Name(),
-				ElapsedTime: check.ElapsedTime.String(),
+				ElapsedTime: float64(check.ElapsedTime.Milliseconds()),
 				Description: check.Metadata().Description,
 			}
 		}
@@ -26,7 +26,7 @@ func getResponse(r runtime.Results) UserResponse {
 		for i, check := range r.Failed {
 			failedChecks[i] = checkExecutionInfo{
 				Name:             check.Name(),
-				ElapsedTime:      check.ElapsedTime.String(),
+				ElapsedTime:      float64(check.ElapsedTime.Milliseconds()),
 				Description:      check.Metadata().Description,
 				Help:             check.Help().Message,
 				Suggestion:       check.Help().Suggestion,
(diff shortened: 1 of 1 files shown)


certification/formatters/util.go
```diff
@@ -16,7 +18,7 @@ func getResponse(r runtime.Results) UserResponse {
 		for i, check := range r.Passed {
 			passedChecks[i] = checkExecutionInfo{
 				Name:        check.Name(),
-				ElapsedTime: check.ElapsedTime.String(),
+				ElapsedTime: float64(check.ElapsedTime) / float64(time.Millisecond),
```
bcrochet, line 21: Since check.ElapsedTime is a time.Duration, you can just call check.ElapsedTime.Milliseconds().

certification/formatters/util.go
```diff
@@ -26,7 +28,7 @@ func getResponse(r runtime.Results) UserResponse {
 		for i, check := range r.Failed {
 			failedChecks[i] = checkExecutionInfo{
 				Name:             check.Name(),
-				ElapsedTime:      check.ElapsedTime.String(),
+				ElapsedTime:      float64(check.ElapsedTime) / float64(time.Millisecond),
```
bcrochet, line 31: Ditto.

certification/formatters/util.go
```diff
@@ -40,7 +42,7 @@ func getResponse(r runtime.Results) UserResponse {
 		for i, check := range r.Errors {
 			erroredChecks[i] = checkExecutionInfo{
 				Name:        check.Name(),
-				ElapsedTime: check.ElapsedTime.String(),
+				ElapsedTime: float64(check.ElapsedTime) / float64(time.Millisecond),
```
bcrochet, line 45: Ditto.

bcrochet (changes requested):
Small change, other than that, looks good.

certification/formatters/util.go
edreyer-1: I tried that but Milliseconds() returns an int and since we wanted a float I didn't use it.

jomkz: We should be able to cast that to float64 like you are doing already no?

`float64(check.ElapsedTime.Milliseconds())`

edreyer-1: Yes I could do that. The only thing with doing this is we lose some precision after the decimal point but if we don't care about that then it should be fine.

komish: @edreyer-1 I'm okay with losing that precision for this use case 👍

bcrochet (approved):
/lgtm

komish (approved):
/lgtm

PR #150, opened 2021-08-04
Inject the PodmanEngine into BasedOnUBICheck

bcrochet:
This does a few things:

* Creates a PodmanEngine implementation in certification/engine package
* Adds the PodmanEngine field to the BaseOnUBICheck struct
* Creates a NewBasedOnUBICheck "constructor" that access a PodmanEngine
* Modifies the tests to use the constructor, and set PodmanEngine
  accordingly

This negates the need for the podmanEngine package-level variable. It
also makes the unit tests more self-contained.


komish:
@bcrochet Given the current direction, should we close this?

bcrochet:
Now obsolete.

PR #152, opened 2021-08-05
Make some changes to assist with Podman in Podman

### LINKED ISSUE #109: Pulling and unpacking an image fails when running Preflight as a container
## Bug Description

Pulling and unpacking an image using `podmanEngine` fails when running Preflight as a container.

## Steps to Reproduce:

`docker run --env 'PFLT_LOGLEVEL=trace' quay.io/opdev/preflight:<my-local-tag>  check operator quay.io/itroyano/conjur-operator-bundle:v0.0.2`

`time="2021-07-19T12:58:51Z" level=error msg="unable to pull the container from the registry: failed to pull remote container: exit status 125"`

also

`docker run -it --entrypoint podman quay.io/opdev/preflight:<my-local-tag>  pull quay.io/itroyano/conjur-operator-bundle:v0.0.2`

`Error: 'overlay' is not supported over overlayfs, a mount_program is required: backing file system is unsupported for this graph driver`

## Additional Context

OS: Mac (Darwin x64 arch)
Reference: https://www.redhat.com/sysadmin/podman-inside-container

bcrochet:
* Check if user is UID == 0 in unshare, and just call command if it is
* Change base image to quay.io/podman/stable
* Add findutils to image build
* Fix Makefile so that IMAGE_REPO and IMAGE_BUILDER can be overridden

Fixes #109 


Dockerfile
```diff
@@ -7,7 +7,7 @@ WORKDIR /go/src/preflight
 RUN make build
 
 # ubi-minimal:latest image created 2021-06-22T13:04:30.956781
```
jomkz, line 9: I added this comment originally to make it easier to determine what/when the digest was for, but up to everyone if we want to keep this, as I can see it getting out of date or forgotten when we upgrade the version we are using here.

jomkz (changes requested):

Dockerfile
bcrochet: Removed for now.

jomkz (approved):
/lgtm

samira-barouti (approved):

PR #155, opened 2021-08-09
Update sha tag to use v3.2.2

bcrochet:
The v3.2.3 tag keeps getting rebuilt and tagged the same as latest. This
drops the version back one, as that sha is actually stable. Update when
podman automation is fixed.


acornett21 (approved):

jomkz (approved):
/lgtm

samira-barouti (approved):

PR #156, opened 2021-08-10
Move RelateImagesAreSchemaVersion2 test to policy/operator package

### LINKED ISSUE #149: Check for Related Images Schema Version 2 should use Podman Unshare and the Mounted Engine
## Bug Description
We currently end up with an `extracted-bundle` directory that is coming from the operator check called `RelatedImagesAreSchemaVersion2`. 

This check currently "fake" runs the bundle in order to copy out the bundle to the caller filesystem. 

`preflight check operator <operator-bundle>`

## Steps to Reproduce:
Run the above against a bundle of your choosing, and observe that the `extracted-bundle` directory exists on disk

## Expected Result
This check should instead leverage the mounted engine that is being used by other checks so that no assets have to be extracted from the bundle onto the host filesystem.

## Actual Result
Preflight extracts the bundle, and it also extracts it with the container's privileges and so frequently the host filesystem cannot delete the bundle without root privileges.

bcrochet:
The unit test for RelatedImages check needed to be migrated to the new package.

Fixes #149

### DIFF
--- a/certification/internal/policy/operator/has_v2_related_image_schema_version_test.go
+++ b/certification/internal/policy/operator/has_v2_related_image_schema_version_test.go
@@ -1,0 +1,113 @@
+package operator
+
+import (
+	operatorsv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("RelatedImagesAreSchemaVersion2", func() {
+	check := RelatedImagesAreSchemaVersion2Check{}
+
+	// tests check.validate(...) - this check does not currently have a case
+	// that throws an error, so no test case exists for it.
+	Context("When validating a mapping of images to their image manifest schema versions", func() {
+		Context("with images that all have image manifests in schema version 2", func() {
+			goodImageSchemaVersMap := map[string]int{
+				"dummy/image1": 2,
+			}
+
+			It("should pass validation", func() {
+				passed, err := check.validate(goodImageSchemaVersMap)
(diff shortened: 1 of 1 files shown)


certification/internal/shell/has_v2_related_image_schema_version.go
```diff
@@ -19,192 +12,16 @@ type RelatedImagesAreSchemaVersion2Check struct{}
...
-// checks the ClusterServiceVersion's related images declaration, and assembles a list of
-// images and their image manifest schema version values.
-func (p *RelatedImagesAreSchemaVersion2Check) getDataToValidate(image string) (map[string]int, error) {
-	bundlePath, err := p.extractBundleFromImage(image)
-	if err != nil {
-		log.Error(err)
-		return nil, fmt.Errorf("%w: %s", errors.ErrDeterminingRelatedImageSchemaVers, err)
-	}
-	csv, err := p.readBundle(bundlePath)
+	mounted := true
```
komish, line 15: Should we just pass `true` into `UnshareWithCheck` and leave out the variable assignment?

komish (approved):
Left a standalone comment. Thanks for taking care of this.

/lgtm

certification/internal/shell/has_v2_related_image_schema_version.go
bcrochet: I don't like bare booleans. It makes it explicit what the called function is expecting. And it shouldn't affect performance in any way, as I'm quite sure the compiler would just optimize this out.

komish: Ah, I can get behind that. 👍

PR #157, opened 2021-08-10
updating vagrant file to include crun

acornett21:
Adding `crun` to vagrant so that `make build-image` does not throw any errors and completes successfully.


samira-barouti (approved):
/lgtm

bcrochet (approved):

PR #159, opened 2021-08-11
Make sure logs are written to the artifacts/ dir

bcrochet:
In order to make sure that logs are persisted in a containerized
run, this writes the logs to a relative dir of 'artifacts/'.

It is assumed that the workdir for the container is '/'

Also, multiple invocations of unshare will overwrite the logfile
of a previous run. This will separate them out.


acornett21 (approved):

komish:
@bcrochet should we consider some kind of shared function all checks can utilize to write artifacts to the same location?

bcrochet:
> @bcrochet should we consider some kind of shared function all checks can utilize to write artifacts to the same location?

That's probably a good play. It probably also needs to be a config setting, tbh.

samira-barouti (approved):
/lgtm

samira-barouti (approved):
/lgtm

komish (approved):
/lgtm

PR #163, opened 2021-08-12
list only files greater than minLicenseSize

### LINKED ISSUE #161: License check should validate file size
Not only should preflight check for file or files in /licenses it should check that those files are not zero length, which would be a useless license.

skattoju:
fixes #161 
- now checking if the license folder contains files larger than 1 byte
- using find instead of ls

### DIFF
--- a/certification/internal/shell/has_license.go
+++ b/certification/internal/shell/has_license.go
@@ -13,6 +13,8 @@ const (
 	licensePath         = "/licenses"
 	newLine             = "\n"
 	minLicenseFileCount = 1
+	//size format string for find command, c denotes bytes
+	minLicenseSize = "+1c"
 )
 
 // HasLicenseCheck evaluates that the image contains a license definition available at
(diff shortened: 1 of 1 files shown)


samira-barouti (approved):
/lgtm
Out of curiosity: is this how this check is being done in the Ansible playbooks?

skattoju:
not sure could not find the ansible version of the check
https://github.com/redhat-openshift-ecosystem/operator-test-playbooks

samira-barouti:
@skattoju IMO we better check for a keyword (such as License and ignoring the cases) in the files to ensure the files do not contain random contents.

skattoju:
I think we said we don't want to get into verifying the contents as this is
enforced by legal agreements. They could always create a dummy license
with the word "license" right ?

bcrochet:
> I think we said we don't want to get into verifying the contents as this is enforced by legal agreements. They could always create a dummy license with the word "license" right ?

That is correct. We are at least checking that they put *something* in a file in /licenses. We just want to protect someone touching a file in /licenses, trying to get this to pass, then to forget to go back and put something in.

PR #168, opened 2021-08-16
Implement support command

acornett21:
- fixed some spelling mistakes in the project
- implemented support command to generate/output a support ticket url for a user to copy/paste into a browser

note: url package was not used to generate the URL, since the base URL has a '#' in it, which was also getting encoded along with the query params.

### DIFF
--- a/cmd/support.go
+++ b/cmd/support.go
@@ -1,18 +1,104 @@
 package cmd
 
 import (
+	"errors"
 	"fmt"
+	"net/url"
+	"regexp"
 
+	"github.com/manifoldco/promptui"
+	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
+)
+
+const (
+	baseURL              = "https://connect.redhat.com/support/technology-partner/#/case/new?"
+	typeParam            = "type"
+	typeValue            = "CERT"
+	sourceParam          = "source"
+	sourceValue          = "preflight"
+	certProjectTypeParam = "cert_project_type"
+	certProjectIDParam   = "cert_project_id"
+	pullRequestURLParam  = "pull_request_url"
+	promptErrorMessage   = "Prompt Failed, Please Try re-running support command."
 )
 
 var supportCmd = &cobra.Command{
 	Use:   "support",
 	Short: "Submits a support request",
-	Long: `This command will submit a support request to Red Hat along with the logs from the latest Preflight checck.
+	Long: `This interactive command will generate a URL; based on user input which can then be used to create a Red Hat Support Ticket.
 	This command can be used when you'd like assistance from Red Hat Support when attempting to pass your certification checks. `,
-	Run: func(cmd *cobra.Command, args []string) {
-		fmt.Println("Support command not implemented")
+	RunE: func(cmd *cobra.Command, args []string) error {
+
+		certProjectTypeLabel := promptui.Select{
+			Label: "Select a Certification Project Type",
+			Items: []string{"Container Image", "Operator Bundle Image"},
+		}
+
+		_, certProjectTypeValue, err := certProjectTypeLabel.Run()
+
+		if err != nil {
+			fmt.Println(promptErrorMessage)
+			return err
+		}
+
+		log.Debugf("certification project type: %s", certProjectTypeValue)
+
+		certProjectIDLabel := promptui.Prompt{
+			Label: "Please Enter Connect Certification Project ID",
+			Validate: func(s string) error {
+				isLegacy, _ := regexp.MatchString(`^p.*`, s)
+				if isLegacy {
+					return errors.New("please remove leading character p from project id")
+				}
+
+				isOSPID, _ := regexp.MatchString(`^ospid-.*`, s)
+				if isOSPID {
+					return errors.New("please remove leading character ospid- from project id")
+				}
+
+				return nil
+			},
+		}
+
+		certProjectIDValue, err := certProjectIDLabel.Run()
+
+		if err != nil {
+			fmt.Println(promptErrorMessage)
+			return err
+		}
+
+		log.Debugf("certification project id: %s", certProjectIDValue)
+
+		pullRequestURLLabel := promptui.Prompt{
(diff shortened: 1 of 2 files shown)


cmd/support.go
```diff
@@ -2,17 +2,81 @@ package cmd
...
+	certProjectIDParam   = "cert_project_id"
+	pullRequestURLParam  = "pull_request_url"
+)
+
 var supportCmd = &cobra.Command{
 	Use:   "support",
 	Short: "Submits a support request",
-	Long: `This command will submit a support request to Red Hat along with the logs from the latest Preflight checck.
+	Long: `This command will submit a support request to Red Hat along with the logs from the latest Preflight check.
 	This command can be used when you'd like assistance from Red Hat Support when attempting to pass your certification checks. `,
```
komish, line 26: We need to reword this. This command now returns a "fast-track" way to create a support request, but doesn't actually submit the support request because it requires authentication, correct?

cmd/support.go
```diff
@@ -2,17 +2,81 @@ package cmd
...
+
+		// building and encoding query params
+		queryParams := url.Values{}
+		queryParams.Add(typeParam, typeValue)
+		queryParams.Add(sourceParam, sourceValue)
+		queryParams.Add(certProjectTypeParam, certProjectTypeValue)
+		queryParams.Add(certProjectIDParam, certProjectIDValue)
+		queryParams.Add(pullRequestURLParam, pullRequestURLValue)
+
+		fmt.Printf("URL to Create Support Desk Ticket: %q\n", baseURL+queryParams.Encode())
```
komish, line 77: Nit: can we space and tab this so it spits out something like...

```
URL to create a support ticket
        https://...
```

Also a nit, but can we change this verbiage to b e a bit more instructional, like "Create a support ticket by..."

komish (changes requested):
Couple of nits

cmd/support.go
komish: You can word this however you think makes sense, but the current wording doesn't seem right to me.

acornett21: You are correct, this was the existing skeleton code, but because of sso login requirement this will only print out a URL for the user to copy/paste into a browser and then login to be able to submit. I'll think about a more accurate wording.

cmd/support.go
acornett21: This has been updated to list the steps.

cmd/support.go
acornett21: This has been updated as well.

samira-barouti (approved):

skattoju (commented):
you might need to rebase

cmd/support.go
```diff
@@ -2,17 +2,86 @@ package cmd
...
 import (
 	"fmt"
+	"net/url"
 
+	"github.com/manifoldco/promptui"
 	"github.com/spf13/cobra"
 )
 
+const (
+	baseURL              = "https://connect.dev.redhat.com/support/technology-partner/#/case/new?"
```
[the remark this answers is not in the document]
samira-barouti: I believe we should leverage viper for the environment-specific configurations.

cmd/support.go
```diff
@@ -1,18 +1,102 @@
...
+			fmt.Println("Prompt Failed, Please Try re-running support command.")
+			return err
+		}
+
+		fmt.Printf("You Selected:  %q\n", certProjectTypeValue)
+
+		certProjectIDLabel := promptui.Prompt{
+			Label: "Please Enter Connect Certification Project ID",
+			Validate: func(s string) error {
+				isLegacy, _ := regexp.MatchString(`^p.*`, s)
```
jomkz, line 48: Could we just strip these known prefixes when we identify them? Maybe log a debug or trace message indicating that it was stripped and try the result?

cmd/support.go
```diff
@@ -1,18 +1,102 @@
...
+
+		certProjectTypeLabel := promptui.Select{
+			Label: "Select a Certification Project Type",
+			Items: []string{"Container Image", "Operator Bundle Image"},
+		}
+
+		_, certProjectTypeValue, err := certProjectTypeLabel.Run()
+
+		if err != nil {
+			fmt.Println("Prompt Failed, Please Try re-running support command.")
```
jomkz, line 39: Feels like this message string should be a constant, as it's repeated several times.

cmd/support.go
```diff
@@ -1,18 +1,102 @@
...
+		}
+
+		certProjectIDValue, err := certProjectIDLabel.Run()
+
+		if err != nil {
+			fmt.Println("Prompt Failed, Please Try re-running support command.")
+			return err
+		}
+
+		fmt.Printf("You Entered: %q\n", certProjectIDValue)
```
jomkz, line 69: Do we need to echo back the responses as they are provided? Maybe this should be a debug or trace log message instead?

jomkz (commented):
Thanks for adding this! I just had a few comments.

cmd/support.go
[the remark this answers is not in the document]
acornett21: There will only be the production URL in the code base, with no override option for this iteration of this integration.

cmd/support.go
acornett21: I initially thought about that, but went with this blocking approach since it is built into the prompt library and I feel that gives a better user experience, since we are going with an interactive approach for this.

cmd/support.go
acornett21: I will convert this text to be a constant.

cmd/support.go
acornett21: Since the prompt framework displays what is entered, I'll convert these print statements to write to the log.

cmd/support.go
acornett21: done

cmd/support.go
acornett21: done

komish (approved):
/LGTM

jomkz (approved):
/lgtm

PR #171, opened 2021-08-18
Switch license check to use unshare/mounted env

### LINKED ISSUE #170: HasLicense check should use an Unshare environment
A mounted Unshare environment should be used when possible.

bcrochet:
In order to use native file operations, switch the license check to
use the unshare and mounted environment.

Fixes #170


certification/internal/shell/has_license_mounted.go
```diff
@@ -0,0 +1,88 @@
...
+	"path/filepath"
+
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification"
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/errors"
+	log "github.com/sirupsen/logrus"
+)
+
+const (
+	licensePath         = "/licenses"
+	newLine             = "\n"
```
skattoju, line 16: not needed any more

certification/internal/shell/has_license_mounted.go
```diff
@@ -0,0 +1,88 @@
...
+	minLicenseFileCount = 1
```
skattoju, line 17: can be removed

certification/internal/shell/has_license_mounted.go
```diff
@@ -0,0 +1,88 @@
...
+	//size format string for find command, c denotes bytes
+	minLicenseSize = "+1c"
```
skattoju, line 19: we don’t use find anymore so can be removed

skattoju (approved):
/lgtm

jomkz (approved):
/lgtm

acornett21 (approved):

certification/internal/policy/container/has_license.go
```diff
@@ -2,32 +2,67 @@ package container
...
-	licenseFileCount := strings.Count(licenseFileList, newLine)
-	log.Infof("%d Licenses found", licenseFileCount)
-	return licenseFileCount >= minLicenseFileCount, nil
+func (p *HasLicenseCheck) validate(licenseFileList []fs.DirEntry) (bool, error) {
+	nonZeroLength := false
+	for _, f := range licenseFileList {
+		info, err := f.Info()
+		if err != nil {
+			continue
+		}
```
samira-barouti, line 58: if the intention is to ignore the errors, can we update the above line to `info, _ := f.Info()`?

bcrochet: That's not just ignoring the error. It keeps the line after the block (info.Size()) from executing. It's assuming that if there is an error, the info return value is invalid.

samira-barouti: BTW, if the err is nil, doesn't the Info() method return nil for the value of `FileInfo` which will automatically escape the following if block?

bcrochet: I'm fine with leaving this as explicit as it is. I'm not seeing a gain in readability. Whether err is nil has nothing to do with the what the size of the file is.

Issue #165, opened 2021-08-16
Preflight running in a container should not require privilege


bcrochet:
## Bug Description
Preflight when run in a container currently requires a privileged container. This should not be the case.

PR #174, opened 2021-08-20
Implement a crane-based certification.CheckEngine

komish:
Context #165 

**Context**

In my testing, the [Crane](https://github.com/google/go-containerregistry/tree/main/cmd/crane) project seems to allow us to pull images, flatten images, and otherwise accomplish what we need with our container and bundle images without relying on an underlying/installed container engine, which should help make preflight more portable and require less privilege in OCP.

This PR introduces a CraneEngine, which should replace the PodmanEngine, and potentially the SkopeoEngine, in time. PodmanEngine and SkopeoEngine both rely on shelling out to skopeo and podman, which, in my testing, relies on underlying storage drivers to do some of the container and image manipulation.

This PR does not put the CraneEngine into use. In effect, nothing should have changed, other than starting to lay down the ground work for migration.

I've also started to lay things out in the filesystem without a reference to the underlying implementation, as that didn't really help us in the migration effort. My apologies for the original layout!

**Summary**

* The CraneEngine partially implements the CheckEngine interface. I believe the CheckEngine interface may benefit from some adjustments, but in an attempt to avoid breaking existing, working code, this PR does not make those changes. We can do so when it's time to shift over.
* CraneEngine assumes the responsibility of authenticating to the registry, so this PR introduces functions (`AuthConfig`) to read the credentials from the CraneEngine and determine the proper authentication context to pass to crane.
* CraneEngine provides access to a tarball containing the flattened image, so I copied over @bcrochet's `Untar` function so that we could leverage that (thanks @bcrochet)!

**Additional**

I think we could see some benefit by passing in more information to the Check.Validate(...) function. Historically, we've expected each check to operate independently, but in cases where multiple checks are running in the same execution (e.g. all of current use cases), the CraneEngine would see a benefit if it didn't have to extract the container multiple times. To this end, I've staged in the CraneEngine this idea of an ImageReference which containers some additional information. That said, one of the values in the ImageReference is an image struct that's part of the Crane project, which may be tying the Validate function too closely to an implementation.

Open for discussion, nothing to change here from my perspective because we've made no changes to this end, but I wanted to share that context.

### DIFF
--- a/certification/internal/engine/auth.go
+++ b/certification/internal/engine/auth.go
@@ -1,0 +1,33 @@
+package engine
+
+import (
+	"github.com/google/go-containerregistry/pkg/authn"
+	"github.com/google/go-containerregistry/pkg/crane"
+)
+
+// AuthConfig will check the provided credentials and return
+// a properly configured authentication option to be used by crane function
+// calls, or nil if no auth configuration was derived.
+func AuthConfig(creds RegistryCredentials) *crane.Option {
+	if notEmpty([]string{creds.Username, creds.Password}...) {
+		authOption := crane.WithAuth(&authn.Basic{
+			Username: creds.Username,
+			Password: creds.Password,
+		})
+
+		return &authOption
+	}
+
+	return nil
+}
+
+// notEmpty returns true if all strings in s are non-zero length strings
+func notEmpty(s ...string) bool {
+	for _, v := range s {
+		if len(v) <= 0 {
+			return false
+		}
+	}
+
+	return true
+}
--- a/certification/internal/engine/engine.go
+++ b/certification/internal/engine/engine.go
@@ -1,0 +1,156 @@
+package engine
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"path"
+	"time"
+
+	"github.com/google/go-containerregistry/pkg/crane"
+	log "github.com/sirupsen/logrus"
+
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification"
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/errors"
+	fileutils "github.com/redhat-openshift-ecosystem/openshift-preflight/certification/internal/utils/file"
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/runtime"
+)
+
+// CraneEngine implements a certification.CheckEngine, and leverage crane to interact with
+// the container registry and target image.
+type CraneEngine struct {
+	// Image is what is being tested, and should contain the
+	// fully addressable path (including registry, namespaces, etc)
+	// to the image
+	Image string
+	// Checks is an array of all checks to be executed against
+	// the image provided.
+	Checks []certification.Check
+	// RegistryCreds are the credentials used to access to registry
+	// from which the image will be pulled.
+	RegistryCreds RegistryCredentials
+	// IsBundle is an indicator that the asset is a bundle.
+	IsBundle bool
+
+	imageRef ImageReference
+	results  runtime.Results
+}
+
+func (c *CraneEngine) ExecuteChecks() error {
+	log.Info("target image: ", c.Image)
+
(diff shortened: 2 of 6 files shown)


certification/internal/engine/engine.go
```diff
@@ -0,0 +1,156 @@
...
+
+	// export/flatten, and extract
+	log.Debug("exporting and flattening image")
+	r, w := io.Pipe()
+	go func() {
+		defer w.Close()
+		log.Debug("writing container filesystem to output dir", containerFSPath)
+		err = crane.Export(img, w)
+		if err != nil {
+			// TODO: Handle this error more effectively. Right now we rely on
```
bcrochet, line 86: The [Export](https://github.com/google/go-containerregistry/blob/de8aff85c9090d37ed6fe436d696f3a820e5c767/pkg/crane/export.go#L26) function is *really* simple. In fact, the error that is returned would be related to the Writer. So, we probably don't even care about it here. I feel like we could actually just call crane.Export(), ignore the err, and w.Close(), and exit the goroutine. Any error would then be reported on the Reader side, so as long as the errors are handled there (which are with the Untar func), that should be sufficient.

certification/internal/engine/engine.go
```diff
@@ -0,0 +1,156 @@
...
+		c.results.PassedOverall = false
+	} else {
+		c.results.PassedOverall = true
+	}
+
+	// hash contents if bundle
+	if c.IsBundle {
+		// TODO: Implement! The current implementation of this requires a podman engine
+		// and leverages shell exec calls. Leaving this out for the moment knowing we must
+		// implement before we transition.
```
bcrochet, line 147: Should be a relatively simple transition. Happy to handle this once this lands.

bcrochet (commented):
Just a comment about error handling. Otherwise, lgtm.

certification/internal/engine/engine.go
komish: @bcrochet Understood. Would the action here be to remove the comment, or remove the logging? The logging seems beneficial, even if it's just one more breadcrumb telling us that it was some issue in the export vs. an issue in the extraction.

What do you think?

certification/internal/engine/engine.go
komish: Perfect, thanks!

bcrochet (approved):
/lgtm

certification/internal/engine/auth.go
```diff
@@ -0,0 +1,33 @@
...
+		})
+
+		return &authOption
+	}
+
+	return nil
+}
+
+// notEmpty returns true if all strings in s are non-zero length strings
+func notEmpty(s ...string) bool {
```
samira-barouti, line 25: Not a blocker: it would be great to refactor it into a util function.

certification/internal/engine/engine.go
```diff
@@ -0,0 +1,156 @@
...
+		if err != nil {
+			// TODO: Handle this error more effectively. Right now we rely on
+			// error handling in the logic to extract this export in a lower
+			// line, but we should probably exit early if the export encounters
+			// an error, which requires watching multiple error streams.
+			log.Error("unable to export and flatten container filesystem:", err)
+		}
+	}()
+
+	log.Debug("extracting container filesystem to", containerFSPath)
```
samira-barouti, line 94: seems like a space is missing after `to`

certification/internal/utils/file/helper.go
```diff
@@ -83,3 +85,73 @@ func ArtifactPath(artifact string) string {
 	}
 	return filepath.Join(artifactDir, artifact)
 }
+
+// Untar takes a destination path and a reader; a tar reader loops over the tarfile
+// creating the file structure at 'dst' along the way, and writing any files
+func Untar(dst string, r io.Reader) error {
```
samira-barouti, line 91: IMO, it would be better to pass the path to the tar file as an arg, rather than io.Reader.

PR #176, opened 2021-08-20
Implement initial crane policies into cmd chain and migrate to that as the primary path of execution

komish:
This is a bit of a hefty PR. My apologies.

This PR defines two initial checks using the new Crane-based CheckEngine. This further enables the execution of preflight on MacOS without the need for a remote podman service[1]  🎉 Woohoo! 🎉  

- The implemented operator check is RelatedImagesAreSchemaVersion2
- The "implemented" container check is HasLicense (this is "implemented", because it's actually just staged to accept the changes made by @bcrochet in #171)

In this work, I needed to
- adjust the definition of a Check so that it used the new certification.ImageReference, which contains a bit more contextual data for the new approach to utilize per-check, such as filesystem mounts, and image metadata such as layers
- updated all the previous checks to use the signature, and then added a small migration function in the shell check engine to accommodate the new signature without breaking everything
- shifted our engine.NewForConfig ==> engine.NewShellEngineForConfig, and moved the policy naming around so that we could keep both sets of policies simultaneously while we finish the migration to crane.
- Moved the commandline invocation using the old engine to `preflight check container-old` and `preflight check operator-old` should we need them.

Happy to add more detail or demo as needed. I'm sure I'll think of more detail/clarity I can add on Monday.

[1]
```
18:07:19 ~/.go/src/github.com/redhat-openshift-ecosystem/openshift-preflight ≪ branch=feature/initial-crane-policy-impl ≫
$ PFLT_LOGLEVEL=trace KUBECONFIG=foo ./preflight check operator quay.io/komish/preflight-test-bundle-passes:latest
time="2021-08-20T18:07:26-05:00" level=info msg="config file not found, proceeding without it"
time="2021-08-20T18:07:26-05:00" level=info msg="target image: quay.io/komish/preflight-test-bundle-passes:latest"
time="2021-08-20T18:07:26-05:00" level=info msg="no authentication context derived from execution configuration"
time="2021-08-20T18:07:26-05:00" level=info msg="pulling image from target registry"
time="2021-08-20T18:07:27-05:00" level=debug msg="temporary directory is/var/folders/cn/r49cg2wx5z97qwq_s48f4vm80000gn/T/preflight-453238028"
time="2021-08-20T18:07:27-05:00" level=debug msg="exporting and flattening image"
time="2021-08-20T18:07:27-05:00" level=debug msg="extracting container filesystem to/var/folders/cn/r49cg2wx5z97qwq_s48f4vm80000gn/T/preflight-453238028/fs"
time="2021-08-20T18:07:27-05:00" level=debug msg="writing container filesystem to output dir/var/folders/cn/r49cg2wx5z97qwq_s48f4vm80000gn/T/preflight-453238028/fs"
time="2021-08-20T18:07:28-05:00" level=info msg="executing checks"
time="2021-08-20T18:07:28-05:00" level=info msg="running check: RelatedImagesAreSchemaVersion2"
time="2021-08-20T18:07:28-05:00" level=info msg="check completed: RelatedImagesAreSchemaVersion2" result=PASSED
{
    "image": "quay.io/komish/preflight-test-bundle-passes:latest",
    "passed": true,
    "test_library": {
        "version": "0.0.0",
        "commit": "af50e97102a32abadafd5f37f76f47497b4e4ace"
    },
    "results": {
        "passed": [
            {
                "name": "RelatedImagesAreSchemaVersion2",
                "elapsed_time": 10,
                "description": "An operator bundle's relatedImages must be accessible at image manifest schema version 2"
            }
        ],
        "failed": [],
        "errors": []
    }
}
```

### DIFF
--- a/certification/certification.go
+++ b/certification/certification.go
@@ -1,11 +1,15 @@
 package certification
 
+import (
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+)
+
 // Check as an interface containing all methods necessary
 // to use and identify a given check.
 type Check interface {
 	// Validate will test the provided image and determine whether the
 	// image complies with the check's requirements.
-	Validate(image string) (result bool, err error)
+	Validate(imageReference ImageReference) (result bool, err error)
 	// Name returns the name of the check.
 	Name() string
 	// Metadata returns the check's metadata.
@@ -38,4 +42,10 @@ type HelpText struct {
 	// Suggestion is text provided to the user indicating what might need to
 	// change in order to pass a check.
 	Suggestion string `json:"suggestion" xml:"suggestion"`
+}
+
+type ImageReference struct {
+	ImageURI    string
+	ImageFSPath string
+	ImageInfo   *v1.Image
 }
--- a/certification/internal/policy/operator/has_v2_related_image_schema_version.go
+++ b/certification/internal/policy/operator/has_v2_related_image_schema_version.go
@@ -1,14 +1,15 @@
-package shell
+package operator
 
 import (
 	"encoding/json"
 	"fmt"
+	"path"
 
-	"github.com/operator-framework/api/pkg/manifests"
+	"github.com/google/go-containerregistry/pkg/crane"
 	operatorsv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification"
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/errors"
-	"github.com/redhat-openshift-ecosystem/openshift-preflight/cli"
+	containerutils "github.com/redhat-openshift-ecosystem/openshift-preflight/certification/internal/utils/container"
 	log "github.com/sirupsen/logrus"
 )
 
@@ -19,7 +20,7 @@ type RelatedImagesAreSchemaVersion2Check struct{}
 // Validate checks the image manifest for each related image referenced in a
 // ClusterServiceVersion and ensures that the schema version used is version 2.
 func (p *RelatedImagesAreSchemaVersion2Check) Validate(imgRef certification.ImageReference) (bool, error) {
-	imageToSchemaVersion, err := p.getDataToValidate(imgRef.ImageURI)
+	imageToSchemaVersion, err := p.getDataToValidate(imgRef)
 	if err != nil {
 		return false, fmt.Errorf("%w: %s", errors.ErrRunContainerFailed, err)
 	}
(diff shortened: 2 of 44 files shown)


certification/certification.go
```diff
@@ -1,11 +1,15 @@
...
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+)
+
 // Check as an interface containing all methods necessary
 // to use and identify a given check.
 type Check interface {
 	// Validate will test the provided image and determine whether the
 	// image complies with the check's requirements.
-	Validate(image string) (result bool, err error)
+	Validate(imageReference ImageReference) (result bool, err error)
```
bcrochet, line 12: If we're going to be changing this interface, do we want to make sure that our implementation matches what we've learned? Make ImageReference and interface, and return a struct instead of a bool?

bcrochet (changes requested):
Couple of requests.

certification/internal/policy/operator/has_v2_related_image_schema_version.go
```diff
@@ -0,0 +1,182 @@
...
+		log.Error(err)
+		return nil, fmt.Errorf("%w: %s", errors.ErrDeterminingRelatedImageSchemaVers, err)
+	}
+
+	return imageSchemaVersionMap, nil
+}
+
+// readBundle will accept the manifests directory where a bundle is expected to live,
+// and walks the directory to find all bundle assets.
+func (p *RelatedImagesAreSchemaVersion2Check) readBundle(manifestsDir string) (*operatorsv1alpha1.ClusterServiceVersion, error) {
```
samira-barouti, line 59: Can we have this method in util/container instead? It seems a couple of checks could leverage this method.

certification/internal/policy/operator/has_v2_related_image_schema_version.go
```diff
@@ -0,0 +1,182 @@
...
+
+// getRelatedImagesForCSV will return a slice of strings containing the images found in the relatedImages field of
+// the input ClusterServiceVersion.
+func (p *RelatedImagesAreSchemaVersion2Check) getRelatedImagesForCSV(csv *operatorsv1alpha1.ClusterServiceVersion) ([]string, error) {
+	relatedImages := csv.Spec.RelatedImages
+
+	// no related images == nothing to check
+	if len(relatedImages) == 0 {
+		return nil, nil
+	}
```
samira-barouti, line 76: I believe you can remove this if block, and this method will return an empty slice. I think returning an empty slice is preferable to nil, to reduce the risk of nil dereference.

certification/internal/policy/operator/has_v2_related_image_schema_version.go
```diff
@@ -0,0 +1,182 @@
...
+		}
+	}
+
+	// if we determine that all images are using schemaVersion 2, then we pass the check
+	if len(v2ImageManifests) == len(imageToSchemaVersionMap) {
+		return true, nil
+	}
+
+	log.Info("the following related images found in CSV were not image manifests with schema version 2", notv2ImageManifests)
+	return false, nil
```
samira-barouti, line 159: A nit: I would have the log (LoC 158) printed in an if block and use return len(notv2ImageManifests) == 0, nil as the return statement.

certification/certification.go
komish: Per discussion, you mentioned you'll look at what a PoC of this could look like! Closing this out.

certification/internal/policy/operator/has_v2_related_image_schema_version.go
komish: Great idea, I've made the function call generic and am leveraging it in the new policy definition.

A note: I did not change the way this works in the old engine's equivalent policy definition as we're moving away from it anyway.

certification/internal/policy/operator/has_v2_related_image_schema_version.go
komish: I don't think it returns an empty slice if it's uninitialized, so I've changed this to explicitly return an empty slice here instead of the nil to reduce the likelihood of a nil dereference.

certification/internal/policy/operator/has_v2_related_image_schema_version.go
komish: I personally prefer the explicit return, as I believe it makes it a bit easier to follow in debug cases.

If I return the `len` evaluation, then a human debugger reviewing this would need to evaluate if there are any places in the code where the input to `len()` would have changed from what they expected. I feel it makes things a bit more cumbersome to troubleshoot.

bcrochet (approved):

jomkz (approved):
I am good with this, pending the comments from Samira.

bcrochet (approved):
/lgtm

komish:
> I am good with this, pending the comments from Samira.

+1 given the size and open comments, waiting for explicit approval from @samira-barouti. Samira, if you're set, feel free to hit merge!

samira-barouti (approved):
/lgtm

PR #192, opened 2021-08-31
Migrate DeployableByOLM to Crane

### LINKED ISSUE #184: Migrate Deployable by OLM to Crane
## Bug Description
Migrate the Deployable by OLM check to Crane and increase the unit testing code coverage

## Steps to Reproduce:
(How can we reproduce this?)
1)
2)
3)

## Expected Result
(What did you expect to happen and why?)

## Actual Result
(What actually happened)

## Additional Context
(Anything else you think might help us troubleshoot, like your platform, dependency versions, etc).

samira-barouti:
This PR migrates the DeployableByOLM to Crane. Also refactoring has been done to improve the code readability.

* Previously, the operator data were defined as globals vars. This PR introduces a new type, `OperatorData`, that stores the data that the check needs for the successful deployment of an operator. `OperatorData` is the arg that will be passed to the downstream methods that need these info.
* A new method has been created to fetch operator data from viper and bundle image and return a reference to`OperatorData`. 
* The new OpenshiftEngine has a `*rest.Config` as a field. The check does not pass the `*rest.Config` as a method arg, any longer. 
* All bundle-related constants are moved to default.go

Fixes #184

### DIFF
--- a/certification/internal/policy/operator/operator_pkg_name_uniqueness.go
+++ b/certification/internal/policy/operator/operator_pkg_name_uniqueness.go
@@ -11,14 +11,6 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
-const (
-	// apiEndpoint is the endpoint used to query for package uniqueness.
-	apiEndpoint = "https://catalog.redhat.com/api/containers/v1/operators/packages"
-
-	// packageKey is the packageKey in annotations.yaml that contains the package name.
-	packageKey = "operators.operatorframework.io.bundle.package.v1"
-)
-
 // apiRespondData is the response received from the defined API
 type apiResponseData struct {
 	Data     []packageData `json:"data"`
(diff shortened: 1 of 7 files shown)


certification/internal/policy/operator/operator_pkg_name_uniqueness.go
```diff
@@ -11,14 +11,6 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
-const (
```
bcrochet, line 14: This belongs in a separate PR.

certification/internal/k8s/deployable_by_olm_mounted.go
```diff
@@ -75,7 +75,7 @@ func (p *DeployableByOlmMountedCheck) setUp(bundle string) error {
 	}
 
 	// retrieve the operator metadata from bundle image
-	data, err := containerutil.GetAnnotationsFromBundle(bundle)
```
bcrochet, line 78: I wouldn't bother with these changes, as this code is going to go away very soon. I'd rather not see it in the same commit.

bcrochet (changes requested):
Couple of comments.

certification/internal/policy/operator/operator_pkg_name_uniqueness.go
samira-barouti: The DeployableByOLM requires the package key, among the others. The PackageUniqueness check also requires this key. I was not comfortable using a const which has been defined in another file, and that was the main motivation to move all these common constants into a central place.

certification/internal/k8s/deployable_by_olm_mounted.go
samira-barouti: sure, I can revert this change.

certification/internal/policy/operator/operator_pkg_name_uniqueness.go
bcrochet: Ok. Perfect. I didn't gleam immediately that it was related. Totally fine with that.

bcrochet (approved):
/lgtm

PR #188, opened 2021-08-31
Add Validate Operator Bundle check to CraneEngine

bcrochet:
The Validate Operator Bundle will now look for a bundle path, instead of a bundle image.

Signed-off-by: Brad P. Crochet <brad@redhat.com>

PR #195, opened 2021-09-01
Move the scorecard checks to use the crane engine

### LINKED ISSUE #165: Preflight running in a container should not require privilege

bcrochet:
The scorecard checks were already enabled, but still pulling an image via podman. This now just uses the on-disk bundle.

Depends-on: #188 
Fixes #165 

### DIFF
--- a/certification/internal/policy/operator/scorecard_basic_check_spec.go
+++ b/certification/internal/policy/operator/scorecard_basic_check_spec.go
@@ -1,0 +1,53 @@
+package operator
+
+import (
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification"
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/cli"
+	log "github.com/sirupsen/logrus"
+)
+
+// ScorecardBasicSpecCheck evaluates the image to ensure it passes the operator-sdk
+// scorecard check with the basic-check-spec-test suite selected.
+type ScorecardBasicSpecCheck struct {
+	scorecardCheck
+}
+
+const scorecardBasicCheckResult string = "operator_bundle_scorecard_BasicSpecCheck.json"
+
+func NewScorecardBasicSpecCheck(operatorSdkEngine *cli.OperatorSdkEngine) *ScorecardBasicSpecCheck {
+	return &ScorecardBasicSpecCheck{
+		scorecardCheck{OperatorSdkEngine: *operatorSdkEngine},
+	}
+}
+
+func (p *ScorecardBasicSpecCheck) Validate(bundleRef certification.ImageReference) (bool, error) {
+	log.Debug("Running operator-sdk scorecard check for ", bundleRef.ImageURI)
+	selector := []string{"test=basic-check-spec-test"}
+	log.Debugf("--selector=%s", selector)
+	scorecardReport, err := p.getDataToValidate(bundleRef.ImageFSPath, selector, scorecardBasicCheckResult)
+	if err != nil {
+		return false, err
+	}
+
+	return p.validate(scorecardReport.Items)
+}
+
+func (p *ScorecardBasicSpecCheck) Name() string {
+	return "ScorecardBasicSpecCheck"
+}
+
+func (p *ScorecardBasicSpecCheck) Metadata() certification.Metadata {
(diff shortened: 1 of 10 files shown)


certification/engine/operatorsdk.go
```diff
@@ -0,0 +1,184 @@
...
+	if opts.ServiceAccount != "" {
+		cmdArgs = append(cmdArgs, "--service-account", opts.ServiceAccount)
+	}
+
+	configFile, err := createScorecardConfigFile()
+	if err != nil {
+		log.Error("could not create scorecard config file", err)
+		return nil, err
+	}
+	defer os.Remove(configFile)
```
samira-barouti, line 51: I think the defer statement should be placed right after line 46. Otherwise if an error occurred while executing `createScorecardConfigFile`, the Scorecard method returns and the execution will never reach this line to push the deferred method to the stack.

certification/internal/policy/operator/scorecard_basic_check_spec.go
```diff
@@ -0,0 +1,53 @@
...
+const scorecardBasicCheckResult string = "operator_bundle_scorecard_BasicSpecCheck.json"
+
+func NewScorecardBasicSpecCheck(operatorSdkEngine *cli.OperatorSdkEngine) *ScorecardBasicSpecCheck {
+	return &ScorecardBasicSpecCheck{
+		scorecardCheck{OperatorSdkEngine: *operatorSdkEngine},
+	}
+}
+
+func (p *ScorecardBasicSpecCheck) Validate(bundleRef certification.ImageReference) (bool, error) {
+	log.Debug("Running operator-sdk scorecard check for ", bundleRef.ImageFSPath)
```
samira-barouti, line 24: I think we want to print bundleRef.ImageURI.

certification/engine/operatorsdk.go
```diff
@@ -0,0 +1,184 @@
...
+	cmdArgs = append(cmdArgs, "--config", configFile)
+	cmdArgs = append(cmdArgs, "--verbose")
```
samira-barouti, line 53: I suggest for the sake of consistency, we follow the same path as BundleValidate and let the caller method set a `verbosity` option.

acornett21 (approved):

certification/engine/operatorsdk.go
```diff
@@ -0,0 +1,188 @@
...
+		cmdArgs = append(cmdArgs, "--namespace", opts.Namespace)
+	}
+	if opts.ServiceAccount != "" {
+		cmdArgs = append(cmdArgs, "--service-account", opts.ServiceAccount)
+	}
+
+	configFile, err := createScorecardConfigFile()
+	if configFile != "" {
+		defer os.Remove(configFile)
+	}
```
samira-barouti, line 49: Just did a quick test and it seems os.Remove returns an error that the directory does not exist. I think you can remove the if block and ignore the returned error.

samira-barouti (approved):
Just a little nit

certification/engine/operatorsdk.go
```diff
@@ -0,0 +1,188 @@
+package engine
```
samira-barouti, line 1: This is probably a broader discussion on the code design: we have two engine directories atm; I believe we should be more prescriptive about which one should be used for the preflight internal. My intuition is to use the certification/internal/engine for everything which is internal to preflight (such as operatorsdk, openshift, crane) and let the certification/engine be for the check engine. I would like to hear everyone's feedback on this matter. 

@komish @bcrochet @acornett21 @jomkz

certification/engine/operatorsdk.go
bcrochet: This will be handled in #188

certification/engine/operatorsdk.go
bcrochet: I will test whether this can continue to live in the internal directory. If that's the case, then it's probably a moot point, and I'll just leave it there.

bcrochet: And... it can. I'll move this to internal.

samira-barouti (approved):
/lgtm

PR #199, opened 2021-09-01
add PreRun to all subcommands to load configurations

### LINKED ISSUE #119: Do not generate log file when running help sub-command
## Bug Description

When running `preflight help`, an empty log file is created.

```bash
$ ./preflight --version
preflight version 0.0.0 <commit: 2d3bb671bff8a95d385621382f31215234877d44>
```

## Steps to Reproduce:

1) Build the preflight binary locally: `make build` 
2) Run the help sub-command: `./preflight help`

## Expected Result

An empty log file should not be created when running the `help` sub-command.

## Actual Result

An empty log file named `preflight.log` is created.
```bash
ls -la preflight*
-rwxrwxr-x. 1 john john 11072372 Jul 21 10:07 preflight
-rw-------. 1 john john        0 Jul 21 10:08 preflight.log
```

## Additional Context

N/A

acornett21:
This PR will add PreRun to all subcommands to load configurations, and remove `init()` from the root cmd to insure that configurations are only loaded when a given `subcommand` is executed.

fixes #119

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -2,14 +2,11 @@
 package cmd
 
 import (
-	"io"
 	"os"
 
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/version"
-	"github.com/spf13/cobra"
-	"github.com/spf13/viper"
-
 	log "github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
 )
 
 var rootCmd = &cobra.Command{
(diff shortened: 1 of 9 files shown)


cmd/root.go
```diff
@@ -74,3 +82,17 @@ func initConfig() {
...
+// isHelpCMD is used to determine if the user is trying to execute a help or a -h command,
+// if so then preflight does not need to load configurations or create a log file
+func isHelpCMD(args []string) bool {
+	var isHelpCMD bool
+	for _, a := range args {
+		if strings.Contains(a, help) || strings.Contains(a, dashH) {
+			isHelpCMD = true
+			break
+		}
+	}
```
samira-barouti, line 95: I think it would be great to refactor this method as a utility function for the future use. I suggest a method with this signature: `func ContainsAtLeast(keys []string, objects []string) bool` or `func ContainsAtLeast(keys ...string, objects []string) bool`

bcrochet (changes requested):
I feel like we are straying very far from the design of Cobra. We should be utilize pre (and possibly post) run hooks in order to so some of these things. IIUC, it already has the checking for -h and help, and we shouldn't need to be writing code to check for those. If I'm wrong, please tell me.

acornett21:
@bcrochet I can't say for certain if we are straying away from the design of cobra or not. What I can say is that since each sub command calls `init()` and then the root command's `init()` calls `initConfig()` this causes the configurations to get loaded and the log file placed on disk. Since `help` and `-h` aren't our commands, this was the only way I could think of to stop the `init()`from getting triggered. If there is a way natively for go to stop `init()` calls from happening let me know.

cmd/root.go
acornett21: The implementation was changed to use `PreRun` in the subcommands, so this method no longer exists. I'll keep things like this in mind in the future, to make things more generic and reusable.

bcrochet (approved):
Love it. Much cleaner.

/lgtm

samira-barouti (approved):
Beautiful!

PR #203, opened 2021-09-02
Move 3 container checks to crane

### LINKED ISSUE #202: Migrate LayerCountAcceptable Check to Crane
## Bug Description
Migrate LayerCountAcceptable Check to Crane

NA

## Steps to Reproduce:
NA

## Expected Result
NA

## Actual Result
NA

## Additional Context
NA

bcrochet:
Moves 3 container checks:
Max layers
Prohibited packages
Required Labels

Fixes #202

### DIFF
--- a/certification/internal/policy/container/has_prohibited_packages.go
+++ b/certification/internal/policy/container/has_prohibited_packages.go
@@ -1,0 +1,113 @@
+package container
+
+import (
+	"path/filepath"
+	"strings"
+
+	rpmdb "github.com/knqyf263/go-rpmdb/pkg"
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification"
+	log "github.com/sirupsen/logrus"
+)
+
+// HasProhibitedPackages evaluates that the image does not contain prohibited packages,
+// which refers to packages that are not redistributable without an appropriate license.
+type HasNoProhibitedPackagesCheck struct{}
+
+func (p *HasNoProhibitedPackagesCheck) Validate(imgRef certification.ImageReference) (bool, error) {
+	pkgList, err := p.getDataToValidate(imgRef.ImageFSPath)
+	if err != nil {
+		log.Error("unable to get a list of all packages in the image")
+		return false, err
+	}
+
+	return p.validate(pkgList)
+}
+
+func (p *HasNoProhibitedPackagesCheck) getDataToValidate(dir string) ([]string, error) {
+	db, err := rpmdb.Open(filepath.Join(dir, "var", "lib", "rpm", "Packages"))
+	if err != nil {
+		return nil, err
+	}
+	pkgList, err := db.ListPackages()
+	if err != nil {
+		return nil, err
+	}
+	pkgs := make([]string, len(pkgList))
+	for i, pkg := range pkgList {
+		pkgs[i] = pkg.Name
+	}
+	return pkgs, nil
+}
+
+func (p *HasNoProhibitedPackagesCheck) validate(pkgList []string) (bool, error) {
+	var prohibitedPackages []string
+	for _, pkg := range pkgList {
+		_, ok := prohibitedPackageList[pkg]
+		if ok {
+			prohibitedPackages = append(prohibitedPackages, pkg)
+			continue
+		}
+		for _, name := range prohibitedPackageGlobList {
+			if strings.HasPrefix(pkg, name) {
+				prohibitedPackages = append(prohibitedPackages, pkg)
+				continue
+			}
+		}
+	}
+
+	if len(prohibitedPackages) > 0 {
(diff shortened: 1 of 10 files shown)


certification/internal/policy/container/has_required_labels.go
```diff
@@ -0,0 +1,63 @@
...
+)
+
+var requiredLabels = []string{"name", "vendor", "version", "release", "summary", "description"}
+
+// HasRequiredLabelsCheck evaluates the image manifest to ensure that the appropriate metadata
+// labels are present on the image asset as it exists in its current container registry.
+type HasRequiredLabelsCheck struct{}
+
+func (p *HasRequiredLabelsCheck) Validate(imgRef certification.ImageReference) (bool, error) {
+	// TODO: if we're going have the image json on disk already, we should use it here instead of podman inspect-ing.
```
samira-barouti, line 16: This TODO could be removed now \o/

certification/internal/policy/container/less_than_max_layers.go
```diff
@@ -0,0 +1,55 @@
...
+
+const (
+	acceptableLayerMax = 40
+)
+
+// UnderLayerMaxCheck ensures that the image has less layers in its assembly than a predefined maximum.
+type UnderLayerMaxCheck struct{}
+
+func (p *UnderLayerMaxCheck) Validate(imageRef certification.ImageReference) (bool, error) {
+	// TODO: if we're going have the image json on disk already, we should use it here instead of podman inspect-ing.
```
samira-barouti, line 19: Same goes here

samira-barouti (approved):
Two small nits... otherwise good to go

samira-barouti (approved):
/lgtm

certification/internal/policy/container/less_than_max_layers.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification"
+	log "github.com/sirupsen/logrus"
+)
+
+const (
+	acceptableLayerMax = 40
+)
+
+// UnderLayerMaxCheck ensures that the image has less layers in its assembly than a predefined maximum.
+type UnderLayerMaxCheck struct{}
```
acornett21, line 16: I think the file name and the check name are confusing for this check, and this one doesn't seem to follow the conventions of the other one. However of all the check, the rest have more natural names by the nature of what the check is doing.

certification/internal/policy/container/has_prohibited_packages.go
```diff
@@ -0,0 +1,114 @@
...
+		}
+	}
+
+	log.Warn("The number of prohibited package found in the container image: ", len(prohibitedPackages))
+	if len(prohibitedPackages) == 0 {
+		return true, nil
+	}
+
+	log.Warn("found the following prohibited packages: ", prohibitedPackages)
+	return false, nil
```
samira-barouti, line 64: I'd probably rewrite this as follow to be consistent with the required label check:
```
if len(prohibitedPackages) > 0 {
    log.Warn("The number of prohibited package found in the container image: ", len(prohibitedPackages))
    log.Warn("found the following prohibited packages: ", prohibitedPackages)
}
return len(prohibitedPackages) == 0, nil
```

certification/internal/policy/container/has_prohibited_packages.go
```diff
@@ -0,0 +1,114 @@
...
+		return nil, err
+	}
+	pkgList, err := db.ListPackages()
+	if err != nil {
+		return nil, err
+	}
+	pkgs := make([]string, len(pkgList))
+	for i, pkg := range pkgList {
+		pkgs[i] = pkg.Name
+	}
```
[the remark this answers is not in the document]
bcrochet: So, there are two reasons that I implemented it this way. One, I did not want to leak the data structure returned by db.ListPackages, so iterating over that list to generate a string slice was done. Two, supporting the glob. By iterating over the list of packages (as opposed to iterating over the prohibited packages), I can also check the list of globs and see if the package name starts with anything in the glob list (which happens to be just one atm). There is no operator to see if a key starts with a value (without iterating over the map). So I don't think this optimization would save us anything. If anything, we'd be doing more operations. I see where you're going, and if it weren't for the glob, I'd agree.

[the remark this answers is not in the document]
samira-barouti: Makes sense... I overlooked the partial matching that needs to be done for the glob.

RE the other point, I still see value in merging getDataToValidate() and validate() and doing one iteration; but if the number of container packages is not huge, the extra iteration is probably negligible in terms of elapsed time.

samira-barouti (approved):
/lgtm

PR #205, opened 2021-09-03
remove related image check which is no longer required

komish:
This PR removes the check that inspects related images and ensures their Manifests are schemaVersion 2. This is being removed because it is no longer required.

It has been removed from the Crane engine as well as the older Shell engine.


samira-barouti (approved):
/lgtm

bcrochet (approved):

PR #206, opened 2021-09-03
Migrating HasUniqueTag

yashoza19:


certification/internal/policy/container/has_unique_tag.go
```diff
@@ -0,0 +1,61 @@
...
+	tags, err := p.getDataToValidate(imgRef.ImageURI)
+	if err != nil {
+		return false, err
+	}
+	return p.validate(imgRef.ImageURI, tags)
+}
+
+func (p *HasUniqueTagCheck) getDataToValidate(image string) ([]string, error) {
+	skopeoEngine := SkopeoCLIEngine{}
+	runReport, err := skopeoEngine.ListTags(image)
```
samira-barouti, line 25: Crane has a ListTag method. I am curious to know if it is possible to return the tags for a specific image by using it: https://pkg.go.dev/github.com/google/go-containerregistry/pkg/crane#ListTags

PR #209, opened 2021-09-03
Deprecate ioutil for Go 1.16

sebrandon1:
Per the change log for Go 1.16, the `ioutil` library has some routines being deprecated/moved: https://golang.org/doc/go1.16#ioutil


komish:
I see that `ioutil.TempDir` == `os.MkdirTemp` in 1.17, but os.MkdirTemp already exists in 1.16 in some form, and I do believe we use it elsewhere so this...

/LGTM

Thanks @sebrandon1

bcrochet (approved):
/lgtm

PR #210, opened 2021-09-03
Remove unnecessary auth and document new auth approach

### LINKED ISSUE #204: Allow preflight to target container registries requiring authentication for image pulling and any checks
Yes, Preflight, utilizing Crane, does not allow for interacting with registries behind authentication.

To allow for cases where we need to authenticate to multiple registries, we should accept an environment variable `PFLT_REGISTRY_CREDENTIAL` that points to a /path/on/disk/dockerconfig.json-style authentication file. This can then be leveraged by CI cases in k8s, where the authentication details may be mounted into a container running preflight that can then be leveraged.

This authentication also needs to be passed to to checks such as RelatedImageManifestsAreSchemaVersion2 in case the related images are also in private registries.

Currently, there is a partially wired solution that allows for authentication using a username and password, but using a credential file is probably more in line with what we want.

That partially implemented solution can be seen here: https://github.com/redhat-openshift-ecosystem/openshift-preflight/blob/main/certification/internal/engine/engine.go#L47-L53

komish:
Fixed #204 

This PR just removes the partially wired RegistryCredentials that would have been passed as an option to Crane, enabling authentication to registries. It then adds documentation indicating that users should rely on `$DOCKER_CONFIG/config.json` as the recommended approach to handle authentication.

Some markdown formatting tweaks that the linter complained about also made it into this PR.


bcrochet (approved):
/lgtm

samira-barouti (approved):
/lgtm

acornett21 (approved):

PR #214, opened 2021-09-07
Move BasedOnUBI to Crane Engine

bcrochet:
This moves the check to Crane, and implements a check for the
"com.redhat.component" label, in that it contains "ubi".

It still checks for the contents of os-release.

### DIFF
--- a/certification/internal/policy/container/base_on_ubi.go
+++ b/certification/internal/policy/container/base_on_ubi.go
@@ -1,0 +1,90 @@
+package container
+
+import (
+	"os"
+	"path/filepath"
+	"strings"
+
+	cranev1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification"
+	log "github.com/sirupsen/logrus"
+)
+
+// BasedOnUBICheck evaluates if the provided image is based on the Red Hat Universal Base Image
+// by inspecting the contents of the `/etc/os-release` and identifying if the ID is `rhel` and the
+// Name value is `Red Hat Enterprise Linux`
+type BasedOnUBICheck struct{}
+
+func (p *BasedOnUBICheck) Validate(imgRef certification.ImageReference) (bool, error) {
+	labels, err := p.getLabels(imgRef.ImageInfo)
+	if err != nil {
+		return false, err
+	}
+
+	osRelease, err := p.getOsReleaseContents(imgRef.ImageFSPath)
+	if err != nil {
+		log.Debugf("could not retrieve contents of os-release")
+		return false, err
+	}
+
+	return p.validate(labels, osRelease)
+}
+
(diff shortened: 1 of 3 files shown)


acornett21 (approved):

certification/internal/policy/container/base_on_ubi.go
```diff
@@ -0,0 +1,91 @@
...
+		if strings.HasPrefix(value, `ID="rhel"`) {
+			hasRHELID = true
+		} else if strings.HasPrefix(value, `NAME="Red Hat Enterprise Linux"`) {
+			hasRHELName = true
+		}
+	}
+
+	component, hasComponentLabel := labels["com.redhat.component"]
+	if hasComponentLabel {
+		hasComponentLabel = strings.Contains(component, "ubi")
```
samira-barouti, line 63: I think it would make sense to use a separate var to indicate whether ubi is present in the label or not. Reusing the same var is a bit confusing and the var name is generic.

bcrochet: I could do that. Then I'd have to declare it in the correct scope. Think of it as a two-step process. If the label is not present at all, then it fails. If the label is present, but isn't actually ubi, then it fails. If you feel very strongly about it, I can change it. If not, I'm fine with it as is.

samira-barouti: I don't have strong opinion... As a code reviewer, I feel having a descriptive variable name will help a lot. Based on your explanation, like 63 is looking for the `ubi` token. IMO using a var like `hasUbiLabel` makes it easier to understand what is happening. The hasComponentLabel totally makes sense as we are looking for the `component` tag. For the UBI label, I think a more specific var makes more sense.

bcrochet: I get where you're coming from. hasComponentLabel does describe what line 61 is doing. The point of line 63 is that the label must be present AND contain the "ubi" substring. Otherwise, I will then have to do an && of 4 booleans to test the validity, when the second check is already dependent on the first check, so I just don't feel it warrants its own variable. Does hasUbiComponentLabel satisfy the clarity?

samira-barouti: Thank you for further elaboration. The new variable name is definitely made the code readable. I however believe enforcing the immutable objects and avoid reassigning value to vars after they have been created, is a good coding practice; it avoids confusion/unwanted side effects and make the code readable and maintainable.

acornett21: I've been following this thread, and following the code changes. I think the latests with `exists` is the cleanest/clearest code over the last day. I personally like that style of `exists` or `ok` check then `var` assignment. I'd call this good.

acornett21 (approved):

acornett21 (approved):

PR #220, opened 2021-09-08
Dependency updates

sebrandon1:
Dependency updates for some Go packages used in this project.


bcrochet (approved):
This will change a bit after the deprecated removal, but step in the right direction.

acornett21 (approved):

PR #226, opened 2021-09-09
Remove unwanted log message

bcrochet:
A log message, at default "info" level (since it was a log.Println)
was in use when first debugging the Untar function, but it was an
extremely verbose log, so even with trace it is overwhelming, and
doesn't add any good info for the end user. So, this removes it.


komish (approved):
/lgtm

PR #227, opened 2021-09-09
Fix the license check so that failures vs errors are correct

bcrochet:
If the license dir did not exist, it was being reports as an error. It
is supposed to be a failure. If /licenses is not a dir, that a failure,
not an error. If the dir can't be read, then that's treated as a failure.


acornett21 (approved):

komish (approved):
/lgtm

PR #228, opened 2021-09-09
Update hash label in results

### LINKED ISSUE #225: Change `bundle_hash` in results to instead be `certification_hash`
## Bug Description
The API that will receive the results submission has the value of our bundle hash under the key `certification_hash`. Preflight needs to change this to facilitate result submission.

https://github.com/redhat-openshift-ecosystem/openshift-preflight/blob/4a625234b2b8057a9c7be2eb73d7154bae08ae70/certification/formatters/util.go#L69

## Expected Result
A result blob that looks like this (truncated)

```
{
    "image": "quay.io/komish/kogito-testing:bundle",
    "passed": false,
    "certification_hash": "d41d8cd98f00b204e9800998ecf8427e",      <--- this key!
    "test_library": {
	...
    },
    "results": {
	...
    }
}
```

komish:
Fixes #225

The bundle_hash key in the result is not a part of the spec for submitting results to pyxis, and it should be certification_hash instead.

### DIFF
--- a/certification/formatters/util.go
+++ b/certification/formatters/util.go
@@ -48,10 +48,10 @@ func getResponse(r runtime.Results) UserResponse {
 	}
 
 	response := UserResponse{
-		Image:       r.TestedImage,
-		Passed:      r.PassedOverall,
-		LibraryInfo: version.Version,
-		BundleHash:  r.BundleHash,
+		Image:             r.TestedImage,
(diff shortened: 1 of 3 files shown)


acornett21 (approved):

bcrochet (approved):
/lgtm

PR #230, opened 2021-09-09
cd'ing into the proper bundle directory to generate md5 hash

### LINKED ISSUE #224: preflight process does not complete when running in a container
## Bug Description
When running preflight in a container on a cluster, `check operator` process does not complete. It appears to be an issues with the MD5 hash logic based on the logs

## Steps to Reproduce:
(How can we reproduce this?)
1) Deploy preflight to an cluster OpenShift cluster
2) Run `oc rsh <pod-name>`
3) Run `preflight check operator <any-bundle-image>`
4) Wait for the process to hang

## Expected Result
preflight should finish all checks and produce results

## Actual Result
preflight does not complete and the process hangs

acornett21:
The directory that was being passed is was not being used, which caused `exec.Command` to be run against the entire file system, causing preflight to hang. This PR makes sure the we use the image directory and run the `exec.Command` in that directory.

fixes #224

### DIFF
--- a/certification/internal/utils/container/helper.go
+++ b/certification/internal/utils/container/helper.go
@@ -20,25 +20,30 @@ import (
 
 func GenerateBundleHash(image string) (string, error) {
 	// TODO: Convert this to regular Go commands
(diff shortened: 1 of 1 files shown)


bcrochet (approved):
/lgtm

komish:
We've already discussed that this doesn't work on mac because of the use of gnufind vs bsd find, and md5sum vs. md5. That said, in our container image, this seems to work. 

/lgtm

PR #232, opened 2021-09-10
Add unit tests for deployableByOLM check + refactoring

### LINKED ISSUE #198: Add unit tests for DeployableByOLM Check
## Bug Description
Add unit tests for DeployableByOLM Check

## Steps to Reproduce:
(How can we reproduce this?)
1)
2)
3)

## Expected Result
(What did you expect to happen and why?)

## Actual Result
(What actually happened)

## Additional Context
(Anything else you think might help us troubleshoot, like your platform, dependency versions, etc).

samira-barouti:
Add unit tests for deployableByOLM check + refactoring

Fixes #198

### DIFF
--- a/certification/internal/cli/openshift.go
+++ b/certification/internal/cli/openshift.go
@@ -4,18 +4,11 @@ import (
 	operatorv1 "github.com/operator-framework/api/pkg/operators/v1"
 	operatorv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/client-go/rest"
 )
 
 type OpenshiftOptions struct {
 	Namespace string
 	Labels    map[string]string
-}
-
-type OpenshiftReport struct {
-	Stdout string
-	Stderr string
-	Items  []OperatorSdkScorecardItem `json:"items"`
 }
 
 type SubscriptionData struct {
@@ -37,21 +30,23 @@ type OperatorGroupData struct {
 }
 
 type OpenshiftEngine interface {
-	CreateNamespace(name string, opts OpenshiftOptions, config *rest.Config) (*OpenshiftReport, error)
-	DeleteNamespace(name string, opts OpenshiftOptions, config *rest.Config) error
-	GetNamespace(name string, config *rest.Config) (*corev1.Namespace, error)
+	Setup() error
 
-	CreateOperatorGroup(data OperatorGroupData, opts OpenshiftOptions, config *rest.Config) (*OpenshiftReport, error)
-	DeleteOperatorGroup(name string, opts OpenshiftOptions, config *rest.Config) (*OpenshiftReport, error)
-	GetOperatorGroup(name string, opts OpenshiftOptions, config *rest.Config) (*operatorv1.OperatorGroup, error)
+	CreateNamespace(name string, opts OpenshiftOptions) (*corev1.Namespace, error)
+	DeleteNamespace(name string, opts OpenshiftOptions) error
+	GetNamespace(name string) (*corev1.Namespace, error)
 
-	CreateCatalogSource(data CatalogSourceData, opts OpenshiftOptions, config *rest.Config) (*OpenshiftReport, error)
-	DeleteCatalogSource(name string, opts OpenshiftOptions, config *rest.Config) (*OpenshiftReport, error)
-	GetCatalogSource(name string, opts OpenshiftOptions, config *rest.Config) (*operatorv1alpha1.CatalogSource, error)
+	CreateOperatorGroup(data OperatorGroupData, opts OpenshiftOptions) (*operatorv1.OperatorGroup, error)
+	DeleteOperatorGroup(name string, opts OpenshiftOptions) error
+	GetOperatorGroup(name string, opts OpenshiftOptions) (*operatorv1.OperatorGroup, error)
 
-	CreateSubscription(data SubscriptionData, opts OpenshiftOptions, config *rest.Config) (*OpenshiftReport, error)
-	DeleteSubscription(name string, opts OpenshiftOptions, config *rest.Config) (*OpenshiftReport, error)
-	GetSubscription(name string, opts OpenshiftOptions, config *rest.Config) (*operatorv1alpha1.Subscription, error)
+	CreateCatalogSource(data CatalogSourceData, opts OpenshiftOptions) (*operatorv1alpha1.CatalogSource, error)
+	DeleteCatalogSource(name string, opts OpenshiftOptions) error
+	GetCatalogSource(name string, opts OpenshiftOptions) (*operatorv1alpha1.CatalogSource, error)
 
-	GetCSV(name string, opts OpenshiftOptions, config *rest.Config) (*operatorv1alpha1.ClusterServiceVersion, error)
+	CreateSubscription(data SubscriptionData, opts OpenshiftOptions) (*operatorv1alpha1.Subscription, error)
+	DeleteSubscription(name string, opts OpenshiftOptions) error
+	GetSubscription(name string, opts OpenshiftOptions) (*operatorv1alpha1.Subscription, error)
+
+	GetCSV(name string, opts OpenshiftOptions) (*operatorv1alpha1.ClusterServiceVersion, error)
 }
--- a/certification/internal/engine/openshift.go
+++ b/certification/internal/engine/openshift.go
@@ -3,6 +3,7 @@ package engine
 import (
 	"context"
 	"fmt"
+	"os"
 
 	log "github.com/sirupsen/logrus"
 
@@ -15,15 +16,31 @@ import (
 	operatorv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/internal/cli"
 	client "github.com/redhat-openshift-ecosystem/openshift-preflight/certification/internal/client"
+	"k8s.io/client-go/tools/clientcmd"
 )
 
-type OpenshiftEngine struct {
+type openshiftEngine struct {
 	KubeConfig *rest.Config
 }
 
-func (pe OpenshiftEngine) CreateNamespace(name string, opts cli.OpenshiftOptions) (*cli.OpenshiftReport, error) {
+func NewOpenshiftEngine() *cli.OpenshiftEngine {
+	var engine cli.OpenshiftEngine = &openshiftEngine{}
+	return &engine
+}
 
-	k8sClientset, err := kubernetes.NewForConfig(pe.KubeConfig)
+func (oe *openshiftEngine) Setup() error {
+	k8sconfig, err := clientcmd.BuildConfigFromFlags("", os.Getenv("KUBECONFIG"))
+	if err != nil {
+		log.Error("unable to parse kubecofig: ", err)
+		return err
+	}
+	oe.KubeConfig = k8sconfig
+	return nil
+}
+
+func (oe *openshiftEngine) CreateNamespace(name string, opts cli.OpenshiftOptions) (*corev1.Namespace, error) {
+
+	k8sClientset, err := kubernetes.NewForConfig(oe.KubeConfig)
 
 	if err != nil {
 		log.Error("unable to obtain k8s client: ", err)
@@ -42,24 +59,18 @@ func (pe OpenshiftEngine) CreateNamespace(name string, opts cli.OpenshiftOptions
 		Create(context.Background(), nsSpec, metav1.CreateOptions{})
 
 	if err != nil {
-		log.Error(fmt.Sprintf("error while creating Namespace %s:", name), err)
-		return &cli.OpenshiftReport{
-			Stdout: "",
-			Stderr: err.Error(),
-		}, err
+		log.Error(fmt.Sprintf("error while creating Namespace %s: ", name), err)
+		return nil, err
 	}
 
 	log.Debug("Namespace created: ", name)
 	log.Trace("Received Namespace object from API server: ", resp)
 
-	return &cli.OpenshiftReport{
-		Stdout: nsSpec.String(),
-		Stderr: "",
-	}, nil
+	return resp, nil
 }
 
-func (pe OpenshiftEngine) DeleteNamespace(name string, opts cli.OpenshiftOptions) error {
-	k8sClientset, err := kubernetes.NewForConfig(pe.KubeConfig)
+func (oe *openshiftEngine) DeleteNamespace(name string, opts cli.OpenshiftOptions) error {
+	k8sClientset, err := kubernetes.NewForConfig(oe.KubeConfig)
 
 	if err != nil {
 		log.Error("unable to obtain k8s client: ", err)
@@ -71,8 +82,8 @@ func (pe OpenshiftEngine) DeleteNamespace(name string, opts cli.OpenshiftOptions
 		Delete(context.Background(), name, metav1.DeleteOptions{})
 }
 
-func (pe OpenshiftEngine) GetNamespace(name string) (*corev1.Namespace, error) {
-	k8sClientset, err := kubernetes.NewForConfig(pe.KubeConfig)
+func (oe *openshiftEngine) GetNamespace(name string) (*corev1.Namespace, error) {
+	k8sClientset, err := kubernetes.NewForConfig(oe.KubeConfig)
 
 	if err != nil {
 		log.Error("unable to obtain k8s client: ", err)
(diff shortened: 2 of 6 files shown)


certification/internal/policy/operator/deployable_by_olm.go
```diff
@@ -231,50 +216,49 @@ func (p *DeployableByOlmCheck) installedCSV(operatorData OperatorData, openshift
...
 		log.Error("unable to retrieve the namespace")
 	}
 	p.writeToFile(ns, operatorData.InstallNamespace, "namespace")
 
 	log.Trace("Deleting the resources created by Check")
-	openshiftengine.DeleteSubscription(operatorData.App, cli.OpenshiftOptions{Namespace: operatorData.InstallNamespace})
-	openshiftengine.DeleteCatalogSource(operatorData.App, cli.OpenshiftOptions{Namespace: catalogSourceNS})
-	openshiftengine.DeleteOperatorGroup(operatorData.App, cli.OpenshiftOptions{Namespace: operatorData.InstallNamespace})
-	openshiftengine.DeleteNamespace(operatorData.InstallNamespace, cli.OpenshiftOptions{})
+	p.OpenshiftEngine.DeleteSubscription(operatorData.App, cli.OpenshiftOptions{Namespace: operatorData.InstallNamespace})
```
bcrochet, line 248: Should all of these be deferred? Is that a best practice?

certification/internal/policy/operator/deployable_by_olm.go
```diff
@@ -231,50 +216,49 @@ func (p *DeployableByOlmCheck) installedCSV(operatorData OperatorData, openshift
...
 	yamlData, err := yaml.Marshal(data)
 	if err != nil {
 		log.Error("unable to serialize the data")
+		return err
 	}
-
-	if err = os.WriteFile(filepath.Join("artifacts", fmt.Sprintf("%s-%s.yaml", resource, resourceType)), yamlData, 0644); err != nil {
-		log.Error("failed to write the k8s object to the file")
-	}
+	_, err = fileutils.WriteFileToArtifactsPath(fmt.Sprintf("%s-%s.yaml", resource, resourceType), string(yamlData))
```
bcrochet, line 260: Just a comment. Great use of that utility. :)

certification/internal/policy/operator/deployable_by_olm_test.go
```diff
@@ -0,0 +1,71 @@
...
+		err = os.Mkdir(metadataDir, 0755)
+		Expect(err).ToNot(HaveOccurred())
+		err = os.WriteFile(filepath.Join(metadataDir, annotationFilename), []byte(annotations), 0644)
+		Expect(err).ToNot(HaveOccurred())
+	})
+	Describe("When deploying an operator using OLM", func() {
+		Context("When CSV has been created successfully", func() {
+			It("Should pass Validate", func() {
+				log.Info("************ " + os.Getenv("PFLT_INDEXIMAGE"))
+				ok, err := deployableByOLMCheck.Validate(migration.ImageToImageReference("dummy/image"))
```
bcrochet, line 59: The migration struct doesn't exist anymore. Everything should be using certification.ImageReference now

cli/openshift_deprecated.go
```diff
@@ -0,0 +1,28 @@
+package cli
```
bcrochet, line 1: This file shouldn't be in cli. It should be in internal/cli now. I think you need to rebase.

bcrochet (changes requested):
A few things to start. Looks like a rebase is necessary.

certification/internal/policy/operator/deployable_by_olm.go
samira-barouti: The cleanUp() method (where all these deletions are taking place) is a deferred method. From the code:
```
	err = p.setUp(*operatorData)
	defer p.cleanUp(*operatorData)
```

certification/internal/policy/operator/deployable_by_olm_test.go
samira-barouti: I'll upload a new patch in which I fixed this.

cli/openshift_deprecated.go
samira-barouti: This will be removed. I did not rebase my code and there are old codes.

certification/internal/cli/openshift.go
```diff
@@ -37,21 +35,21 @@ type OperatorGroupData struct {
 }
 
 type OpenshiftEngine interface {
-	CreateNamespace(name string, opts OpenshiftOptions, config *rest.Config) (*OpenshiftReport, error)
```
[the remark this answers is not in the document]
samira-barouti: I see... so you mean having a separate interface for each resource? I can do that, but these are all openshift-related methods.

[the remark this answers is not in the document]
bcrochet: It's not completely necessary. But is more idiomatic golang.

certification/internal/policy/operator/deployable_by_olm.go
bcrochet: I see. That works.

certification/internal/engine/openshift.go
```diff
@@ -15,13 +16,26 @@ import (
...
 )
 
-type OpenshiftEngine struct {
+type openshiftEngine struct {
 	KubeConfig *rest.Config
 }
 
-func (pe OpenshiftEngine) CreateNamespace(name string, opts cli.OpenshiftOptions) (*cli.OpenshiftReport, error) {
+func NewOpenshiftEngine() *cli.OpenshiftEngine {
+	k8sconfig, err := clientcmd.BuildConfigFromFlags("", os.Getenv("KUBECONFIG"))
```
bcrochet, line 27: Should we be taking into account the possibility of running within a cluster? Doesn't the BuildConfig change? Maybe attempt to use the in-cluster first, then fall back to this?

certification/internal/cli/openshift.go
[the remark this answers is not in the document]
komish: I think I voted the other way when this was initially getting written. 

In fairness, I think I expected a `CreateResource` instead of per-resource method definitions, but I still think that all of them share the same context (rest config, openshift options), so to me, the way it sits here makes sense.

I'm fine going a different direction if we need to, but I don't think this should block this PR from merging. 

Just my 2c. I'm happy to discuss.

certification/internal/engine/openshift.go
komish: The in-cluster case assumes the cluster under test is the same cluster where preflight runs. I think we've historically discouraged this (because we can't guarantee the two processes won't cross paths... they shouldn't, but well).

certification/internal/engine/openshift.go
```diff
@@ -58,7 +72,7 @@ func (pe OpenshiftEngine) CreateNamespace(name string, opts cli.OpenshiftOptions
 	}, nil
 }
 
-func (pe OpenshiftEngine) DeleteNamespace(name string, opts cli.OpenshiftOptions) error {
+func (pe openshiftEngine) DeleteNamespace(name string, opts cli.OpenshiftOptions) error {
```
komish, line 75: NIT: should this be `oe` instead of `pe`?

certification/internal/engine/openshift.go
```diff
@@ -183,7 +189,7 @@ func (pe OpenshiftEngine) CreateCatalogSource(data cli.CatalogSourceData, opts c
 	}, nil
 }
 
-func (pe OpenshiftEngine) DeleteCatalogSource(name string, opts cli.OpenshiftOptions) (*cli.OpenshiftReport, error) {
+func (pe openshiftEngine) DeleteCatalogSource(name string, opts cli.OpenshiftOptions) (*cli.OpenshiftReport, error) {
 	crdClient, err := client.CatalogSourceClient(pe.KubeConfig, opts.Namespace)
```
komish, line 193: NIT: `crdcClient` seems like a copy-paste ism, and probably should be something else.

certification/internal/engine/openshift.go
bcrochet: That's fair.

bcrochet, line 27: It appears that this is causing an issue in the unit tests. This will have to be done elsewhere, or don't use this in the unit test and just instantiate the openshiftEngine directly.

bcrochet (changes requested):

komish (changes requested):
Thanks @samira-barouti! Couple of comments left in-line. And as already mentioned by @bcrochet, the unit tests fail with this PR.

certification/internal/engine/openshift.go
samira-barouti: Done :)

certification/internal/cli/openshift.go
```diff
@@ -37,21 +35,23 @@ type OperatorGroupData struct {
...
-	DeleteNamespace(name string, opts OpenshiftOptions, config *rest.Config) error
-	GetNamespace(name string, config *rest.Config) (*corev1.Namespace, error)
+	Setup() error
 
-	CreateOperatorGroup(data OperatorGroupData, opts OpenshiftOptions, config *rest.Config) (*OpenshiftReport, error)
-	DeleteOperatorGroup(name string, opts OpenshiftOptions, config *rest.Config) (*OpenshiftReport, error)
-	GetOperatorGroup(name string, opts OpenshiftOptions, config *rest.Config) (*operatorv1.OperatorGroup, error)
+	CreateNamespace(name string, opts OpenshiftOptions) (*OpenshiftReport, error)
```
bcrochet, line 40: It appears that the OpenShiftReport is being ignored for the most part. I think we need to think about this interface a bit more. OpenshiftReport doesn't make sense (with Stdout, and Stderr), since that's not what the implementation is meant to do anyway. We aren't using an external cli here. What should be returned from a Create? The entiry that was created? Just the error?

bcrochet: Looking at the fake engine that you created, it does seem like Create should return the object, and an error. The deletes are inconsistent... Some just return error, others return the OpenshiftReport (which again, doesn't really make sense).

bcrochet (changes requested):

certification/internal/cli/openshift.go
samira-barouti: Thank you Brad for the feedbacks. I am with you on removing the OpenShiftReport all together and redesign the interface. My initial intention was to be consistent with the other engines, but we probably don't need this for Openshift. I am open to discuss this further.

bcrochet: I'm fine with going with it as is, and iterating on it. I think the longer-term design shouldn't be rushed.

bcrochet (approved):
/lgtm

We can iterate on the interface/API later.

certification/internal/cli/openshift.go
samira-barouti: I removed the OpenshiftReport type and made the methods consistent. We will add it back if needed.

certification/internal/policy/operator/deployable_by_olm_test.go
```diff
@@ -0,0 +1,78 @@
...
+	BeforeEach(func() {
+		fakeImage := fakecranev1.FakeImage{}
+		imageRef.ImageInfo = &fakeImage
+
+		// set env var for index image
+		os.Setenv("PFLT_INDEXIMAGE", "test_indeximage")
+		os.Setenv("PFLT_ARTIFACTS", metadataDir)
+
+		// mock bundle directory
+		err = os.Mkdir(metadataDir, 0755)
```
bcrochet, line 41: When I run this locally, I'm getting an error that it already exists. This should use os.MkdirTemp, and clean it up in AfterEach

bcrochet (changes requested):
Issue with the tests

certification/internal/policy/operator/deployable_by_olm_test.go
samira-barouti: There is an afterEach block that deleted the metadata directory. I'll try to create the resources under /tmp though.

certification/internal/policy/operator/deployable_by_olm_test.go
```diff
@@ -0,0 +1,80 @@
...
+		annotationFilename = "annotations.yaml"
+		annotations        = `annotations:
+  operators.operatorframework.io.bundle.package.v1: testPackage
+  operators.operatorframework.io.bundle.channel.default.v1: testChannel
+`
+	)
+
+	BeforeEach(func() {
+		// mock bundle directory
+		tmpDir, err := os.MkdirTemp("", "bundle-metadata-*")
```
bcrochet, line 31: Still need AfterEach at this level to clean it up.

bcrochet (changes requested):

certification/internal/policy/operator/deployable_by_olm_test.go
samira-barouti: uh gotcha! Good catch!

certification/internal/engine/openshift.go
```diff
@@ -15,15 +16,31 @@ import (
...
+	k8sconfig, err := clientcmd.BuildConfigFromFlags("", os.Getenv("KUBECONFIG"))
+	if err != nil {
+		log.Error("unable to parse kubecofig: ", err)
+		return err
+	}
+	oe.KubeConfig = k8sconfig
+	return nil
+}
+
+func (oe *openshiftEngine) CreateNamespace(name string, opts cli.OpenshiftOptions) (*corev1.Namespace, error) {
```
komish, line 41: Given that `Setup()` doesn't take any parameters, would it make sense to just perform the tasks in that method in the `NewOpenshiftEngine`?

bcrochet (changes requested):
I'm seeing this:

time="2021-09-15T13:12:38-04:00" level=debug msg="Waiting for Subscription.status.installedCSV to become ready..."
time="2021-09-15T13:12:38-04:00" level=debug msg="Subscription.status.installedCSV is not set yet, retrying..."
time="2021-09-15T13:12:40-04:00" level=debug msg="Waiting for Subscription.status.installedCSV to become ready..."
time="2021-09-15T13:12:40-04:00" level=debug msg="Subscription.status.installedCSV is not set yet, retrying..."
time="2021-09-15T13:12:42-04:00" level=debug msg="Waiting for Subscription.status.installedCSV to become ready..."
time="2021-09-15T13:12:42-04:00" level=debug msg="Subscription.status.installedCSV is not set yet, retrying..."
time="2021-09-15T13:12:44-04:00" level=debug msg="Waiting for Subscription.status.installedCSV to become ready..."
time="2021-09-15T13:12:44-04:00" level=debug msg="Subscription.status.installedCSV is not set yet, retrying..."
time="2021-09-15T13:12:46-04:00" level=debug msg="Waiting for Subscription.status.installedCSV to become ready..."
time="2021-09-15T13:12:46-04:00" level=debug msg="Subscription.status.installedCSV is not set yet, retrying..."
time="2021-09-15T13:12:48-04:00" level=debug msg="Waiting for Subscription.status.installedCSV to become ready..."

That doesn't seem like it should be occurring.

samira-barouti:
@bcrochet This is the expected behavior when the second test is running. The test ensures that Validate() returns an error if the check is unable to fetch the Subscription.status.installedCSV.

certification/internal/engine/openshift.go
samira-barouti: I initially did this. But it causes the tests in engine/ folder to fail because `KUBECONFIG` has not been set and `clientcmd.BuildConfigFromFlags` returns nil dereference (engine.go registers and initializes all checks). The way it is done now, lets us to mock the call to Setup(). I will be glad to revert it back to what it was before if there is any way to mock this.

bcrochet:
> @bcrochet This is the expected behavior when the second test is running. The test ensures that Validate() returns an error if the check is unable to fetch the Subscription.status.installedCSV.

It just hangs for a bit. I don't think we want to introduce an unnecessary delay to the unit tests.

samira-barouti:
> > @bcrochet This is the expected behavior when the second test is running. The test ensures that Validate() returns an error if the check is unable to fetch the Subscription.status.installedCSV.
> 
> It just hangs for a bit. I don't think we want to introduce an unnecessary delay to the unit tests.

Yes. The timeout duration (which is 180 sec) is hard coded and the unit test is hanging until the context is timed-out. Parametrizing the timeout duration might be an overkill.

bcrochet:
> > > @bcrochet This is the expected behavior when the second test is running. The test ensures that Validate() returns an error if the check is unable to fetch the Subscription.status.installedCSV.
> > 
> > 
> > It just hangs for a bit. I don't think we want to introduce an unnecessary delay to the unit tests.
> 
> Yes. The timeout duration (which is 180 sec) is hard coded and the unit test is hanging until the context is timed-out. Parametrizing the timeout duration might be an overkill.

I don't want to introduce a 180 second pause on every unit test run. That will add up very quickly throughout the day.

samira-barouti:
> > > > @bcrochet This is the expected behavior when the second test is running. The test ensures that Validate() returns an error if the check is unable to fetch the Subscription.status.installedCSV.
> > > 
> > > 
> > > It just hangs for a bit. I don't think we want to introduce an unnecessary delay to the unit tests.
> > 
> > 
> > Yes. The timeout duration (which is 180 sec) is hard coded and the unit test is hanging until the context is timed-out. Parametrizing the timeout duration might be an overkill.
> 
> I don't want to introduce a 180 second pause on every unit test run. That will add up very quickly throughout the day.

Ok.. Are you suggesting to leveraging viper to pass the timeout duration?

bcrochet:
> > > > > @bcrochet This is the expected behavior when the second test is running. The test ensures that Validate() returns an error if the check is unable to fetch the Subscription.status.installedCSV.
> > > > 
> > > > 
> > > > It just hangs for a bit. I don't think we want to introduce an unnecessary delay to the unit tests.
> > > 
> > > 
> > > Yes. The timeout duration (which is 180 sec) is hard coded and the unit test is hanging until the context is timed-out. Parametrizing the timeout duration might be an overkill.
> > 
> > 
> > I don't want to introduce a 180 second pause on every unit test run. That will add up very quickly throughout the day.
> 
> Ok.. Are you suggesting to leveraging viper to pass the timeout duration?

Not at all. There are other ways. I'm testing something out now. I'm not sure of the solution quite yet.

bcrochet:
> > > > > > @bcrochet This is the expected behavior when the second test is running. The test ensures that Validate() returns an error if the check is unable to fetch the Subscription.status.installedCSV.
> > > > > 
> > > > > 
> > > > > It just hangs for a bit. I don't think we want to introduce an unnecessary delay to the unit tests.
> > > > 
> > > > 
> > > > Yes. The timeout duration (which is 180 sec) is hard coded and the unit test is hanging until the context is timed-out. Parametrizing the timeout duration might be an overkill.
> > > 
> > > 
> > > I don't want to introduce a 180 second pause on every unit test run. That will add up very quickly throughout the day.
> > 
> > 
> > Ok.. Are you suggesting to leveraging viper to pass the timeout duration?
> 
> Not at all. There are other ways. I'm testing something out now. I'm not sure of the solution quite yet.

Make a var at the top of the file (var csvTimeout time.Duration = 180), and override it in the test.

komish:
> > > > > > > @bcrochet This is the expected behavior when the second test is running. The test ensures that Validate() returns an error if the check is unable to fetch the Subscription.status.installedCSV.
> > > > > > 
> > > > > > 
> > > > > > It just hangs for a bit. I don't think we want to introduce an unnecessary delay to the unit tests.
> > > > > 
> > > > > 
> > > > > Yes. The timeout duration (which is 180 sec) is hard coded and the unit test is hanging until the context is timed-out. Parametrizing the timeout duration might be an overkill.
> > > > 
> > > > 
> > > > I don't want to introduce a 180 second pause on every unit test run. That will add up very quickly throughout the day.
> > > 
> > > 
> > > Ok.. Are you suggesting to leveraging viper to pass the timeout duration?
> > 
> > 
> > Not at all. There are other ways. I'm testing something out now. I'm not sure of the solution quite yet.
> 
> Make a var at the top of the file (var csvTimeout time.Duration = 180), and override it in the test.

With this, this PR  looks good.

komish (approved):
/lgtm

PR #234, opened 2021-09-10
Remove unneeded deps from docs

komish:
Just going through things and seeing some references to deps in our docs that we've removed. There might be more, but these are the ones of which I was fairly certain.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,9 +23,7 @@ RUN dnf install -y \
       gzip \
       iptables \
       findutils \
-      openscap-scanner \
       podman \
-      buildah \
       skopeo \
     && dnf clean all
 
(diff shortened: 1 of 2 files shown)


bcrochet (approved):
/lgtm

Dockerfile
```diff
@@ -25,7 +25,6 @@ RUN dnf install -y \
       findutils \
       openscap-scanner \
```
samira-barouti, line 26: This could be removed as well; we no longer support HasMinimalVulns check

komish: Fair! I was trying to be cautious here because I thought we might reintroduce it at some point. Either way, I've removed this dependency - we can re-add it later if we need it.

acornett21 (approved):

sebrandon1 (approved):

PR #244, opened 2021-09-13
Introduce an interface to list tags

### LINKED ISSUE #243: HasUniqueTag check needs to migrate away from Skopeo
## Bug Description
This is the last check that does not use Crane directly. It will either need access to Crane (to call ListTags) or have the tags passed in via the ImageReference.

bcrochet:
The implementation is backed by Crane, but is used via an interface that
abstracts Crane, so that the implementation could be something else, should
that be necessary, without having to change the check.

Fixes #243


acornett21 (approved):

certification/internal/service/service.go
```diff
@@ -0,0 +1,5 @@
+package service
+
+type TagLister interface {
```
samira-barouti, line 3: IMO, using a generic name might be enable us to extend this interface for the future. I cannot think of a name that I like enough, but sth along the line of RepositoryService which includes all repo-related operations might be ok.

certification/internal/policy/container/has_unique_tag.go
```diff
@@ -4,44 +4,33 @@ import (
...
-		log.Error("unable to execute skopeo on the image: ", err)
-		return nil, err
-	}
-
-	return runReport.Tags, nil
+	return p.TagLister.ListTags(image)
 }
 
-func (p *HasUniqueTagCheck) validate(tags []string) (bool, error) {
+func (p *HasUniqueTagCheck) validate(image string, tags []string) (bool, error) {
```
samira-barouti, line 33: We are not using `image`; any reason we added this var?

certification/internal/service/service.go
bcrochet: This is idiomatic golang. Small interfaces that are generic. This IMO is a generic name. See all the io interfaces, like Reader, Writer, Closer. They contain one method. But they might all be implemented together.

certification/internal/policy/container/has_unique_tag.go
bcrochet: That may have come over from the old implementation. Good catch. I'll remove it.

samira-barouti (approved):
/lgtm

acornett21 (approved):

PR #246, opened 2021-09-13
Reorganize code for clarity

bcrochet:
Some of the code was put (likely prematurely) into a "utils" package, even
if just a single package was using it. The moves that "utility" code closer
to the consumer, and in some cases, no longer exported.

This also moves some of the truly utility functions into a more sane package
(such as artifacts) that are shared, but make the package name more
descriptive, and therby allowing a shortening of the function name itself.


acornett21 (approved):

certification/artifacts/artifacts.go
```diff
@@ -9,6 +10,16 @@ import (
 	"github.com/spf13/viper"
 )
 
+func WriteFile(filename, contents string) (string, error) {
```
komish, line 13: Nit: Could you add comments to this?

certification/artifacts/artifacts.go
```diff
@@ -28,7 +39,7 @@ func createArtifactsDir(artifactsDir string) (string, error) {
 	return artifactsDir, nil
 }
 
-func ArtifactPath() string {
+func Path() string {
```
komish, line 42: Nit: Could you add comments to this?

certification/internal/policy/operator/deployable_by_olm.go
```diff
@@ -76,26 +75,26 @@ func (p *DeployableByOlmCheck) newOpenshiftEngine() (*openshiftengine.OpenshiftE
...
-	annotations, err := containerutil.GetAnnotationsFromBundle(bundleRef.ImageFSPath)
+	annotations, err := getAnnotationsFromBundle(bundleRef.ImageFSPath)
 
 	if err != nil {
 		log.Errorf("unable to get annotations.yaml from the bundle")
 		return nil, err
 	}
 
-	catalogImage, err := viperutil.GetString(indexImageKey)
+	catalogImage, err := getString(indexImageKey)
```
komish, line 85: This function call seems rather hard to follow to me. If I'm reading the code, I'd have a hard time deriving why I needed a function call to get a string, and this no longer has any context as to the Viper configuration side of the house. Is there any reason why we why we didn't just handle this config-key-empty-value case in-line?

certification/internal/policy/operator/operator.go
```diff
@@ -0,0 +1,64 @@
...
+	var bundleMeta metadata
+	if err := yaml.Unmarshal(annotationBytes, &bundleMeta); err != nil {
+		log.Error("metadata/annotations.yaml found but is malformed")
+		return nil, err
+	}
+
+	return bundleMeta.Annotations, nil
+}
+
+// Annotation() accepts the annotations map and searches for the specified annotation corresponding
```
komish, line 53: Nit: this doc reference should be `annotation`from `Annotation()`

certification/internal/policy/operator/operator_pkg_name_uniqueness.go
```diff
@@ -35,7 +34,7 @@ type packageData struct {
 type OperatorPkgNameIsUniqueCheck struct{}
 
 func (p *OperatorPkgNameIsUniqueCheck) Validate(bundleRef certification.ImageReference) (bool, error) {
-	annotations, err := containerutils.GetAnnotationsFromBundle(bundleRef.ImageFSPath)
+	annotations, err := getAnnotationsFromBundle(bundleRef.ImageFSPath)
```
komish, line 37: So these were initially in util-like packages _because_ they could be used regardless of engine. Moving them into this check means only this check can use it. We probably don't have any other case right now where we can reuse these functions, but this organizational move _does_ limit our ability to reuse this. 

Do you think we reconsider collapsing methods like this one into this check (where it's currently used) and instead just consider a more proper name and organization for the package?

komish (changes requested):
Requested a few changes in comments, and also opened up a couple of topics for discussion.

The `artifacts` package is 💯

certification/artifacts/artifacts.go
```diff
@@ -9,6 +10,16 @@ import (
 	"github.com/spf13/viper"
 )
 
+func WriteFile(filename, contents string) (string, error) {
+	fullFilePath := filepath.Join(Path(), filename)
+
+	err := ioutil.WriteFile(fullFilePath, []byte(contents), 0644)
```
sebrandon1, line 16: `ioutil.WriteFile` is [deprecated/moved](https://golang.org/doc/go1.16#ioutil) in Go 1.16.  Use `os.WriteFile`.

sebrandon1 (changes requested):
One note about deprecation.

certification/artifacts/artifacts.go
bcrochet: Will do.

certification/artifacts/artifacts.go
bcrochet: Will do.

certification/internal/policy/operator/deployable_by_olm.go
bcrochet: @samira-barouti would have to answer that. I'm not sure what the original intention of the function is.

certification/internal/policy/operator/operator.go
bcrochet: Will do.

certification/internal/policy/operator/operator_pkg_name_uniqueness.go
bcrochet: It's not in the check itself. It's package level. I'm pretty sure it's used in two different checks.

If it does need to be used by, say, a container check, then we might consider moving it to its own package. But for now, it's just an unexported, package-level function defined in operator.go.

The point of a lot of this is that I think we jumped to some of the utils-like stuff too quickly, rather than moving them when we do actually need them in other places.

certification/artifacts/artifacts.go
bcrochet: Thanks. I'll fix that.

certification/internal/policy/operator/operator_pkg_name_uniqueness.go
komish: Yup, okay, brain is lagging behind - I read too much into the file name and forgot that those don't really matter.

With that said - I think there's still an opportunity to move this somewhere _else_ where it might make more sense. We can move forward for now and re-organize this when it needs to be done.

Feel free to Resolve Conversation on this point.

certification/internal/policy/operator/deployable_by_olm.go
samira-barouti: DeployableByOLM initially read many values from viper; that's why I had to manage the error handling outside the code. Right now, we only read `PFLT_INDEXIMAGE` which makes this code unnecessary.

bcrochet: I inlined it. Thanks for the clarification.

certification/internal/engine/engine.go
```diff
@@ -82,7 +87,7 @@ func (c *CraneEngine) ExecuteChecks() error {
 	}()
 
 	log.Debug("extracting container filesystem to ", containerFSPath)
-	err = fileutils.Untar(containerFSPath, r)
```
samira-barouti, line 85: HasMinimalVulns check is also using this method to extract the oval file. Not sure if there is any plan to add it back.

bcrochet: We can address that when we get there.

komish (approved):
/lgtm

samira-barouti (approved):

PR #247, opened 2021-09-13
switching to use UBI 8 base image

### LINKED ISSUE #241: Consider migrating to UBI
We no longer depend on podman, so we probably don't need to ship the preflight container based on the podman image anymore.

acornett21:
Switching to use UBI 8 as the base image. 

Tested this running as a container in OpenShift for both `check container` and `check operator` 

fixed #241

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,8 @@ WORKDIR /go/src/preflight
 RUN make build
 
 # podman:v3.2.2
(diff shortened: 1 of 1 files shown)


samira-barouti (approved):

jomkz (approved):
/lgtm

komish:
Thanks @acornett21!

PR #248, opened 2021-09-13
add more detailed use examples

komish:
This PR just adds a docs file RECIPES.md that discusses a few ways to use preflight (Binary, Podman, Kubernetes) with manifests and command invocations to use as a starting point.

### DIFF
--- a/docs/RECIPES.md
+++ b/docs/RECIPES.md
@@ -1,0 +1,136 @@
+# Preflight Usage Examples
+
+Below are detailed examples on how to leverage `preflight` in various
+environments.
+
+These examples are shown using the Operator policy against an operator bundle
+(e.g. `preflight check operator <bundle`>), but the same concepts can be applied
+to the Container policy (give or take a few configuration changes).
+
+You will also need an index image containing your bundle for each of these approaches.
+See [DOCS](BUILDING_AN_INDEX.md)
+
+## As a Binary on Your Workstation
+
+To run `preflight` on your workstation, you'll first need to download and
+[install](../README.md#Installation)
+the binary to your path.
+
+You will also need:
+
+- Your bundle image published to a container registry,
+- An Index Image containing your operator's new bundle published to a container
+  registry.
+- A Kubeconfig for a user with cluster-admin privileges to an OpenShift 4.5+
+  Cluster running Operator Lifecycle Manager.
+
+The `preflight` tool will use the above to execute the Operator policy against
+your test asset. With the above in hand, you can execute preflight against your
+test asset by executing the following.
+
+```bash
+export KUBECONFIG=/path/to/your/kubeconfig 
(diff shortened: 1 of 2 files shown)


acornett21 (approved):

samira-barouti (approved):
/lgtm

docs/RECIPES.md
```diff
@@ -0,0 +1,136 @@
...
+(e.g. `preflight check operator <bundle`>), but the same concepts can be applied
+to the Container policy (give or take a few configuration changes).
+
+You will also need an index image containing your bundle for each of these approaches.
+See [DOCS](BUILDING_AN_INDEX.md)
+
+## As a Binary on Your Workstation
+
+To run `preflight` on your workstation, you'll first need to download and
+[install](../README.md#Installation)
```
bcrochet, line 16: Why is this wrapped weird?

docs/RECIPES.md
```diff
@@ -0,0 +1,136 @@
...
+the binary to your path.
+
+You will also need:
+
+- Your bundle image published to a container registry,
+- An Index Image containing your operator's new bundle published to a container
+  registry.
+- A Kubeconfig for a user with cluster-admin privileges to an OpenShift 4.5+
```
bcrochet, line 24: Is this true? Does it have to be cluster-admin? My testing has shown that to not be the case.

docs/RECIPES.md
```diff
@@ -0,0 +1,136 @@
...
+  quay.io/opdev/preflight:stable check operator registry.example.org/your-namespace/your-bundle-image:sometag
+```
+
+## As a Job In OpenShift (or Kubernetes)
+
+You should be able to run `preflight` as a job in OpenShift without requiring
+additional privileges or security context constraints.
+
+As with previous examples, you will need to provide `preflight`a Kubeconfig
+mapping a user with cluster-admin privileges to an OpenShift cluster that can be
```
bcrochet, line 74: Same as above. I created a standard SA, with a rolebinding to the 'edit' role, then passed that in as PFLT_SERVICEACCOUNT, and a namespace via PFLT_NAMESPACE.

bcrochet (changes requested):
Couple of comments.

docs/RECIPES.md
komish: Markdown linting was likely complaining about line length. When rendered, they should collapse to a single line.

docs/RECIPES.md
komish: I'll address this in the other comment related to the same topic 👍

docs/RECIPES.md
bcrochet: Actually, no. I think that may be why it was failing for me. I hadn't circled back to see why it was failing.

bcrochet (approved):
/lgtm

PR #250, opened 2021-09-13
log the version info immediately

komish:
If we happen to run into an issue with the preflight execution, we may not get the version of preflight that experienced the issue which is normally stored in results (e.g. if we run into a panic at runtime). 

This adds a log line right away that logs the version context to both the `check operator` and `check container` runs.

### DIFF
--- a/cmd/check_container.go
+++ b/cmd/check_container.go
@@ -12,6 +12,8 @@ import (
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/formatters"
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/runtime"
 	certutils "github.com/redhat-openshift-ecosystem/openshift-preflight/certification/utils"
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/version"
+	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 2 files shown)


acornett21 (approved):

sebrandon1 (approved):

cmd/check_container.go
```diff
@@ -30,6 +32,8 @@ var checkContainerCmd = &cobra.Command{
 		// Expect exactly one positional arg. Check here instead of using builtin Args key
 		// so that we can get a more user-friendly error message
 
+		log.Info("certification library version ", version.Version.String())
```
bcrochet, line 35: Nit: This could just go at the end of the preRunConfig. That would log it very early, and only need to be in one place.

bcrochet (changes requested):
Small nit. But ignore if  you like.

cmd/check_container.go
komish: I was thinking that might work better, but then it would show up in the Certify and Support workflows, and I'm not sure we want it to be there.

https://github.com/redhat-openshift-ecosystem/openshift-preflight/blob/main/cmd/certify.go#L13
https://github.com/redhat-openshift-ecosystem/openshift-preflight/blob/main/cmd/support.go#L30

bcrochet (approved):

PR #251, opened 2021-09-13
Add unit tests

sebrandon1:


certification/formatters/generic_test.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+
+	for _, tc := range testCases {
+
+		// Patch the function if we expect an error
+		if tc.marshalIndentFailure {
+			xmlMarshalIndent = func(v interface{}, prefix, indent string) ([]byte, error) {
+				return nil, errors.New("this is an error")
+			}
+		} else {
```
bcrochet, line 140: Nit: Could get rid of the else clause by assigning to xmlMarshalIndent ahead of the if statement. Feel free to ignore.

certification/formatters/generic_test.go
```diff
@@ -0,0 +1,168 @@
+package formatters
```
bcrochet, line 1: Any particular reason to not use ginkgo/gomega? The rest of the code base is being tested with that framework.

bcrochet (commented):
Small nit and a question. Otherwise, thanks for the contrib!

certification/formatters/generic_test.go
sebrandon1: I can re-write in Ginkgo if desired.  Didn't know it was a requirement.

bcrochet: Not a hard requirement. Just curious, really.

sebrandon1: I just don't prefer the Ginkgo syntax but I'm good with it either way.

komish: Yep, I think we probably prefer to keep things consistent, but this doesn't introduce any other frameworks other than gotest, and it's a relatively small package. I could go either way. I'm fine accepting it as is.

komish (approved):
Thanks @sebrandon1 

/lgtm

bcrochet (approved):
I'm good as is.
/lgtm

certification/formatters/generic_test.go
sebrandon1: Thanks!

PR #255, opened 2021-09-13
Fix service account in bundle-passes

bcrochet:
The latest operator-sdk bundle validate requires that the service account
not be the same as what's in the CSV. This fixes that.

### DIFF
--- a/test/containerfiles/successful-bundle-assets/manifests/sample-operator-controller-manager_v1_serviceaccount.yaml
+++ b/test/containerfiles/successful-bundle-assets/manifests/sample-operator-controller-manager_v1_serviceaccount.yaml
@@ -2,4 +2,4 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   creationTimestamp: null
-  name: sample-operator-controller-manager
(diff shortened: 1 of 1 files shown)


komish (approved):
I wonder if this is because OLM automatically creates the service account detailed in the CSV, so the extra file isn't needed. 

Either way
/lgtm

bcrochet:
> I wonder if this is because OLM automatically creates the service account detailed in the CSV, so the extra file isn't needed.
> 
> Either way
> /lgtm

This was the error that started popping up:

{
    "passed": false,
    "outputs": [
        {
            "type": "error",
            "message": "Error: Value sample-operator-controller-manager: invalid service account found in bundle. sa name cannot match service account defined for deployment spec in CSV"
        },
        {
            "type": "warning",
            "message": "Warning: Value : (sample-operator.v0.0.1) csv.Spec.minKubeVersion is not informed. It is recommended you provide this information. Otherwise, it would mean that your operator project can be distributed and installed in any cluster version available, which is not necessarily the case for all projects."
        }
    ]
}

acornett21 (approved):

PR #257, opened 2021-09-14
switch from podman to none in operator sdk bundle validate cmd

### LINKED ISSUE #256: move away from using podman in operatork-sdk bundle validate cmd
Currently for `ValidateOperatorBundle` preflight is passing in `podman` as an argument, even though the bundle has already been pulled/extracted/etc. Since the bundle files are already on disk, we can use `none` as an argument.

acornett21:
Since we already pulled and extracted the bundle file, we can pass `none` as a cmd argument to `operator-sdk` and get the same results. Thus, one step closer to potentially removing podman from the docker image.

Test Results
```
time="2021-09-14T17:39:08Z" level=info msg="running check: ValidateOperatorBundle"
time="2021-09-14T17:39:08Z" level=debug msg="Command being run: [operator-sdk bundle validate -b none --output json-alpha1 --select-optional name=community --select-optional name=operatorhub --verbose /tmp/preflight-3006179402/fs]"
time="2021-09-14T17:39:09Z" level=info msg="check completed: ValidateOperatorBundle" result=PASSED
```

fixes #256


jomkz (approved):
/lgtm

bcrochet (approved):
/lgtm

PR #258, opened 2021-09-14
Set an expiration of 1w for snapshot builds

bcrochet:
The amount of quay.io tags has built up, and they are just sha tags
that don't really have much value the older they get. Let's set an
auto-expiry via build-args. The default is to never expire, which
is what would be used for a release tag.


jomkz (approved):
/lgtm

Issue #108, opened 2021-07-19
Image size is pretty big


itroyano:
## Bug Description
`REPOSITORY                                 TAG                                        IMAGE ID       CREATED          SIZE

quay.io/opdev/preflight                    f07567b6bd629c2275bb2a2d9a39d502d31c5789   e6a85a623c42   34 seconds ago   **699MB**`

## Steps to Reproduce:
make image-build

komish:
#241 might help with this.

bcrochet:
What's our cutoff? I think we're down to 200M or so. Is that small enough?

itroyano:
current image size is ~160Mb.  closing the issue.

PR #259, opened 2021-09-14
remove unused binaries from docker image

acornett21:
Removing `podman` and `skopeo` from preflights docker image, and removing commented out lines.  Switching to UBI (previous pr that was merged) and removing `podman`  and `skopeo` has decreased the image size. 

New image size:
```
quay.io/opdev/preflight                   79aead7423212ff837195777e827d07e7c67d4ce  88d4c9de8a4a  17 hours ago  448 MB
```

I'm not sure if there is more work to be done to the image to be able to close out #108, what are others thoughts?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,12 +6,10 @@ COPY . /go/src/preflight
 WORKDIR /go/src/preflight
 RUN make build
 
-# podman:v3.2.2
-#FROM quay.io/podman/stable@sha256:c281b604477e10e3e84f81af9b68db59ee616bf6a34391224e5e5c74db0428d0
+# ubi8:latest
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -24,15 +23,9 @@ RUN dnf install -y \
       gzip \
       iptables \
       findutils \
-      podman \
       skopeo \
     && dnf clean all
 
-# Install oscap-podman binary
```
samira-barouti, line 31: You can remove line 13 as well.

acornett21: Nice catch!

bcrochet (approved):
/lgtm

komish (approved):
/lgtm

PR #262, opened 2021-09-15
remove local reference to passing bundle and replace with working operator

komish:
The E2E tests previously used a "known good bundle", and the bundle Containerfile were stored in-repo. This is mostly fine, but it didn't make reference to a working operator with published assets.

This PR replaces that bundle reference with one that will need to be maintained https://github.com/opdev/simple-demo-operator. This will also run, and therefore should pass the DeployableByOLM test.


test/e2e/e2e_test.go
```diff
@@ -14,8 +14,8 @@ import (
 var _ = Describe("policy validation", func() {
 	Describe("When enforcing operator policy", func() {
 		var (
-			// TODO: replace this with a project-accessible namespace once identified.
-			goodImage = "quay.io/komish/preflight-test-bundle-passes:latest"
+			// TODO: replace the failure case with crafted examples.
+			goodImage = "quay.io/opdev/simple-demo-operator-bundle:v0.0.2"
```
samira-barouti, line 18: Out of curiosity: do we plan to have a pipeline or a manual procedure to create/update this bundle image?

samira-barouti (approved):
A small question

test/e2e/e2e_test.go
```diff
@@ -59,7 +59,8 @@ var _ = Describe("policy validation", func() {
 	Describe("When enforcing container policy", func() {
 
 		var (
-			// TODO: replace this with a project-accessible namespace once identified.
+			// TODO: replace the passing case with the container used by
+			// https://github.com/opdev/simple-demo-operator
 			goodImage = "quay.io/komish/preflight-test-container-passes:latest"
```
bcrochet, line 64: Do we want these to reference opdev?

sebrandon1 (approved):

test/e2e/e2e_test.go
komish: No, they don't exist in opdev yet =\

bcrochet (approved):
/lgtm

PR #264, opened 2021-09-16
Fix nil pointer reference in operatorMetadata

bcrochet:
A subtle bug was introduced in the operatorMetadata function in OLM check.
If the index image envvar was not set, it was returning (nil, err), but err
was not set to anything, so the wrapping check for err != nil was not
succeeding. This ensures an error is returned.


komish (approved):
/lgtm

acornett21 (approved):

PR #265, opened 2021-09-16
Fix the timeouts in DeployableByOLM check

bcrochet:
The timeouts were made into variables (for testing purposes). When this
was changed, the "*time.Second" was not added, so the context was timing
out basically immediately.

### DIFF
--- a/certification/internal/policy/operator/deployable_by_olm.go
+++ b/certification/internal/policy/operator/deployable_by_olm.go
@@ -32,8 +32,8 @@ type DeployableByOlmCheck struct {
 }
 
 var (
-	subscriptionTimeout time.Duration = 180
-	csvTimeout          time.Duration = 90
+	subscriptionTimeout time.Duration = 180 * time.Second
+	csvTimeout          time.Duration = 90 * time.Second
 )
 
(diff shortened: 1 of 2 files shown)


acornett21 (approved):

certification/internal/policy/operator/deployable_by_olm.go
```diff
@@ -193,7 +193,7 @@ func (p *DeployableByOlmCheck) installedCSV(operatorData OperatorData) (string,
 
 	// query API server for the installed CSV field of the created subscription
 	go func(parent context.Context) {
-		ctx, cancel := context.WithTimeout(parent, subscriptionTimeout)
+		ctx, cancel := context.WithTimeout(parent, subscriptionTimeout*time.Second)
```
komish, line 196: Can we replace this in the assignment instead of inline? E.g.

```Go
var subscriptionTimeout = 180 * time.Second
```
And the same for other places? Reads a bit better to me, keeping the value and unit together.

komish (changes requested):
Small change

komish (approved):
/lgtm

PR #266, opened 2021-09-21
Leverage controller-manager GetConfigOrDie() to parse kubeconfig

samira-barouti:
This PR removes the logic that fetches kubeconfig from environment variable and parse Kubeconfig file.


bcrochet (approved):
/lgtm

PR #271, opened 2021-09-22
Reintroduce the JUnit XML formatter

### LINKED ISSUE #268: Is it possible to get junitxml output format back?
In the past, preflight tool generated output both in json and junit. The plan was to use that while running preflight checks with DCI. DCI UI is able to parse junit and to display failed and passed tests in a nice way. 

Unfortunately, junit format was removed: https://github.com/redhat-openshift-ecosystem/openshift-preflight/commit/ee0ebc476e18714e4344c0a5ca58b1227966ff2a#diff-7b226e2c43cb28a869b02277ad5184ffce22c756fd9893e3b3109869e7eb50ed 

Get junit format back directly in the preflight tool.

If it's impossible to do on preflight side, we'll probably have to parse preflight json output on DCI side.

bcrochet:
A Viper boolean is introduced merely called "junit". If it is true,
it will write the results to the artifacts dir called results-junit.xml.

Set PFLT_JUNIT envvar is set to 't', 'true', 1 or set "junit": true in
config file.

Fixes #268

### DIFF
--- a/certification/formatters/junitxml.go
+++ b/certification/formatters/junitxml.go
@@ -1,0 +1,108 @@
+package formatters
+
+import (
+	"encoding/xml"
+	"fmt"
+	"time"
+
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/errors"
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/runtime"
+)
+
+type JUnitTestSuites struct {
+	XMLName xml.Name         `xml:"testsuites"`
+	Suites  []JUnitTestSuite `xml:"testsuite"`
+}
+
+type JUnitTestSuite struct {
+	XMLName    xml.Name        `xml:"testsuite"`
+	Tests      int             `xml:"tests,attr"`
+	Failures   int             `xml:"failures,attr"`
+	Time       string          `xml:"time,attr"`
+	Name       string          `xml:"name,attr"`
+	Properties []JUnitProperty `xml:"properties>property,omitempty"`
+	TestCases  []JUnitTestCase `xml:"testcase"`
+}
+
+type JUnitTestCase struct {
+	XMLName     xml.Name          `xml:"testcase"`
+	Classname   string            `xml:"classname,attr"`
+	Name        string            `xml:"name,attr"`
+	Time        string            `xml:"time,attr"`
+	SkipMessage *JUnitSkipMessage `xml:"skipped,omitempty"`
+	Failure     *JUnitFailure     `xml:"failure,omitempty"`
+	SystemOut   string            `xml:"system-out,omitempty"`
+	Message     string            `xml:",chardata"`
+}
+
+type JUnitSkipMessage struct {
+	Message string `xml:"message,attr"`
+}
+
+type JUnitProperty struct {
+	Name  string `xml:"name,attr"`
+	Value string `xml:"value,attr"`
+}
+
+type JUnitFailure struct {
+	Message  string `xml:"message,attr"`
+	Type     string `xml:"type,attr"`
+	Contents string `xml:",chardata"`
+}
+
+func junitXMLFormatter(r runtime.Results) ([]byte, error) {
+	response := getResponse(r)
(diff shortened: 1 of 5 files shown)


cmd/check.go
```diff
@@ -13,3 +18,25 @@ var checkCmd = &cobra.Command{
 func init() {
 	rootCmd.AddCommand(checkCmd)
 }
+
+func checkJunitAndWrite(cfg runtime.Config, results runtime.Results) error {
```
samira-barouti, line 22: I believe having separate methods for check and write will improve the code maintenance and readability.

certification/formatters/junitxml.go
```diff
@@ -0,0 +1,108 @@
...
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/errors"
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/runtime"
+)
+
+type JUnitTestSuites struct {
+	XMLName xml.Name         `xml:"testsuites"`
+	Suites  []JUnitTestSuite `xml:"testsuite"`
+}
+
+type JUnitTestSuite struct {
```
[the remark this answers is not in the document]
bcrochet: No. I don't want to introduce a runtime dependency on a testing framework.

cmd/check.go
bcrochet: They do the exact same thing. There is no need to separate them at the moment.

acornett21 (approved):

tkrishtop (approved):
Hi @bcrochet LGTM

cmd/check.go
```diff
@@ -13,3 +18,25 @@ var checkCmd = &cobra.Command{
...
+		junitformatter, err := formatters.NewForConfig(cfg)
+		if err != nil {
+			return err
+		}
+		junitResults, err := junitformatter.Format(results)
+		if err != nil {
+			return err
+		}
+
+		junitFilename, err := artifacts.WriteFile("results-junit.xml", string(junitResults))
```
komish, line 34: SUPER NIT (sorry): Wasn't there an extension method built into the formatter that we should use here?

cmd/check.go
komish, line 22: NIT: Can we rename this to something like `writeJunitIfIEnabled` so we know the exact case causing an action?

komish: Also, do we need `runtime.Config` to be passed in here?

cmd/check.go
```diff
@@ -13,3 +18,25 @@ var checkCmd = &cobra.Command{
 func init() {
 	rootCmd.AddCommand(checkCmd)
 }
+
+func checkJunitAndWrite(cfg runtime.Config, results runtime.Results) error {
+	if viper.GetBool("junit") {
+		cfg.ResponseFormat = "junitxml"
```
komish, line 24: Do we want to copy this config instead of changing it? Not sure it matters in our execution, but I think it might be better to preserve the original.

Might want to check my logic - on the other hand, if we're not using pointers then we might be in the clear here because we passed the cfg into a function.

komish (changes requested):
Couple of changes

cmd/check.go
bcrochet: It is being passed by value, not reference, so we are good here. That's the intent of this. It's a secondary formatter only being utilized when requested in addition to the primary formatter (which we don't allow the user to change anyway).

cmd/check.go
bcrochet: I will check into that.

cmd/check.go
bcrochet: Yes. However, it assumes the filename is "results.<ext>". I don't want to clobber the main results file. I'm ok with this one, as it 1) will always be ".xml", and 2) will always be junit.

cmd/check.go
bcrochet: So after looking at it. There's really no need to pass that in. I'll just create one and use that.

komish (approved):

PR #272, opened 2021-09-23
Deploy catalogsource resource in the custom namespace

samira-barouti:
This PR modifies the DeployableByOLM check to deploy the catalogsource CR
in the `installNamespace`. This ensures that the catalogsource will be deleted
when the namespace is removed.

### DIFF
--- a/certification/internal/policy/operator/default.go
+++ b/certification/internal/policy/operator/default.go
@@ -1,9 +1,6 @@
 package operator
 
 const (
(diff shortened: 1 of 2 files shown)


bcrochet (approved):
/lgtm

komish (approved):
@samira-barouti I do wonder if we should consider adding a randomization to the end of the install namespace, but that's not required for this PR. Just wanted to plant the idea to see if it's something we should consider

/lgtm

PR #281, opened 2021-10-01
Make scorecard run with --verbose

komish:
This PR just adds the OperatorSdkScorecardOptions key that enables passing of `--verbose` to operator-sdk. The plumbing was already there, but the value was never set. 

I considered adding this as a configurable, but I figured it made sense to just force verbosity in favor of adding another environment variable given that there are probably next-to-no cases where scorecard breaks in unexpected ways and we don't want to the verbosity the first time around.


tkrishtop (approved):

acornett21 (approved):

samira-barouti (approved):
/lgtm

PR #282, opened 2021-10-01
Add support for private registry

### LINKED ISSUE #245: DeployableByOLM fails in k8s use cases if images are private
## Bug Description
If the bundle image and/or index image are in a private registry, the DeployableByOLM check will fail to successfully execute.

Preflight version is bf7386394b9db59ca9020b0d5363ac75

## Steps to Reproduce:
I've only experienced this while running Preflight in k8s as a job, but it may be reproducible otherwise. The basic steps look something like this:

1) Upload an index image and a bundle image in a private repo
2) Run preflight in k8s as a Job, mounting secrets such as the image's registry creds and a kubeconfig to the proper places.
3) Wait for preflight to complete
4) Review logs, see DeployableByOLM listed as failed

## Expected Result
DeployableByOLM should pass this check.

## Actual Result
DeployableByOLM is returned as an errored check (not a failed check) which implies that something went wrong in the execution.

## Additional Context
With `PFLT_LOGLEVEL=trace`, there still isn't much context as to what happened. However, given that the index image is private, I'm thinking that the problem is that the CatalogSource that's created by this check cannot pull the index image.

In addition, the namespace that's used for this check (which, I believe, is introspected from the package name) likely doesn't have pull secrets necessary to try and run this operator, so I would expect the check would fail beacuse the CSV may never reach "Ready". 

The install namespace is dynamically spun up, which makes presenting credentials tough. In theory, presenting credentials to the openshift-marketplace namespace (where we create the catalogsource) might be easier, but still means we're frequently creating credentials in this namespace.

The pipelines don't really have this challenge because they have access to the bundle itself, and dynamically create an index and push it to a known location for testing.

samira-barouti:
This PR enables the preflight to work with the index images that are in private
registries. It introduces a new mandatory DOCKERCONFIG environment variable
which must point to the docker config file.

Fixes #245

### DIFF
--- a/certification/internal/policy/operator/deployable_by_olm.go
+++ b/certification/internal/policy/operator/deployable_by_olm.go
@@ -3,11 +3,12 @@ package operator
 import (
 	"context"
 	"fmt"
-	"strings"
+	"os"
 	"time"
 
 	operatorv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
 	"github.com/spf13/viper"
+	corev1 "k8s.io/api/core/v1"
 
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification"
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/artifacts"
@@ -75,10 +76,6 @@ func (p *DeployableByOlmCheck) operatorMetadata(bundleRef certification.ImageRef
 	}
 
 	catalogImage := viper.GetString(indexImageKey)
-	if len(catalogImage) == 0 {
-		log.Error(fmt.Sprintf("To set the key, export PFLT_%s or add %s:<value> to config.yaml in the current working directory", strings.ToUpper(indexImageKey), indexImageKey))
-		return nil, errors.ErrIndexImageUndefined
-	}
 
 	channel, err := annotation(annotations, channelKey)
 	if err != nil {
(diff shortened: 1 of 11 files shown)


certification/internal/policy/operator/deployable_by_olm.go
```diff
@@ -75,10 +76,6 @@ func (p *DeployableByOlmCheck) operatorMetadata(bundleRef certification.ImageRef
 	}
 
 	catalogImage := viper.GetString(indexImageKey)
-	if len(catalogImage) == 0 {
```
acornett21, line 78: Is this validation no longer needed? Just want to make sure it wasn't removed on accident, as I don't see any validation for the new secrete code.

samira-barouti: I moved this to check_operator to ensure the check is failing fast.

bcrochet (commented):
What happens if there is no need for the secret? There are certainly going to be instances where this is not necessary. It seems to force it no matter what.

certification/internal/policy/operator/deployable_by_olm.go
acornett21: That makes sense, thanks for clarifying!

samira-barouti:
> What happens if there is no need for the secret? There are certainly going to be instances where this is not necessary. It seems to force it no matter what.

Just ran the preflight with a publicly-available index image and it is working fine. 
I also test it with multiple auth keys (one per repo) in the config.json and that is working well.

bcrochet (approved):
/lgtm

PR #285, opened 2021-10-05
Make docker config file optional

### LINKED ISSUE #284: `DOCKERCONFIG` should be `PFLT_DOCKERCONFIG` and should not be required
## Bug Description
PR #282 resolves an issue with private registry access when running the DeployableByOLM check by pushing an explicitly provided dockerconfigjson file to the cluster so that the CatalogSource can pull from a private registry.

With that said, it is **now** required and it should not be. If a user does _not_ set this environment variable, preflight currently exits with an error. 

It's plausible that the index image used by the CatalogSource will be publicly available, and so we should not have to create any secret, or pass a secret reference to a CatalogSource resource in order to run the check.

ef91135080a4a4a8d927b7d78b797d269be837f9

## Steps to Reproduce
Attempt to run `preflight check operator <bundle>` without DOCKERCONFIG set in the environment.

## Current result:

```
$ ./preflight check operator quay.io/opdev/simple-demo-operator-bundle:latest
time="2021-10-05T11:58:01-05:00" level=info msg="config file not found, proceeding without it"
time="2021-10-05T11:58:01-05:00" level=info msg="certification library version 0.0.0 <commit: ef91135080a4a4a8d927b7d78b797d269be837f9>"
Error: no environment variable DOCKERCONFIG could be found

The checks that will be executed are the following:
- DeployableByOLM
- ValidateOperatorBundle
- OperatorPackageNameIsUnique
- ScorecardBasicSpecCheck
- ScorecardOlmSuiteCheck

Usage:
  preflight check operator <url to Operator bundle image> [flags]

Flags:
  -h, --help   help for operator

time="2021-10-05T11:58:01-05:00" level=fatal msg="no environment variable DOCKERCONFIG could be found"
```

## Expected Result
Preflight to execute for a publicly available image without having to set the variable.

samira-barouti:
This PR makes the docker config file optional. It also leverages
viper to fetch the docker config file path.

Fixes #284

### DIFF
--- a/certification/errors/errors.go
+++ b/certification/errors/errors.go
@@ -38,4 +38,3 @@ var ErrRemoveOSPIDFromProjectID = errors.New("please remove leading character os
 var ErrRemoveSpecialCharFromProjectID = errors.New("please remove all special characters from project id")
 var ErrPullRequestURL = errors.New("please enter a valid url: including scheme, host, and path to pull request")
 var ErrIndexImageUndefined = errors.New("no environment variable PFLT_INDEXIMAGE could be found")
-var ErrNoDockerconfig = errors.New("no environment variable DOCKERCONFIG could be found")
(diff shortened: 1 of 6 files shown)


komish (approved):
/lgtm

bcrochet (approved):
/lgtm

PR #286, opened 2021-10-05
Centralize container images used at runtime into a single place, and provide access functions and a relevant subcommand.

### LINKED ISSUE #276: [Disconnected env] Have --config flag to be exposed by preflight to allow custom config for scorecard
Preflight automatically generates operator-sdk scorecard config `scorecard-test-config.yaml` which contains images like `quay.io/operator-framework/scorecard-test:master`.
I work from disconnected environment where I can't pull external-registry images, I always have to pull from my local registry. 

I'd like to have a way to pass my custom config into scorecard. Or, at least to pass my local-registry scorecard images. 
[Here is how that could be done when using scorecard directly](https://github.com/redhat-cip/ansible-role-dci-cvp/blob/e91ba3f512dbfbfabafa82720632d75f6b43a76e/tasks/mirror_bundle.yml#L48)

To have this feature is mandatory to run scorecard inside preflight in disconnected environments.

komish:
Fixes #276

Disconnected environments need to know what images preflight might use to execute checks ahead of actual execution. Ideally, those in disconnected environments would want to provide their own scorecard configs (for example) so that they could rewrite the images, but we don't necessarily want that to change dynamically within preflight.

The compromise was that we'd expose exactly what assets we use at runtime (if any). Currently, I see our scorecard image is in scope, since we write that configuration. This PR just adds in a centralized management approach at the `runtime` package level for container images we use. The output looks something like this (today). 

```
$ ./preflight --version
preflight version 0.0.0 <commit: c53ff6816ff7c376ab682ad0dba3df4ccd693d32>

$ ./preflight runtime-assets
{
    "images": [
        "quay.io/operator-framework/scorecard-test@sha256:d655333b0246f75ac9e5f6e67a2c04c506ae77b0c8b0c5eb70e6ddc8c2123e55"
    ]
}
```

There are hooks into the runtime package that expose the images we use, and those hooks are used in our DeployableByOLM check to make sure we don't drift.

### DIFF
--- a/certification/internal/engine/operatorsdk.go
+++ b/certification/internal/engine/operatorsdk.go
@@ -11,6 +11,7 @@ import (
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/artifacts"
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/errors"
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/internal/cli"
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/runtime"
 	log "github.com/sirupsen/logrus"
 )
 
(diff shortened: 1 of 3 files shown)


acornett21 (approved):

samira-barouti (approved):
/lgtm

PR #287, opened 2021-10-06
Simplify the OpenShift client abstraction

bcrochet:
The OpenShift/k8s client abstraction seemed to be a bit complex. I
don't think we were seeing any benefit from such a layered abstraction.
This removes a bunch of the layers, and just uses the controller-runtime
client interface as the abstraction instead.


samira-barouti:
> > Thank you Brad for the code simplification. I need to go into the details for a comprehensive code review. I, however, have a comment regarding the overall design:
> > I like the idea of simplifying the code for the k8s APIs interaction. But, I feel that deployableByOLM is now so overloaded with k8s-specific methods. This seems more like a functional programming style, rather than OOP which leverages the abstraction to have a cleaner interface. With that being said, I prefer to keep the notion of OpenshiftEngine, but simplify the implementation to have a single client for all resources.
> 
> Thank you for the feedback. I'm happy to iterate on it a bit. I really wanted to get something in front of everyone so that we could discuss the direction.
> 
> One thing I may do is separate out the watch dedupe into its own PR. That work can be made separate I believe.
> 
> I'm not completely sold on keeping OpenshiftEngine. It feels like a forced abstraction. The more idiomatic approach for golang is that the consumer creates the interface, not the producer.
> 
> As for functional vs. OOP. After my years of using Python, I've gotten away from the Java mentality of "everything's an object", and that's definitely coming out here. I personally am OK with that. But happy to have further discussion on the topic.

I am not in favor of "Everything is an object" mentality as well; but I believe "separation of concerns" and "encapsulation" are still good practices when applicable. The implementation for check should be focused on the check itself; in this PR, there are 15+ methods in the check which implements the k8s client, for which the implementation details should be hidden from the check (as a k8s APIs consumer).

bcrochet:
I'm going to close this for now. Making progress in moving to the typed client.

PR #296, opened 2021-10-18
Update operator-sdk to 1.13.1

### LINKED ISSUE #294: Operator SDK needs to be updated to 1.13.1
To be in sync with CVP, the operator sdk version should be 1.13.1.

bcrochet:
In order to support new tests geared towards deprecated APIs and
OpenShift 4.9, bump operator-sdk to 1.13.1.

Fixes #294

### DIFF
--- a/certification/runtime/assets.go
+++ b/certification/runtime/assets.go
@@ -11,8 +11,8 @@ var (
 	// to enable disconnected environments.
 	images = map[string]string{
 		// operator policy, operator-sdk scorecard
-		// quay.io/operator-framework/scorecard-test:v1.12.0
-		"scorecard": "quay.io/operator-framework/scorecard-test@sha256:d655333b0246f75ac9e5f6e67a2c04c506ae77b0c8b0c5eb70e6ddc8c2123e55",
+		// quay.io/operator-framework/scorecard-test:v1.13.1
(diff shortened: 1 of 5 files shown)


certification/runtime/assets.go
```diff
@@ -11,8 +11,8 @@ var (
 	// to enable disconnected environments.
 	images = map[string]string{
 		// operator policy, operator-sdk scorecard
-		// quay.io/operator-framework/scorecard-test:v1.12.0
-		"scorecard": "quay.io/operator-framework/scorecard-test@sha256:d655333b0246f75ac9e5f6e67a2c04c506ae77b0c8b0c5eb70e6ddc8c2123e55",
+		// quay.io/operator-framework/scorecard-test:v1.13.1
+		"scorecard": "quay.io/operator-framework/scorecard-test@sha256:417d5157a782f72ba83cc613a0a3997cea088fe025c188121f782e93b38bf04e",
```
acornett21, line 15: Nonblocking: FYI, when scorecard does a build of a new version, they are overwriting older tags with new `shas` so for us this is problematic. We have an ask to stop this overwrite, but we might have to solve this issue in preflight so that connected and disconnected environments can continue to work.

PR #297, opened 2021-10-19
Add check for image sources to DeployableByOLM

bcrochet:
In order to check the image sources, we look for the diff in images in pods and
image streams, before and after the operator is deployed. This diff is then checked
for the sources to ensure they are from approved sources.

### DIFF
--- a/certification/internal/engine/openshift.go
+++ b/certification/internal/engine/openshift.go
@@ -6,11 +6,14 @@ import (
 
 	log "github.com/sirupsen/logrus"
 
+	imagestreamv1 "github.com/openshift/api/image/v1"
 	operatorv1 "github.com/operator-framework/api/pkg/operators/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/kubernetes"
 	ctrl "sigs.k8s.io/controller-runtime"
+	crclient "sigs.k8s.io/controller-runtime/pkg/client"
 
 	operatorv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/internal/cli"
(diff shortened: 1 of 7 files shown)


certification/internal/engine/openshift.go
```diff
@@ -306,3 +309,47 @@ func (oe *openshiftEngine) GetCSV(name string, opts cli.OpenshiftOptions) (*oper
...
+		for _, container := range pod.Spec.Containers {
+			imageList[container.Image] = struct{}{}
+		}
+	}
+
+	scheme := runtime.NewScheme()
+	imagestreamv1.AddToScheme(scheme)
+	isClient, err := crclient.New(kubeconfig, crclient.Options{Scheme: scheme})
+	if err != nil {
+		log.Error("could nod create isClient: ", err)
```
samira-barouti, line 338: typo: nod -> not

certification/internal/policy/operator/deployable_by_olm.go
```diff
@@ -63,7 +80,41 @@ func (p *DeployableByOlmCheck) Validate(bundleRef certification.ImageReference)
...
+
+	afterOperatorImages, err := p.getImages()
+	if err != nil {
+		return false, err
+	}
+
+	operatorImages := diffImageList(beforeOperatorImages, afterOperatorImages)
+	p.validImages = checkImageSource(operatorImages)
+
+	return p.csvReady || p.validImages, nil
```
samira-barouti, line 96: I feel this has to be && and not ||. Am I missing sth?

certification/internal/policy/operator/deployable_by_olm.go
```diff
@@ -63,7 +80,42 @@ func (p *DeployableByOlmCheck) Validate(bundleRef certification.ImageReference)
 		return false, err
 	}
 
-	return p.isCSVReady(installedCSV, *operatorData)
+	p.csvReady, err = p.isCSVReady(installedCSV, *operatorData)
+	if err != nil {
```
samira-barouti, line 84: I think even if the isCSVReady returns an error, we should still run the imageRegistryCheck. Therefore, returning on the line 85 will skip the imageRegistryCheck all together. This means that the return statement on line 96 should be a bit complicated, considering all possible execution outcomes. I do not have an easy solution for this, but I'd be glad to discuss further.

bcrochet: Definitely an issue when combining checks like this. In this particular case, I (and I could be mistaken here) see the registry check as dependent on whether the first part succeeds. If it never becomes ready, then checking for the image registry is probably a waste of time, as it may have never even pulled additional images. If you feel strongly about this one, let me know. I'm inclined to leave it as is, for now.

PR #303, opened 2021-10-24
fix: should not fail if pkgname is already owned by the requestor

tonytcampbell:
The package name needs to be unique in the catalog but if the person requesting certification already owns the package name, this check should pass.  In other words, if the user already has a lock on the package name they should be allowed to proceed with certification.


umeshkumhar:
@tonytcampbell This will also fail because this is validated with cert_project_id which is the project ID pulled from URL (as mentioned in the guide ) and not OSPID ID, because it will fail on upload log artifacts afterwards because endpoint formation is done via cert_project_id var.

PR #321, opened 2021-11-01
Adding extra trace logging to DeployableByOLM

tonytcampbell:
Adding extra trace logging to the DeploybleByOLM check.  When this check fails it can be tricky to diagnose the issue.  This extra logging provides a little more context and some extra clues.

### DIFF
--- a/certification/internal/policy/operator/deployable_by_olm.go
+++ b/certification/internal/policy/operator/deployable_by_olm.go
@@ -294,6 +294,7 @@ func (p *DeployableByOlmCheck) installedCSV(ctx context.Context, operatorData Op
 		for {
 			log.Debug("Waiting for Subscription.status.installedCSV to become ready...")
 			subs, _ := p.OpenshiftEngine.GetSubscription(ctx, operatorData.App, operatorData.InstallNamespace)
+			log.Trace(subs.Status)
 			installedCSV := subs.Status.InstalledCSV
 			// if the installedCSV field is present, stop the querying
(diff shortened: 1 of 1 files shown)


samira-barouti (approved):

certification/internal/policy/operator/deployable_by_olm.go
```diff
@@ -294,6 +294,7 @@ func (p *DeployableByOlmCheck) installedCSV(ctx context.Context, operatorData Op
 		for {
 			log.Debug("Waiting for Subscription.status.installedCSV to become ready...")
 			subs, _ := p.OpenshiftEngine.GetSubscription(ctx, operatorData.App, operatorData.InstallNamespace)
```
itroyano, line 296: Is there a case where subs can be nil here ?

samira-barouti: The check returns if API server fails to create the subscription. So, I would say the subscription exists when we get here.

certification/internal/policy/operator/deployable_by_olm.go
```diff
@@ -294,6 +294,7 @@ func (p *DeployableByOlmCheck) installedCSV(ctx context.Context, operatorData Op
...
+			log.Trace(subs.Status)
```
samira-barouti, line 297: I believe adding a debug message will be helpful.

bcrochet (approved):
/lgtm

PR #322, opened 2021-11-01
Hiding test_on from results.json until pipeline is ready

tonytcampbell:
We need to do a coordinated release with the Operator Pipeline.  As it stands this data will throw an error until updates are made to Pyxis and to the Pipeline.   Simply commented it out until we are ready to release.


samira-barouti (approved):
/lgtm

PR #325, opened 2021-11-03
Enable preflight to support all Install modes

### LINKED ISSUE #293: preflight fails DeployableByOLM check for operator with installMode of AllNamespaces
## Bug Description
When running `preflight check operator` against an operator that only has an installMode of `AllNamespaces` the DeployableByOLM check fails even though, this operator can be installed by OLM via the embedded OperatorHub.

`preflight version 0.0.0 <commit: b11314a6b9603ba53d88826888c9856a738161b8>`

## Steps to Reproduce:
(How can we reproduce this?)
1) Run: `export PFLT_INDEXIMAGE="quay.io/acornett/elasticache-index-no-ad:v0.0.6"`
2) Run `preflight check operator quay.io/acornett/elasticache-controller-bundle-no-ad:v0.0.6`

## Expected Result
I expected this operator to be able to pass all checks, especially deployable by olm, since this operator can be installed via `operator-sdk` as well as via a the embedded OperatorHub. And that preflight can support any/all combinations of `installMode`

## Actual Result
The operator failed to install with the below error(s)
- `OwnNamespace InstallModeType not supported, cannot configure to watch own namespace`
- `UnsupportedOperatorGroup`

If you tail the logs of the olm operator, you'll get the below error
- For example `oc logs -f -n openshift-operator-lifecycle-manager olm-operator-`
- ```time="2021-10-15T15:42:20Z" level=info msg="installmodeset does not support operatorgroups namespace selection" csv=ack-elasticache-controller.v0.0.6 id=yMc9n namespace=ack-system phase=Failed reason="OwnNamespace InstallModeType not supported, cannot configure to watch own namespace"```

samira-barouti:
Fixes #293

This PR enables the preflight to support all install modes (namely OwnNamespace, SingleNamespace, MultiNamespace, AllNamespaces). In a nutshell, this PR:
* refactored watch method to use one channel to be used for errors as well as data. This enables the main goroutine to correctly identify if the check has passed or not
* creates a new namespace which will be used by SingleNamespace and MultiNamespace.
* Reads the operator installModes from csv file, stored on the bundle image
* Selects installModes from a prioritized list. The order is from the highest priority to the lowest is OwnNamespace, SingleNamespace, MultiNamespace, AllNamespaces
* Generates the TargetNamespace of the OperatorGroup based on the selected installmode
* Enquires the csv status from one of the namespace in TargetNamespaces in the case of OwnNamespace, SingleNamespace and MultiNamespace
* In the case of AllNamespaces, enquire the csv status from default, openshift-marketplace and operatorData.TargetNamespace

### DIFF
--- a/certification/internal/policy/operator/deployable_by_olm.go
+++ b/certification/internal/policy/operator/deployable_by_olm.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"strings"
+	"sync"
 	"time"
 
 	operatorv1 "github.com/operator-framework/api/pkg/operators/v1"
@@ -18,11 +19,11 @@ import (
 
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification"
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/artifacts"
-	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/errors"
+	pflterr "github.com/redhat-openshift-ecosystem/openshift-preflight/certification/errors"
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/internal/cli"
 
 	log "github.com/sirupsen/logrus"
-	kubeErr "k8s.io/apimachinery/pkg/api/errors"
+	k8serr "k8s.io/apimachinery/pkg/api/errors"
 )
 
 type OperatorData struct {
@@ -31,6 +32,10 @@ type OperatorData struct {
 	PackageName      string
 	App              string
 	InstallNamespace string
+	TargetNamespace  string
+	InstallModes     map[string]bool
+	CsvNamespaces    []string
+	InstalledCsv     string
 }
 
 type DeployableByOlmCheck struct {
@@ -38,19 +43,6 @@ type DeployableByOlmCheck struct {
 	csvReady        bool
 	validImages     bool
 }
-
-var (
-	subscriptionTimeout time.Duration = 180 * time.Second
-	csvTimeout          time.Duration = 90 * time.Second
-	approvedRegistries                = map[string]struct{}{
-		"registry.connect.dev.redhat.com":   {},
-		"registry.connect.qa.redhat.com":    {},
-		"registry.connect.stage.redhat.com": {},
-		"registry.connect.redhat.com":       {},
-		"registry.redhat.io":                {},
-		"registry.access.redhat.com":        {},
-	}
-)
 
 func NewDeployableByOlmCheck(openshiftEngine *cli.OpenshiftEngine) *DeployableByOlmCheck {
 	return &DeployableByOlmCheck{
@@ -74,7 +66,7 @@ func (p *DeployableByOlmCheck) Validate(bundleRef certification.ImageReference) 
 	}
 
 	// create k8s custom resources for the operator deployment
-	err = p.setUp(ctx, *operatorData)
+	err = p.setUp(ctx, operatorData)
 	defer p.cleanUp(ctx, *operatorData)
 
 	if err != nil {
@@ -85,8 +77,10 @@ func (p *DeployableByOlmCheck) Validate(bundleRef certification.ImageReference) 
 	if err != nil {
 		return false, err
 	}
+	operatorData.InstalledCsv = installedCSV
+	log.Trace("the installed CSV is ", operatorData.InstalledCsv)
 
-	p.csvReady, err = p.isCSVReady(ctx, installedCSV, *operatorData)
+	p.csvReady, err = p.isCSVReady(ctx, *operatorData)
 	if err != nil {
 		return false, err
 	}
@@ -137,13 +131,19 @@ func (p *DeployableByOlmCheck) operatorMetadata(bundleRef certification.ImageRef
 
 	channel, err := annotation(annotations, channelKey)
 	if err != nil {
-		log.Error("unable to extract channel name from ClusterServicVersion", err)
+		log.Error("unable to extract channel name from the bundle: ", err)
 		return nil, err
 	}
 
 	packageName, err := annotation(annotations, packageKey)
 	if err != nil {
-		log.Error("unable to extract package name from ClusterServicVersion", err)
+		log.Error("unable to extract package name from the bundle: ", err)
+		return nil, err
+	}
+
+	installedModes, err := getSupportedInstalledModes(bundleRef.ImageFSPath)
+	if err != nil {
+		log.Error("unable to extract operator install modes from ClusterServicVersion: ", err)
 		return nil, err
 	}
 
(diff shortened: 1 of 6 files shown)


certification/internal/policy/operator/deployable_by_olm.go
```diff
@@ -31,6 +31,10 @@ type OperatorData struct {
 	PackageName      string
 	App              string
 	InstallNamespace string
+	TargetNamespace  string
+	InstallModes     map[string]bool
+	CsvNamespaces    []string
+	InstalledCsv     string
```
acornett21, line 37: Nit: `CSV` in var names for consistency.

certification/internal/policy/operator/deployable_by_olm.go
```diff
@@ -234,9 +268,15 @@ func (p *DeployableByOlmCheck) grantRegistryPermissionToServiceAccount(ctx conte
 	return nil
 }
 
-func (p *DeployableByOlmCheck) isCSVReady(ctx context.Context, installedCSV string, operatorData OperatorData) (bool, error) {
+func (p *DeployableByOlmCheck) isCSVReady(ctx context.Context, operatorData OperatorData) (bool, error) {
 
-	log.Trace(fmt.Sprintf("Looking for csv %s in namespace %s", installedCSV, operatorData.InstallNamespace))
+	var CsvNamespace string
```
acornett21, line 273: Nit: CSV in var names for consistency.

certification/internal/policy/operator/deployable_by_olm.go
```diff
@@ -153,6 +151,8 @@ func (p *DeployableByOlmCheck) operatorMetadata(bundleRef certification.ImageRef
 		PackageName:      packageName,
 		App:              packageName,
 		InstallNamespace: packageName,
```
acornett21, line 153: Since we are now parsing the `CSV` does it make sense to use the value for `InstalledNamespace` from whatever is listed in `operatorframework.io/suggested-namespace` if this key is present, or is that overkill? @tonytcampbell Any thoughts on this?

tonytcampbell: I think this may be something to consider for a future enhancement. We can create a GH issue to track that as well.

acornett21: Thanks for the input, will create an issue for this and resolve this comment.

certification/internal/policy/operator/deployable_by_olm.go
```diff
@@ -278,29 +317,68 @@ func csvStatusSucceeded(ctx context.Context, engine cli.OpenshiftEngine, name, n
...
-	csvReadyDone := make(chan string, 1)
-	defer close(csvReadyDone)
+	doneChannel := make(chan string)
+	errChannel := make(chan error)
 
-	contextTimeout := make(chan error, 1)
-	defer close(contextTimeout)
+	var wg sync.WaitGroup
+	go func() {
+		wg.Wait()
```
acornett21, line 338: Does this wait group ever get finished since it's in a different thread than the thread that is adding to the wait group? I'd expect to see this `wg.Wait()` at line 347. This might work, but I've never seen it this way, so I am unsure.

certification/internal/policy/operator/deployable_by_olm.go
```diff
@@ -309,31 +387,57 @@ func subscriptionCsvIsInstalled(ctx context.Context, engine cli.OpenshiftEngine,
...
 	return "", false, nil
 }
 
 func (p *DeployableByOlmCheck) installedCSV(ctx context.Context, operatorData OperatorData) (string, error) {
+	doneChannel := make(chan string)
+	errChannel := make(chan error)
+
+	var wg sync.WaitGroup
+	go func() {
+		wg.Wait()
```
acornett21, line 405: Same comment as above for this.

certification/internal/policy/operator/deployable_by_olm.go
```diff
@@ -219,6 +228,33 @@ func (p *DeployableByOlmCheck) setUp(ctx context.Context, operatorData OperatorD
...
+		targetNamespaces = []string{operatorData.InstallNamespace}
+	case string(operatorv1alpha1.InstallModeTypeSingleNamespace):
+		targetNamespaces = []string{operatorData.TargetNamespace}
+	case string(operatorv1alpha1.InstallModeTypeMultiNamespace):
+		targetNamespaces = []string{operatorData.TargetNamespace, operatorData.InstallNamespace}
+	case string(operatorv1alpha1.InstallModeTypeAllNamespaces):
+		targetNamespaces = []string{}
+
+	}
+	log.Debug(fmt.Sprintf("The operator group's targetNamespaces is %s", targetNamespaces))
```
acornett21, line 253: Should this be `%v` since it is a slice? Or some form to get all the values from the slice?

certification/internal/policy/operator/deployable_by_olm.go
samira-barouti: yes it does. For the check, we need to read the messages from channels and the goroutine should not be blocked when writing to channels. I found this article useful to explain why wg.wait() has to be in another goroutine when you we want to read messages from multiple goroutine over a shared channel: https://dev.to/sophiedebenedetto/synchronizing-go-routines-with-channels-and-waitgroups-3ke2

certification/internal/policy/operator/deployable_by_olm.go
samira-barouti: %s works well with string slice. Found this in the fmt doc:
```%s	the uninterpreted bytes of the string or slice
```

acornett21: That doesn't seem to be the case based on the log you shared. I might be missing something though. I would expect the below to have the values of the slice.
```
 time="2021-11-05T15:12:42Z" level=debug msg="The operator group's targetNamespaces is []"
 ```

certification/internal/policy/operator/deployable_by_olm.go
acornett21: Awesome! Never seen it coded this way so I was unsure. Thanks for the link!

certification/internal/policy/operator/deployable_by_olm.go
samira-barouti: In the case of installMode: AllNamespaces, the target namespace of the operator group is empty. That's why you see `[]` in the output. Here is a sample output if the install mode is OwnNamespace:
```time="2021-11-05T16:54:28Z" level=debug msg="The operator install mode is OwnNamespace"
time="2021-11-05T16:54:28Z" level=debug msg="The operator group's targetNamespaces is [gitlab-runner-operator]"```

certification/internal/policy/operator/deployable_by_olm.go
acornett21: I was confusing `wg.Wait()` and `wg.Done()` I see the `wg.Done()` as a defer in the `watch()` function.

acornett21 (approved):

tonytcampbell (approved):
/lgtm

PR #333, opened 2021-11-09
Add multi-arch binary build capability

rpsene:
This commit adds a new target for
	building multi-arch preflight
	binaries:

	make build_multi_arch

Signed-off-by: Rafael Peria de Sene <rpsene@br.ibm.com>

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -27,8 +27,8 @@ jobs:
     - name: Format
       run: make fmt
 
-    - name: Build
-      run: make build
+    - name: Build Multi-arch
+      run: make build-multi-arch
 
     - name: Test
       run: make test
--- a/Makefile
+++ b/Makefile
@@ -1,13 +1,25 @@
 .DEFAULT_GOAL:=help
 
+BINARY=preflight
 IMAGE_BUILDER?=podman
 IMAGE_REPO?=quay.io/opdev
 VERSION=$(shell git rev-parse HEAD)
 RELEASE_TAG ?= "0.0.0"
 
+PLATFORMS=linux
+ARCHITECTURES=amd64 arm64 ppc64le s390x
+
 .PHONY: build
 build:
-	go build -o preflight -ldflags "-X github.com/redhat-openshift-ecosystem/openshift-preflight/version.commit=$(VERSION) -X github.com/redhat-openshift-ecosystem/openshift-preflight/version.version=$(RELEASE_TAG)" main.go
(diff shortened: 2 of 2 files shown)


rpsene:
@samira-barouti @bcrochet could you take a look on this one?

Makefile
```diff
@@ -37,4 +48,9 @@ vet:
 
 .PHONY: test-e2e
 test-e2e:
-	go test -v ./test/e2e/ 
+	go test -v ./test/e2e/
+
+.PHONY: clean
+clean:
+	go clean
+	rm -f ./preflight*
```
bcrochet, line 56: This seems risky.... Can we iterate over the them like the build_multi_arch target does?

Makefile
```diff
@@ -1,13 +1,24 @@
...
+PLATFORMS=linux
+ARCHITECTURES=amd64 arm64 ppc64le s390x
+
 .PHONY: build
 build:
-	go build -o preflight -ldflags "-X github.com/redhat-openshift-ecosystem/openshift-preflight/version.commit=$(VERSION) -X github.com/redhat-openshift-ecosystem/openshift-preflight/version.version=$(RELEASE_TAG)" main.go
+	go build -o $(BINARY) -ldflags "-X github.com/redhat-openshift-ecosystem/openshift-preflight/version.commit=$(VERSION) -X github.com/redhat-openshift-ecosystem/openshift-preflight/version.version=$(RELEASE_TAG)" main.go
+
+.PHONY: build_multi_arch
+build_multi_arch:
```
bcrochet, line 17: This should use '-' instead of '_' to keep with other targets.

.github/workflows/go.yml
```diff
@@ -30,5 +30,8 @@ jobs:
     - name: Build
       run: make build
 
+    - name: Build Multi-arch
```
bcrochet, line 33: Do we need both this and the prior action? Since this will cover the linux/amd64 target (the default) it should be fine to have just this one. So I'd say replace the previous action with this one.

bcrochet (changes requested):
A couple of small changes, please.

bcrochet:
Why all the '@'? That is doing more in this patch than just adding multiarch support. Can you limit the changes to just what is necessary for adding multiarch?

rpsene:
@bcrochet the @ is to avoid printing the command when running make. I removed it from the parts I didn't touch for multi-arch.

bcrochet:
> @bcrochet the @ is to avoid printing the command when running make. I removed it from the parts I didn't touch for multi-arch.

I am curious as to why that would be desirable? Honest question. When running make, I personally would prefer to see the progress and what is being run, especially if there are variables involved.

bcrochet (approved):
/lgtm

jomkz (approved):
/lgtm

Issue #336, opened 2021-11-12
Automatically test Preflight policy execution on pull request


komish:
## Is your feature request related to a problem? Please describe.
Yes, currently there is no automated testing against new PRs that tests Preflight's functionality with the bugfix against a known good asset and a known supported OCP cluster version.

## Describe the solution you'd like.
Integrate with https://prow.ci.openshift.org's , consuming the ci-operator to claim a cluster of a version we support, and run against existing test assets.

## Describe alternatives you've considered.
- Standing up our own clusters (would require much more admimnistrator than consuming the mentioned platform)
- Standing up a single cluster per version and using task runners to target them in GH actions (less clean, higher likelihood of conflict)

## Additional context.
The work summary looks something like this:
- Enable openshift CI robots on our repositories so that the robots can perform automation on our behalf based on chat-ops commands (note: might conflict with our existing robots).
- Optionally enable merge robots to auto merge (we might not want this)
- Configure this repository in https://github.com/openshift/release to enable Prow plugins of interest (approve, lgtm, merge, etc)
- Define our Presubmit jobs, which should build binaries including the commits from the PR in scope
- Build and/or reuse any existing steps from the step registry to use the ci-operator and run our task. We could also embed our tests as scripts in this repository, and run those as our "steps" using this platform (instead of writing the entirety of the script in the test platform).
- Evaluate results to make sure everything passed, and otherwise just ensure preflight doesn't panic or fail with the change in question (note: these results will be public, but there should be little concern of credential leak because our test assets exist in public registries and don't require configuration or credentials themselves).

## Decisions/Requires Additional Input
|Pending Decision|
|---|
|**(Question)** Should this testing occur on every PR? What PRs should not be subject to this workflow (e.g  Docs PRs?) Can that level of filtering be automated?|
|**(Decision)** TBD|

|Pending Decision|
|---|
|**(Question)** Should we have different automation for releases? Tags? How does this automation differ?|
|**(Decision)** TBD|

|Pending Decision|
|---|
|**(Question)** Do we want to enable automatic merge using Tide|
|**(Decision)** TBD|

vikasmulaje:
Hi @komish,
I am working on same scenario made some progress on build jenkins pipeline job and Preflight automation test.
Lets collaborate on this

komish:
[...]

PR #340, opened 2021-11-16
Add operator testing script to e2e suite

komish:
Related: #336 

This PR adds a script that runs the `check operator` target against known-good assets for use in e2e testing scenarios.

I may pull this at a later date, but for the moment I'm thinking it makes sense to just run something like `make test-e2e` in the DPTP environment as a presubmit job. In theory, we'll be able to build the binary on each PR, then run this script - and if everything passes, it should be considered a successful test.

Down the road, I'll have to add similar for the `check container` target, and any other targets we might want to add. I'm just adding this target for now as it's the most complex (requiring a cluster, index image, etc).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,7 +49,7 @@ vet:
 
 .PHONY: test-e2e
 test-e2e:
-	go test -v ./test/e2e/
+	./test/e2e/operator-test.sh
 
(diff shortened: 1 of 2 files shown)


bcrochet (approved):
/lgtm

PR #341, opened 2021-11-16
Modify runtime assets to dynamically pull the digest

bcrochet:
Instead of hard-coding the digest, this patch now pulls the digest dynamically
at runtime, when 'preflight runtime-assets' is run. The output would then be
used to mirror to a disconnected registry. Also, when preflight is run in a
disconnected cluster, this digest can be passed via PFLT_SCORECARD_IMAGE.

This predicates that the preflight runtime-assets be run on a connected host.
But because the image can be passed by envvar, preflight check operator can
be run on a disconnected cluster without having the sha hard-coded.

### DIFF
--- a/certification/runtime/assets.go
+++ b/certification/runtime/assets.go
@@ -1,18 +1,23 @@
 package runtime
 
+import (
+	"fmt"
+	"strings"
+
+	"github.com/google/go-containerregistry/pkg/crane"
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/viper"
+)
+
 var (
 
 	// images maps the images use by preflight with their purpose.
 	//
 	// these should have accessor functions made available if they are
 	// to be used outside of this package.
-	//
-	// These images should also all be referenced using digests over tags
-	// to enable disconnected environments.
 	images = map[string]string{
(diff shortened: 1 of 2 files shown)


certification/runtime/assets.go
```diff
@@ -23,7 +28,14 @@ func imageList() []string {
 
 	i := 0
 	for _, image := range images {
```
acornett21, line 30: why are we ignoring the index, but creating an index outside of the loop an incrementing? can't we have i declared here and not increment? I might be missing something though.

acornett21:
Would there be a PR that follows this to update the documentation for the new env variable?

certification/runtime/assets.go
bcrochet: We're ignoring the key, not the index. images is a map.

bcrochet:
> Would there be a PR that follows this to update the documentation for the new env variable?

I'll work on adding it to this one.

certification/runtime/assets.go
acornett21: That it is, was thinking slice/array for some reason.

acornett21 (approved):

jomkz (approved):

PR #343, opened 2021-11-17
Exit script execution on error

komish:
I overlooked certain cases where attempting to execute a command would fail but we wouldn't catch it (such as `command not found`). Setting `set -e` to resolve this for the time being.


bcrochet (approved):
/lgtm

jomkz (approved):

PR #345, opened 2021-11-22
[WIP] E2E Preflight go test

vikasmulaje:
Initial Commit for E2E preflight testing


vikasmulaje:
@komish Hi,
Created PR for e2e tests as per discussion.
help to review.
Thanks

komish:
@vikasmulaje thanks for contributing, and your patience while I got the underlying testing infrastructure integrated.

I'm understanding that this intends to replace the [script](https://github.com/redhat-openshift-ecosystem/openshift-preflight/blob/main/test/e2e/operator-test.sh) based approach that we have. Is the goal to follow this PR with some Ginkgo tests that consume these functions?

Our "final" PR for e2e testing is just about to finish merging (https://github.com/openshift/release/pull/25047) using our existing script. Switching out to some other test target is straightforward enough. 

The script-based approach is very incomplete, and only tests a single command (`preflight check operator`). I just needed something to run for the integration.

To summarize, this draft PR is really proposing an alternate solution to plug into DPTP.

@samira-barouti @bcrochet, this PR really just needs a decision as to whether we finish implementing the rest of the tests in pure-bash, using this library + Ginkgo, or some other solution. I think pure-bash is probably easier to get off the ground _quickly_, but probably leaves things incomplete.

I'll be OOO starting next week, so I'll leave it to the other Preflight maintainers to decide how to proceed and move forward.

Thanks again!

vikasmulaje:
Hi @komish,

I'm understanding that this intends to replace the script based approach that we have.--> Correct, we can run these actual tests instead of the shell script.
 Is the goal to follow this PR with some Ginkgo tests that consume these functions?--> right now this is just using "testing" library from go. I can add these into Ginkgo.

The script-based approach is very incomplete, and only tests a single command (preflight check operator). I just needed something to run for the integration.--> in this draft I have aded few basic command-line validations with `preflight check operator` and `preflight check container` More validations will follow later.

Thanks

test/e2e/preflightCliTest_test.go
```diff
@@ -0,0 +1,128 @@
...
+	"fmt"
+	"io/ioutil"
+	"os"
+	"strings"
+	"testing"
+)
+
+func TestPreflightHelp(t *testing.T) {
+	got := preflightHelp()
+	want := "A utility that allows you to pre-test your bundles, operators, and container before submitting for Red Hat Certification.\n\nUsage:\n  preflight [command]\n\nAvailable Commands:\n  certify        Submits check results to Red Hat\n  check          Run checks for an operator or container\n  completion     generate the autocompletion script for the specified shell\n  help           Help about any command\n  runtime-assets Returns information about assets used at runtime.\n  support        Submits a support request\n\nFlags:\n  -h, --help      help for preflight\n  -v, --version   version for preflight\n\nUse \"preflight [command] --help\" for more information about a command.\n"
```
bcrochet, line 15: These are going to be extremely fragile. I don't see the utility of this type of test. The majority of the help text is going to be auto-generated by cobra anyway. In fact, a change was made recently that already would have made these fail.

test/e2e/preflightCliTest_test.go
```diff
@@ -0,0 +1,128 @@
...
+func TestContainerArtifcats(t *testing.T) {
+	_, err := os.OpenFile("artifacts/results.json", os.O_RDONLY, 0)
+	if errors.Is(err, os.ErrNotExist) {
+		t.Errorf("No artifacts generated for Container run")
+	}
+	os.RemoveAll("artifacts")
+}
+
+func TestPreflightOperatorRun(t *testing.T) {
+	preflightOperatorRun()
```
bcrochet, line 104: This is definitely going to fail as written. A lot more goes into calling check operator. A cluster is required for one thing.

bcrochet (changes requested):
Please squash commits to a single commit (with no merges). A couple of other comments as well.

test/e2e/preflightCliTest_test.go
vikasmulaje: Correct, this needs to be changed wrt changes in preflight help messages. 
I don't see the utility of this type of test--> This is just basic test if preflight is displaying correct help message

test/e2e/preflightCliTest_test.go
vikasmulaje: This test assumes Kubeconfig and index image env variables are set before running this.
Should I add check if kubeconfig ENV is set then only run otherwise skip?

test/e2e/preflightCliTest_test.go
vikasmulaje: Sure, I will remove the help tests

test/e2e/preflightCliTest.go
```diff
@@ -0,0 +1,106 @@
...
+	cmd := exec.Command("preflight")
+	output, err := cmd.Output()
+
+	if err != nil {
+		fmt.Println("error:", err)
+	} else {
+		//#fmt.Printf("output: %s \n\n\n%s", string(output), string(want))
+	}
+	return string(output)
+}
```
samira-barouti, line 19: IMO, for this you can take a look how we run shell commands and return an output that contains both stdout and stderr. You can then run some checks against stdout/stderr/error.

https://github.com/redhat-openshift-ecosystem/openshift-preflight/blob/da9e057934e43eefc765b0be3505b2b07630d23b/certification/internal/engine/podman.go#L48

test/e2e/preflightCliTest.go
```diff
@@ -0,0 +1,106 @@
...
+
+func preflightCheck() string {
+	cmd := exec.Command("preflight", "check")
+	output, err := cmd.Output()
+
+	if err != nil {
+		fmt.Println("error:", err)
+	}
+	return string(output)
```
samira-barouti, line 28: same here

test/e2e/preflightCliTest.go
```diff
@@ -0,0 +1,106 @@
...
+	if err != nil {
+		fmt.Println("error:", err)
+	} else {
+		//#fmt.Printf("output: %s \n\n\n%s", string(output), string(want))
+	}
+	return string(output)
+}
+
+func preflightCheck() string {
+	cmd := exec.Command("preflight", "check")
```
samira-barouti, line 22: you better receive the path to the preflight binary from the env var, and give the end user to be in any dir when running the e2e check.

test/e2e/preflightCliTest.go
```diff
@@ -0,0 +1,106 @@
...
+	if err != nil {
+		fmt.Println("error:", err)
+	} else {
+		fmt.Printf("output: %s \n\n\n", string(output))
+	}
+	return string(output)
+}
+
+func preflightSupport() string {
+	cmd := exec.Command("preflight", "support")
```
samira-barouti, line 56: could be removed. We no longer have any `support` subcommand.

test/e2e/preflightCliTest_test.go
```diff
@@ -0,0 +1,128 @@
...
+	if got != want {
+		t.Errorf("got %q \n\n\n\nwant %q", got, want)
+	}
+}
+
+func TestPreflightCheckContainer(t *testing.T) {
+	got := preflightCheckContainer()
+	want := "Error: not enough positional arguments: A container image positional argument is required\n\nThe checks that will be executed are the following:\n-"
+
+	container_checks := []string{"LayerCountAcceptable", "HasNoProhibitedPackagesMounted", "HasRequiredLabel", "RunAsNonRoot", "BasedOnUbi", "HasLicense", "HasUniqueTag"}
```
samira-barouti, line 35: I believe you can get the list of checks by running:
```./preflight check container --list-checks```
It is better not to hard code these values.

File: test/e2e/preflightCliTest_test.go [hunk 8]
```diff
@@ -0,0 +1,128 @@
...
+    want := "Error: not enough positional arguments: An operator image positional argument is required\n\nThe checks that will be executed are the following:\n-"
+
+    if !(strings.Contains(got,want)) {
+        t.Errorf("got %q \n\n\n\nwant %q", got, want)
+    }
+}*/
+
+func TestPreflightContainerRun(t *testing.T) {
+	preflightContainerRun()
+	content, err := ioutil.ReadFile("artifacts/results.json")
```
samira-barouti, line 78: ioutil is deprecated in go 1.16. You better use `os` or `io` packages.

samira-barouti (changes requested):
A few comments to start with :)

jomkz:
I am going to close this PR, as much has changed with the tool since this was originally introduced and there has been no activity in quite some time to address the comments. In addition, there is some question as to the utility of testing the command-line in this way as it more or less is testing that cobra works as expected. Please feel free to submit a new PR that addresses the concerns mentioned here.

PR #350, opened 2021-12-03
Supprort deploying the operator with non-default channel

### LINKED ISSUE #349: DeployableByOLM should support non-default channels
The current version of preflight uses the default channel to ensure the operator could be deployed by OLM. There are certain usecases where the teams want to test non-default channels.

A new optional env variable could be introduced which indicates the name of the channel under test. This takes precedence over `operators.operatorframework.io.bundle.channel.default.v1` in the annotations.yaml of bundle image.

N/A

## Additional context.
DeployableByOLM fetches the channel information from the key `operators.operatorframework.io.bundle.channel.default.v1` of annotations.yaml.

Jira ticket: https://issues.redhat.com/browse/EET-2116

samira-barouti:
* Fixes #349

This PR adds support for deploying operators from non-default
channel. A new environment variable (PFLT_CHANNEL) could be set
by the consumer to pass the name of the operator channel under test.
This environment variable takes precedence over the annotation key
`operators.operatorframework.io.bundle.channel.default.v1`, if present.


certification/internal/policy/operator/deployable_by_olm.go
```diff
@@ -65,6 +65,8 @@ func (p *DeployableByOlmCheck) Validate(bundleRef certification.ImageReference)
 		return false, err
 	}
 
+	log.Debug("The operator Metadata is: ", *operatorData)
```
acornett21, line 68: Since I haven't seen a log statement of logging to a pointer, do we need a `%something in the quotes` or does go just log this properly and in a human readable format?

certification/internal/policy/operator/default.go
```diff
@@ -16,6 +16,9 @@ const (
 	// IndexImageKey is the key in viper that contains the index (catalog) image URI
 	indexImageKey = "indexImage"
 
+	// channel is the key in viper that contains the subscription channel under test
+	channelKeyInViper = "channel"
```
komish, line 20: NIT: I don't like this variable name, simply because it references the technology we use to manage user-provided configuration by name.

I know this name was chosen in an effort to disambiguate this value with the value of `channelKey` which refers to the key for the channel as found in an operator's annotations. I think there's an opportunity to make this more clear, either by changing this variable (and other config-related variables, for consistency) to be identified in some other way, or changing the annotation variables, or both.

Just my 2c. Up to you on how to proceed.

certification/internal/policy/operator/deployable_by_olm_test.go
```diff
@@ -140,6 +139,9 @@ var _ = Describe("DeployableByOLMCheck", func() {
 				Expect(err).ToNot(HaveOccurred())
 				Expect(ok).To(BeTrue())
 			})
+			AfterEach(func() {
+				os.Unsetenv("PFLT_DOCKERCONFIG")
```
komish, line 143: I don't think there's anything to do here, but I wanted to confirm: This is an unrelated change, right? This doesn't impact the ability to deploy to non-default channels?

komish (changes requested):
Couple of notes in-line^. Thanks @samira-barouti!

certification/internal/policy/operator/deployable_by_olm_test.go
samira-barouti: That's right. While doing the other unit test, I realized a deferred statement might not work. That's why I updated this test as well.

certification/internal/policy/operator/default.go
samira-barouti: Fair enough. I'll try to come up with a better var name.

certification/internal/policy/operator/deployable_by_olm.go
samira-barouti: Good catch. I'll update this to print a struct field/value.

komish (approved):
/lgtm

PR #351, opened 2021-12-03
Build ppc64le image

mayurwaghmode:


bcrochet (changes requested):
Please squash these commits to a single commit.

mayurwaghmode:
@bcrochet I have squashed commits. Please review!

Dockerfile-ppc64le
```diff
@@ -0,0 +1,44 @@
+ARG quay_expiration=never
+ARG release_tag=0.0.0
+
+# golang:1.16 image created 2021-06-24T00:31:06.02014601Z
+FROM docker.io/golang AS builder
```
bcrochet, line 5: Please add an explicit tag for the version. If 1.17 is available for ppc, use that.

Dockerfile-ppc64le
```diff
@@ -0,0 +1,44 @@
...
+ARG quay_expiration
+ARG release_tag
+
+# Build the preflight binary
+COPY . /go/src/preflight
+WORKDIR /go/src/preflight
+RUN make build RELEASE_TAG=${release_tag}
+
+# ubi8:latest
+FROM registry.access.redhat.com/ubi8/ubi:latest
```
bcrochet, line 15: Is there no ubi-minimal for ppc?

Dockerfile-ppc64le
```diff
@@ -0,0 +1,44 @@
+ARG quay_expiration=never
```
bcrochet, line 1: Small nit: I'd prefer the filename to be Dockerfile.ppc64le instead of with the '-' but not a blocker.

bcrochet (changes requested):

Dockerfile-ppc64le
mayurwaghmode: Yes, it is available.

Dockerfile.ppc64le
```diff
@@ -0,0 +1,44 @@
...
+      gzip \
+      iptables \
+      findutils \
+    && dnf clean all
+
+# Install OpenShift client binary
+RUN curl -L https://mirror.openshift.com/pub/openshift-v4/clients/ocp/${OPENSHIFT_CLIENT_VERSION}/openshift-client-linux-${OPENSHIFT_CLIENT_VERSION}.tar.gz | tar -xzv -C /usr/local/bin oc
+
+# Install Operator SDK binray
+RUN curl -Lo /usr/local/bin/operator-sdk https://github.com/operator-framework/operator-sdk/releases/download/v${OPERATOR_SDK_VERSION}/operator-sdk_linux_ppc64le \
```
bcrochet, line 40: Couldn't we just make an ARCH ARG? Default it to amd64, and override it in the action? That way we don't need an entirely new Dockerfile. And i'm fine with bumping the go version in the main dockerfile.

bcrochet (changes requested):
Sorry to keep coming back with more changes. I think I'll be satisfied after this. :)

Dockerfile.ppc64le
mayurwaghmode: @bcrochet I have made the changes. Please review it!

Dockerfile
```diff
@@ -2,9 +2,10 @@ ARG quay_expiration=never
 ARG release_tag=0.0.0
```
bcrochet, line 2: You can add the arg here with the default (=amd64). Then you don't even have to change build-main.yaml.

bcrochet (changes requested):

bcrochet (approved):
/lgtm

mayurwaghmode:
@bcrochet can you assign the label again? I wanted to test with the latest changes, so did a few commits. Done with the testing.

acornett21 (approved):

bcrochet (approved):
/lgtm

Issue #355, opened 2021-12-09
File scorecard_BasicSpecCheck.json is empty if operator-sdk produces timeout


tkrishtop:
## Bug Description
Sometimes `operator-sdk scorecard` produces timeout, it seems like a several-minute outage of internal resources. It's not a fault of preflight, because direct `operator-sdk scorecard` tests which run immediately after, timeout as well. 
The problem is that in this case, preflight doesn't generate correct logs and the file scorecard_BasicSpecCheck.json is empty.

## Version and Command Invocation
preflight 1.0.5

## Steps to Reproduce:
Operator to test:
```
name: "testpmd-operator"
version: "v0.2.9"
image: "quay.io/rh-nfv-int/testpmd-operator-bundle@sha256:5e28f883faacefa847104ebba1a1a22ee897b7576f0af6b8253c68b5c8f42815"
index_image: "quay.io/tkrishtop/index-testpmd-operator-bundle:v0.2.9"
```

Preflight command:
```
podman run \
-it \
--rm \
--pull=always \
--privileged \
-e PFLT_JUNIT=true \
-e PFLT_ARTIFACTS=/artifacts \
-e PFLT_LOGFILE=/artifacts/preflight.log \
-e PFLT_LOGLEVEL=trace \
-e KUBECONFIG=/kubeconfig \
-e PFLT_INDEXIMAGE={{ OO_INDEX }} \
-e PFLT_NAMESPACE={{ preflight_namespace }} \
-e PFLT_SERVICEACCOUNT=default \
-e PFLT_SCORECARD_IMAGE={{ config_images[0] }} \
-v {{ preflight_kubeconfig }}/kubeconfig:/kubeconfig \
-v {{ preflight_artifacts }}:/artifacts \
{{ preflight_version }} \
check operator \
{{ operator.image }}
```

[Here are all log files](https://www.distributed-ci.io/jobs/1e267dc5-cbf5-47fe-a6cf-b738bc0b2c07/files).

## Expected Result
I would expect to have an operator-sdk error `error running tests context deadline exceeded` in preflight.log or results.json. I'd like to not run direct `operator-sdk` tests and rely entirely on preflight.

## Actual Result
results.json
```
"errors": [
  {
    "name": "ScorecardBasicSpecCheck",
    "elapsed_time": 240096,
    "description": "Check to make sure that all CRs have a spec block.",
[...]

PR #356, opened 2021-12-15
updating FATA string comparison to convert the error we get back from scorecard test

acornett21:
- Fixes: #355 
- updating comparison to convert to up upper case since `preflgiht` runs `operator-sdk` in a non-tty fashion and `preflight` receives `stderr` in a different format then running `operator-sdk` cmd directly.

Below are `preflight.log` entries from testing this by manually setting `--waittime 5s`
```
12-15T12:09:06-07:00" level=error msg="stdout: "
time="2021-12-15T12:09:07-07:00" level=error msg="stderr: time=\"2021-12-15T12:08:56-07:00\" level=debug msg=\"Debug logging is set\"\nError: error running tests context deadline exceeded\nUsage:\n  operator-sdk scorecard [flags]\n\nFlags:\n  -c, --config string            path to scorecard config file\n  -h, --help                     help for scorecard\n      --kubeconfig string        kubeconfig path\n  -L, --list                     Option to enable listing which tests are run\n  -n, --namespace string         namespace to run the test images in\n  -o, --output string            Output format for results. Valid values: text, json, xunit (default \"text\")\n  -l, --selector string          label selector to determine which tests are run\n  -s, --service-account string   Service account to use for tests (default \"default\")\n  -x, --skip-cleanup             Disable resource cleanup after tests are run\n  -b, --storage-image string     Storage image to be used by the Scorecard pod (default \"docker.io/library/busybox@sha256:c71cb4f7e8ececaffb34037c2637dc86820e4185100e18b4d02d613a9bd772af\")\n  -t, --test-output string       Test output directory. (default \"test-output\")\n  -u, --untar-image string       Untar image to be used by the Scorecard pod (default \"registry.access.redhat.com/ubi8@sha256:910f6bc0b5ae9b555eb91b88d28d568099b060088616eba2867b07ab6ea457c7\")\n  -w, --wait-time duration       seconds to wait for tests to complete. Example: 35s (default 30s)\n\nGlobal Flags:\n      --plugins strings   plugin keys to be used for this subcommand execution\n      --verbose           Enable verbose logging\n\ntime=\"2021-12-15T12:09:01-07:00\" level=fatal msg=\"error running tests context deadline exceeded\"\n"
time="2021-12-15T12:09:10-07:00" level=error msg="stderr: operator-sdk scorecard failed to run properly. Please review the /Users/acornett/go/src/github.com/acornett21/openshift-preflight/artifacts/operator_bundle_scorecard_OlmSuiteCheck.json file for more information. "
```

### DIFF
--- a/certification/internal/engine/operatorsdk.go
+++ b/certification/internal/engine/operatorsdk.go
@@ -73,16 +73,9 @@ func (o operatorSdkEngine) Scorecard(image string, opts cli.OperatorSdkScorecard
 		//
 		// We also conclude/assume that "FATA" being in stderr would indicate an error in the
 		// check execution itself.
-		if stderr.Len() != 0 && strings.Contains(stderr.String(), "FATA") {
-			log.Error("stdout: ", stdout.String())
+		if stderr.Len() != 0 && strings.Contains(strings.ToUpper(stderr.String()), "FATA") {
+			log.Error("operator-sdk scorecard failed to run properly.")
 			log.Error("stderr: ", stderr.String())
-			log.Error("stderr: operator-sdk scorecard failed to run properly. " +
-				"Please review the " + artifacts.Path() + "/" + opts.ResultFile + " file for more information. ")
-
-			if err := o.writeScorecardFile(opts.ResultFile, stderr.String()); err != nil {
-				log.Error("unable to copy result to artifacts directory: ", err)
-				return nil, err
-			}
 
 			return nil, fmt.Errorf("%w: %s", errors.ErrOperatorSdkScorecardFailed, err)
 		}


komish (commented):
@acornett21 I'm a bit confused. When OperatorSDK experiences a timeout, you're seeing this log line.

```
Error: error running tests context deadline exceeded
```

This change takes the stderr line that we get from OperatorSDK and makes it upper case, but still only searches for FATA which doesn't exist in the error string we get back from OperatorSDK for this error. 

It seems like this change wouldn't catch that error case, but I might be misunderstanding. Could you clarify?

---

Further, I'm wondering if we can clarify our messaging (e.g. what gets returned from each check if they fail/error), as right now this log line would only ever exist in preflight.log, and the logfile referenced by this check's message would be empty or not exist.

Do you think it makes sense to create another issue for clarifying our messaging? 

---

Lastly, do you think OperatorSDK should be logging context deadline exceeded with severity contexts? I.e., should we open up a bug with OperatorSDK?

acornett21:
@komish Yeah this is really confusing. I had to look at the logs that were in the issue and the SDK code to see what is what. For your points:

1. It does catch the error because the entire stderr stream has `fatal` in the message, even though the error isn't `log.Fatal()` (will past entire log below)
2. The log file for this check does get populated (which is what DCI requested). I'm not sure we can cleanup our logs until OperatorSDK implements our ask for exit code 1 vs exit code 0, but might be missing cases where we could do a better job.
3. I thought about opening a bug, but like the issue we already raised, would they implement it? and how should it be raised? I'm unsure. I think if the implemented 1 vs 0 like above, this wouldn't even be an issue and we wouldn't have to look/poke around in stderr/stout

Logs that get written to `operator_bundle_scorecard_BasicSpecCheck.json`

```
time="2021-12-17T13:54:30-07:00" level=debug msg="Debug logging is set"
Error: error running tests error creating ConfigMap Unauthorized
Usage:
  operator-sdk scorecard [flags]

Flags:
  -c, --config string            path to scorecard config file
  -h, --help                     help for scorecard
      --kubeconfig string        kubeconfig path
  -L, --list                     Option to enable listing which tests are run
  -n, --namespace string         namespace to run the test images in
  -o, --output string            Output format for results. Valid values: text, json, xunit (default "text")
  -l, --selector string          label selector to determine which tests are run
  -s, --service-account string   Service account to use for tests (default "default")
  -x, --skip-cleanup             Disable resource cleanup after tests are run
  -b, --storage-image string     Storage image to be used by the Scorecard pod (default "docker.io/library/busybox@sha256:c71cb4f7e8ececaffb34037c2637dc86820e4185100e18b4d02d613a9bd772af")
  -t, --test-output string       Test output directory. (default "test-output")
  -u, --untar-image string       Untar image to be used by the Scorecard pod (default "registry.access.redhat.com/ubi8@sha256:910f6bc0b5ae9b555eb91b88d28d568099b060088616eba2867b07ab6ea457c7")
  -w, --wait-time duration       seconds to wait for tests to complete. Example: 35s (default 30s)

Global Flags:
      --plugins strings   plugin keys to be used for this subcommand execution
      --verbose           Enable verbose logging

time="2021-12-17T13:54:30-07:00" level=fatal msg="error running tests error creating ConfigMap Unauthorized"
```
these are the same logs that would get written by OperatorSDK that DCI mentioned in the issue.

acornett21:
@komish FYI: I put some code links from the SDK in the issue when I started looking at this.

komish:
Understood. I think there are a couple of challenges here, then.

- We probably don't want to write non-JSON to the **operator_bundle_scorecard_BasicSpecCheck.json** file. 

- I'm uncertain if the log line `time="2021-12-17T13:54:30-07:00" level=fatal msg="error running tests error creating ConfigMap Unauthorized"` is coming from OperatorSDK, or coming from preflight itself. The only reason I mention this is because the code you linked doesn't seem to return Fatal, at least not in the code you linked. To that end, if we (preflight) are emitting that error, we don't want to fail a test based on our own error handling. This probably just needs some more investigation.

Overall, it just sounds like we're crossing some error handling streams here and it's causing some confusion. To point #2 - I do still think that we might benefit from being more clear, such that users of preflight would know to look at preflight.log in the event that the check errors (instead of fails because it doesn't meet our failure criteria).

I don't think we have to solve all of those problems here, but I do think at least the first two bullets would need to be addressed to resolve this issue.

bcrochet:
I don't think this is how we want to handle this. The "FATA" (which is just abbreviated) is a very specific exit when log.Fatal() is used. I don't think we want to mix up this particular error case with something else. If we have another way of detecting whatever error you are trying to detect here, I would rather see it that way.

acornett21:
@komish I pasted the wrong error from testing another operator. I fixed the above for this scenario/code fix.

@bcrochet I felt you might say that, this is another option that works ie adding an or condition

```
stderr.Len() != 0 && (strings.Contains(stderr.String(), "FATA") || strings.Contains(stderr.String(), "error running tests"))
``

if we want this in a separate block `if` block with a separate comment, that's fine too. We could also jump on a call and we can debug in real time and get either of your thoughts on this.

bcrochet (approved):
/lgtm

jomkz (approved):

PR #362, opened 2022-01-07
updating dependencies and go version

acornett21:
The motivation for this PR was to test that deprecations happening in `operatorframework api` did not affect `DeployableByOLM` check. In updating that dependency I feel like it makes sense to update other reference dependences, and it doing so we are forced to move to a new version of go since other dependencies moved to that.

- updating dependencies to make sure preflight works with various `api` changes
- updating the version of `go`
- tested changes against `check operator quay.io/acornett/operator-certification-operator-bundle:latest`


bcrochet (approved):
/lgtm

Issue #367, opened 2022-01-18
Create a New Parameter to Expose Scorecard `--wait-time` flag  


acornett21:
## Is your feature request related to a problem? Please describe.
Since we are now saying to increase the wait-time in the `help` when errors happen in preflight, we should probably expose this value for users to try to increase.

## Describe the solution you'd like.
We can keep the default value that we currently have hardcoded, and make a new parameter `PFLT_SCORECARD_WAIT_TIME` as optional for user to configure.

## Describe alternatives you've considered.
N/A

## Additional context.
N/A

PR #368, opened 2022-01-20
adding in optional SCORECARD_WAIT_TIME env variable to pass to scorecard

acornett21:
- Fixes: #367 
- Adding in SCORECARD_WAIT_TIME so that if user's experience a timeout and preflight returns an error saying they can try to increase the timeout, this actually exposes that value so they can try again with an increased value before they raise an issue against preflight.
- Updated documentation

### DIFF
--- a/certification/internal/policy/operator/scorecard_check.go
+++ b/certification/internal/policy/operator/scorecard_check.go
@@ -1,6 +1,7 @@
 package operator
 
 import (
+	"fmt"
 	"os"
 	"strings"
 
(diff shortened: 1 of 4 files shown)


samira-barouti (approved):
/lgtm

certification/internal/policy/operator/scorecard_check.go
```diff
@@ -42,7 +45,7 @@ func (p *scorecardCheck) getDataToValidate(bundleImage string, selector []string
 		Namespace:      namespace,
 		ServiceAccount: serviceAccount,
 		Verbose:        true,
-		WaitTime:       "240s",
+		WaitTime:       fmt.Sprintf("%s%s", waitTime, "s"),
```
bcrochet, line 48: Small nit, feel free to ignore, but... The second arg? Is that necessary?

bcrochet (commented):
Small nit, but otherwise lgtm

certification/internal/policy/operator/scorecard_check.go
acornett21: @bcrochet You're thinking`fmt.Sprintf("%ss", waitTime)` right? That's probably clearer/cleaner.

acornett21: Updated this for clarity.

bcrochet (approved):
/lgtm

jomkz (approved):

Issue #380, opened 2022-01-28
Migrate from custom `usage` information in `cmd` help to native cobra help


acornett21:
## Is your feature request related to a problem? Please describe.
Right now, if another flag/cmd/etc is added we do not get that flags output information for `free`, since we are overlaying that information with custom `usage` information.

## Describe the solution you'd like.
1. Remove any method calls to `.SetUsageTemplate(usage)`.
2. Add in additional flags `--list-checks` to list the checks where applicable.
3. Check that the nesting of flags/cmds is still accurate and provides a good UX, especially for `check container` cmd.

## Describe alternatives you've considered.
We could keep maintaining the `usage` var, but this is more problematic because then we have to deal with formatting, and remember to add new flags there manually.

## Additional context.
This was noticed when trying to add in a `--submit` for the new requirements for `check container` and there might be other places where we are not using cobra properly, but this will be a good starting point to make the UX better for the `check container`

PR #382, opened 2022-01-28
removing calls to SetUsageTemplate, instead adding in a flag list-checks 

acornett21:
- Fixes: #380 
- Removing calls to `SetUsageTemplate`
- Adding in a `list-checks` as a  `PersistentFlags` so it's available in both `operator` and `container` checks


jomkz (approved):
/lgtm

bcrochet (approved):
/lgtm

PR #383, opened 2022-01-31
build multi arch images for main and release

### LINKED ISSUE #371: [Container Cert] Develop a Preflight Distribution Pipeline
Partners will need to be able to download Preflight for on premise use during container certification.   
1. We need to build binaries as GitHub releases. There should be binary for each supported architecture. (amd64, ppc64le. s390x etc)
2. We need to create containers containing the binary and prepare for delivering them through HACBS (Aug/Sep timeframe) or CPaaS

itroyano:
The goals of this PR are:

- [ ] Remove the separate ppc64le build and consolidate it into the main and release builds.
- [ ] Add all supported architecture builds in a maintainable (and expandable) way.
- [ ] Standardize image naming.
- [ ] Add a mechanism to build and publish Preflight binaries for a release.

Implemention:

PPC64 yaml is now removed.

Main and Release use a matrix strategy to build multiple architecture images of Preflight, in separate processes, and name them uniquely.

As shown previously in the PPC64 build,  the `archs` attribute of `redhat-actions/buildah-build` will spin up a QEMU vm of the appropriate architecture, to natively compile the binary.
hence no cross-compilation (GOOS or GOARCH) flags are needed in the `make` command itself, for images.

For the binaries build we will use cross-compilation in `make build-multi-arch`, and then upload them to Github using the action `AButler/upload-release-assets` 

Fixes #371  

Signed-off-by: Igor Troyanovsky <itroyano@redhat.com>


itroyano:
@mayurwaghmode

itroyano:
@bcrochet

.github/workflows/build-main.yml
```diff
@@ -6,7 +6,11 @@ on:
 
 jobs:
   build-main:
-    name: Build and push a main snapshot image
+    name: Build and push main snapshot images
+    strategy: 
+      matrix:
+        architecture: [amd64,ppc64le,s390x]
```
lslebodn, line 12: ubi8 image is built also for arm64(aarch64)
```
[root@localhost ~]# skopeo inspect --raw docker://registry.access.redhat.com/ubi8-micro:8.5 | jq .manifests[].platform
{
  "architecture": "amd64",
  "os": "linux"
}
{
  "architecture": "arm64",
  "os": "linux"
}
{
  "architecture": "ppc64le",
  "os": "linux"
}
{
  "architecture": "s390x",
  "os": "linux"
}
```
and the same for ubi9 beta
```
[root@localhost ~]# skopeo inspect --raw docker://registry.access.redhat.com/ubi9-beta/ubi-micro:9.0.0beta-455 | jq .manifests[].platform
{
  "architecture": "amd64",
  "os": "linux"
}
{
  "architecture": "arm64",
  "os": "linux"
}
{
  "architecture": "ppc64le",
  "os": "linux"
}
{
  "architecture": "s390x",
  "os": "linux"
}
```

itroyano: Thanks. will update that.

So we also need other os supported e.g. darwin (mac) or windows?

@tonytcampbell FYI

lslebodn: Thank you for changes. 

I hope it works as expected cause sometime there are different names for architectures in different context
* amd64 vs x86_64
* arm64 vs aarch64

I do not feel suitable for review cause I am not sure which semantic should be used in this context.
I would be guessing I let other review the code.

But lgtm from my POV

jomkz:
I added a comment to the original issue regarding the format that we want to use for the image tags:

https://github.com/redhat-openshift-ecosystem/openshift-preflight/issues/371#issuecomment-1029118359

.github/workflows/build-main.yml
itroyano: Thanks @lslebodn

.github/workflows/build-main.yml
```diff
@@ -35,7 +48,7 @@ jobs:
       uses: redhat-actions/push-to-registry@v2
       with:
         image: preflight
-        tags: ${{ steps.build-image.outputs.tags }}
```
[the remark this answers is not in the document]
itroyano: I'm not entirely sure how this will behave, when split into parallel sub-processes by `matrix`.
to be on the safe side and avoid any race conditions, I'm specifying the tag explicitly as built.

.github/workflows/build-release.yml
```diff
@@ -32,7 +43,7 @@ jobs:
       uses: redhat-actions/push-to-registry@v2
       with:
         image: preflight
-        tags: ${{ steps.build-image.outputs.tags }}
```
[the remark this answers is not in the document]
itroyano: see above.

bcrochet (approved):
/lgtm

jomkz (approved):

Issue #384, opened 2022-01-31
Corrupt Operator-SDK binary in 1.0.7 container image


jomkz:
## Bug Description

The preflight container image for version `1.0.7` contains a corrupt operator-sdk binary.

## Version and Command Invocation

```
preflight version 1.0.7 <commit: c69c72930d8b986dece936bc7daef561e230b7bb>
```

## Steps to Reproduce:

1) `$ podman pull quay.io/opdev/preflight:1.0.7`
2) `$ podman run -it --entrypoint /bin/bash quay.io/opdev/preflight:1.0.7`
3) `# operator-sdk --version`

## Expected Result

Expected the operator-sdk version to be printed.

```
# operator-sdk --version
operator-sdk version: "v1.14.0", commit: "78f08b4852faf344ad3ef457c54f86087aaa0a0a", kubernetes version: "1.21", go version: "go1.16.9", GOOS: "linux", GOARCH: "amd64"
```

## Actual Result

The binary is corrupt.

```
# operator-sdk --version
/usr/local/bin/operator-sdk: line 1: Not: command not found
```

## Additional Context

It looks like the `ARCH` env var is not declared for the final stage of the image build, resulting in a blank value for the ARCH to be used when downloading the operator-sdk binary.
https://github.com/redhat-openshift-ecosystem/openshift-preflight/blob/c69c72930d8b986dece936bc7daef561e230b7bb/Dockerfile#L17-L19

This results in a 404 error and the text "Not Found" to be saved as `operator-sdk` on the filesystem. 

```
# ls -la /usr/local/bin
total 121044
drwxr-xr-x.  2 root root     4096 Jan 28 15:43 .
drwxr-xr-x. 12 root root     4096 Jan 28 15:43 ..
-rwxr-xr-x.  1 root root 72737416 Jun 25  2021 oc
-rwxr-xr-x.  1 root root        9 Jan 28 15:43 operator-sdk
-rwxr-xr-x.  1 root root 51184727 Jan 28 15:43 preflight
# cat /usr/local/bin/operator-sdk
Not Found
```

PR #385, opened 2022-01-31
adding in ARG ARCH in the main container build stage

acornett21:
- Fixes: #384 
- Added in `ARG ARCH` to the main build
- Added `--fail` to all curl `cmds` so if a binary does not exist we will fail our docker build


bcrochet (approved):
/lgtm

jomkz (approved):

PR #386, opened 2022-02-01
Move BundleValidate to a standalone method and call from DBOLM

### LINKED ISSUE #365: Preflight attempts to run a v4.9 operator prior to checking for deprecated Kube 1.22/OpenShift 4.9 APIs
## Bug Description

Preflight attempts to run a v4.9 operator prior to the check for deprecated Kubernetes 1.22/OpenShift 4.9 APIs.

This causes the Preflight test to waste several minutes waiting for the operator to become ready - when it's never going to happen.

After this finally fails the Kube 1.22 API test is run, which indeed fails and explains why the operator could not deploy.

1.0.6

## Steps to Reproduce:
(How can we reproduce this?)
1) Run preflight on an operator that has v4.9 in its supported scope but has Kube 1.22 API violations.
2) 
3)

## Expected Result

Preflight will fail out the operator quickly after checking for Kube 1.22 API violations.

## Actual Result

Preflight wastes several minutes waiting for the operator to deploy (until the context deadline is exceeded), and only then runs the kube 1.22 API test.

## Additional Context

See attached log

[preflight2.log](https://github.com/redhat-openshift-ecosystem/openshift-preflight/files/7872181/preflight2.log)

bcrochet:
In order for DeployableByOLM to check that the bundle is valid before
actually deploying, this moves the bundle validate code to a shared
area, so it can be called by both the DeployableByOLM check and the
actual Bundle Validate check.

DeployableByOLM does not do extensive checking of the result of
BundleValidate. If there is an error, or Passed: false is in the
report, it will fail.

Fixes #365 


samira-barouti (approved):
/lgtm

acornett21 (approved):

PR #388, opened 2022-02-02
Introduce Pyxis Engine

bcrochet:
This is the start of the Pyxis engine. It also adds support for a
PYXIS_HOST envvar, and a PYXIS_API_TOKEN envvar and CLI param.

The PyxisEngine should export useful methods to those that use it.
It is up to the user to create interfaces if necessary for testing
purposes.

### DIFF
--- a/cmd/check_container.go
+++ b/cmd/check_container.go
@@ -15,6 +15,7 @@ import (
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/version"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
 )
 
 var checkContainerCmd = &cobra.Command{
@@ -103,5 +104,8 @@ var checkContainerCmd = &cobra.Command{
 }
 
 func init() {
+	checkContainerCmd.Flags().String("pyxis-api-token", "", "API token for Pyxis authentication")
+	viper.BindPFlag("pyxis_api_token", checkContainerCmd.Flags().Lookup("pyxis-api-token"))
+
 	checkCmd.AddCommand(checkContainerCmd)
 }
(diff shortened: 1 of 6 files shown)


cmd/utils.go
```diff
@@ -44,6 +44,10 @@ func preRunConfig(cmd *cobra.Command, args []string) {
 	// Set up scorecard wait time default
 	viper.SetDefault("scorecard_wait_time", DefaultScorecardWaitTime)
 
+	// Set up pyxis host
+	viper.SetDefault("pyxis_host", DefaultPyxisHost)
+	viper.SetDefault("pyxis_api_token", "")
```
acornett21, line 49: Is this just temporary? We are taking this value in as an argument, do we need to expose it as an arg and an env?

cmd/check_container.go
```diff
@@ -93,6 +93,8 @@ var checkContainerCmd = &cobra.Command{
 }
 
 func init() {
+	checkContainerCmd.Flags().String("pyxis_api_token", "", "API token for Pyxis authentication")
```
acornett21, line 96: Nit: `pyxis-api-token` for consistency with other flags

cmd/defaults.go
```diff
@@ -8,4 +8,5 @@ var (
 	DefaultNamespace         = "default"
 	DefaultServiceAccount    = "default"
 	DefaultScorecardWaitTime = "240"
+	DefaultPyxisHost         = "pyxis.engineering.redhat.com"
```
acornett21, line 11: I don't think this URL is accessible without SSO or VPN. I think we want to use `catalog.redhat.com` for this.

cmd/utils.go
bcrochet: I'd like to see both tbh. Then the secret (api token) can be injected as an env var and not have to specify on the command line.

acornett21: That makes sense and probably provides better usability if someone is running this over and over again before `submit`.

acornett21 (approved):

samira-barouti (approved):
/lgtm

jomkz (approved):

PR #396, opened 2022-02-08
Use qemu action instead of apt install

itroyano:
Signed-off-by: Igor Troyanovsky <itroyano@redhat.com>


.github/workflows/build-main.yml
```diff
@@ -55,3 +53,4 @@ jobs:
 
     - name: Print image url
       run: echo "Image pushed to ${{ steps.push-image.outputs.registry-paths }}"
+
```
acornett21, line 56: Can you add back the new line at the end?

acornett21 (approved):

PR #399, opened 2022-02-09
default pyxis host

itroyano:
change the default host in the doc to match the host in the code: https://catalog.redhat.com/api/containers/v1/ui/#/


bcrochet (approved):
/lgtm

Issue #376, opened 2022-01-27
[Container Cert] Display response data from Pyxis after submit


tonytcampbell:
Display the Image ID from Pyxis back to the user.  The user will need this user id to access the results in Connect.  
 
Also add verbiage indicating to the Partner that once all checks have passed they will be reviewed by Red Hat for final certification.  We should provide a link to connect.redhat.com where the Partner can monitor the progress.

PR #400, opened 2022-02-09
Add code for submitting a certification to Pyxis

bcrochet:
This begings the implementation of submitting to Pyxis.

Some depuplication needs to be done. Some processing of the URL needs
to take place as well.


certification/pyxis/submit.go
```diff
@@ -0,0 +1,54 @@
...
+func (p *pyxisEngine) SubmitResults(containerImage string) error {
+	ctx := context.Background()
+	projectId := p.ProjectId
+	if projectId == "" {
+		return errors.ErrEmptyProjectID
+	}
+	if strings.HasPrefix(projectId, "ospid-") {
+		projectId = strings.Split(projectId, "-")[1]
+	}
+	project, err := p.getProject(ctx)
```
itroyano, line 20: there's a task for displaying the response from Pyxis to the user - https://github.com/redhat-openshift-ecosystem/openshift-preflight/issues/376
can we add it here?

acornett21: We don't have that value in the `struct` yet so we can implement #376 separately. I think we might want to have SubmitResults, return that value, then the man `cmd` can display it before it exits.

certification/pyxis/submit.go
```diff
@@ -0,0 +1,54 @@
...
+	if err != nil {
+		log.Error(err, "could not retrieve project")
+		return err
+	}
+	oldProject := project
+
+	if project.CertificationStatus == "Started" {
+		project.CertificationStatus = "In Progress"
+	}
+
```
acornett21, line 30: we also need to add the `docker config` to the project before updating it. maybe just put a `todo` in for now, since I've added that `flag/env` in my PR?

acornett21 (commented):
Just a couple of things that I don't want us to forget, but I really like the direction this implementation is going.

certification/pyxis/submit.go
bcrochet: Yes. This is really not meant to be an exhaustive patch. It will need to be extended for all of the requirements, but I'm hoping this starts us in the right direction.

certification/pyxis/pyxis.go
```diff
@@ -0,0 +1,179 @@
...
+	return &[]Layer{}, nil
+}
+
+func (p *pyxisEngine) createImage(ctx context.Context, certImage *CertImage) (*CertImage, error) {
+	b, err := json.Marshal(certImage)
+	if err != nil {
+		log.Error(err)
+		return nil, err
+	}
+	req, _ := newRequestWithApiToken(ctx, http.MethodPost, getPyxisUrl("/images"), p.ApiToken)
```
samira-barouti, line 51: It seems the `/` before `/images` is redundant.

certification/pyxis/pyxis.go
```diff
@@ -0,0 +1,179 @@
...
+	rpmManifest := RPMManifest{
+		ImageID: imageId,
+		RPMS:    rpms,
+	}
+	b, err := json.Marshal(rpmManifest)
+	if err != nil {
+		log.Error(err)
+		return err
+	}
+	req, _ := newRequestWithApiToken(ctx, http.MethodPost, getPyxisUrl(fmt.Sprintf("/images/id/%s/rpm-manifest", imageId)), p.ApiToken)
```
samira-barouti, line 89: same here

certification/pyxis/pyxis.go
```diff
@@ -0,0 +1,179 @@
...
+
+func NewPyxisEngine(apiToken string, projectId string, httpClient HTTPClient) *pyxisEngine {
+	return &pyxisEngine{
+		ApiToken:  apiToken,
+		ProjectId: projectId,
+		Client:    httpClient,
+	}
+}
+
+func (p *pyxisEngine) GetRedHatLayers() (*[]Layer, error) {
```
samira-barouti, line 41: Is this method being used?

certification/pyxis/pyxis.go
```diff
@@ -0,0 +1,179 @@
...
+	if err != nil {
+		log.Error(err)
+		return err
+	}
+
+	return nil
+}
+
+func (p *pyxisEngine) getProject(ctx context.Context) (*CertProject, error) {
+	req, _ := newRequestWithApiToken(ctx, http.MethodGet, getPyxisUrl(fmt.Sprintf("/projects/certification/id/%s", p.ProjectId)), p.ApiToken)
```
samira-barouti, line 102: are we going to ignore the error?

certification/pyxis/submit.go
acornett21: Yeah, it totally is the right direction and we can add to it. Since the requirements are across a bunch of issues and in documents, `TODO`'s are just easier for me to look and see what is missing.

certification/pyxis/pyxis.go
bcrochet: Good catch.

certification/pyxis/pyxis.go
bcrochet: Not yet. I'll move it to its own file and leave it out of this patch.

certification/pyxis/pyxis.go
bcrochet: I'll add checks.

certification/pyxis/submit.go
```diff
@@ -0,0 +1,54 @@
...
+	certImage := new(CertImage)
+	certImage, err = p.createImage(ctx, certImage)
+	if err != nil {
+		log.Error(err)
+		return nil, nil, err
+	}
+
+	rpms := make([]RPM, 10)
+	err = p.createRPMManifest(ctx, certImage.ImageID, rpms)
```
samira-barouti, line 47: I am not familiar with Pyxis, so I apologize if my question is trivial. I was wondering if passing an empty RPM array is intentional, or the future commits/PRs will initialize this var.

Issue #369, opened 2022-01-21
[Container Cert] Send Container Certification Results to Pyxis


tonytcampbell:
`preflight check container` command should include an option (maybe a flag) to `--submit` results.  When this options is selected the results should be submitted to the Pyxis API.

amisstea:
@jomkz we just added https://issues.redhat.com/browse/ISV-1664 to our backlog with the expectation that preflight could "phone home" the Started -> In Progress certification state transition. It could happen at any time the partner first starts working with preflight but it could also occur during submission so this seems to be the most relevant issue. Please take a look and let me know if you have any questions or concerns. Thanks!

acornett21:
@amisstea Should we do a call to `GET` call to `projects/certification/id/1234` to verify the status is in a status of `Started` before we call `PATCH` with the new `In Progress` status?

amisstea:
@acornett21 yes, that is probably a good idea. We may allow it to be `PATCH`'d to `In Progress` if currently in that state but going from `Published` to `In Progress` is expected to return an error.

PR #402, opened 2022-02-09
implementing submit flag and saving cert image info and rpm manifest information to disk

acornett21:
- Relates: #369 
- Added in flags to support submission to pyxis
- Converting data that we get from `crane` about the image to pyxis format and saving to disk
- Converting data that we get about the RPMs in the image to pyxis format and saving to disk
- Adding in method to send test results to pyxis


acornett21:
/do-not-merge/hold

certification/pyxis/types.go
```diff
@@ -1,43 +1,48 @@
...
+	Deleted                bool         `json:"deleted" default:"false"`
+	DockerImageDigest      string       `json:"docker_image_digest,omitempty"`
+	DockerImageID          string       `json:"docker_image_id,omitempty"`
+	ImageID                string       `json:"image_id,omitempty"`
+	ISVPID                 string       `json:"isv_pid,omitempty"`
+	ParsedData             *ParsedData  `json:"parsed_data,omitempty"`
+	RawConfig              string       `json:"raw_config,omitempty"`
+	Repositories           []Repository `json:"repositories,omitempty"`
+	SumLayerSizeBytes      int64        `json:"sum_layer_size_bytes,omitempty"`
+	UncompressedTopLayerId string       `json:"uncompressed_top_layer_id,omitempty"` //todo-adam check if this is required
```
bcrochet, line 14: It's not a hard enforcement by Pyxis, but we do need it.

bcrochet (changes requested):
Couple of comments.

certification/internal/engine/engine.go
```diff
@@ -279,3 +298,121 @@ func untar(dst string, r io.Reader) error {
...
+		return err
+	}
+
+	pkgList, err := db.ListPackages()
+	if err != nil {
+		return err
+	}
+
+	// covert rpm struct to pxyis struct
+	var rpms []pyxis.RPM
```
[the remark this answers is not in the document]
acornett21: switched to make

certification/pyxis/types.go
acornett21: Yeah, I double checked the spec, the `TODO's` in types file are now more generic, and anyone can find them.

certification/internal/engine/engine.go
```diff
@@ -93,6 +99,19 @@ func (c *CraneEngine) ExecuteChecks() error {
...
+		err := writeCertImage(img)
+		if err != nil {
+			return err
+		}
+
+		err = writeRPMManifest(containerFSPath)
+		if err != nil {
+			return err
+		}
+	}
```
samira-barouti, line 113: I feel this block should be placed after the check execution. I am unsure of the usecase; can a partner submit a result even though some checks fail?

certification/internal/engine/engine.go
```diff
@@ -279,3 +298,126 @@ func untar(dst string, r io.Reader) error {
...
+	config, err := img.ConfigFile()
+	if err != nil {
+		return fmt.Errorf("%w: %s", errors.ErrImageInspectFailed, err)
+	}
+
+	digest, _ := img.Digest()
+	manifest, _ := img.Manifest()
+	rawConfig, _ := img.RawConfigFile()
+	size, _ := img.Size()
+	layers, _ := img.Layers()
```
samira-barouti, line 312: Are we ignoring the errors?
Also, I think we can refactor this into a single method and use `reflect` library to call methods dynamically. This seems to be a good example:
https://www.geeksforgeeks.org/reflect-methodbyname-function-in-golang-with-examples/

certification/internal/engine/engine.go
```diff
@@ -279,3 +298,126 @@ func untar(dst string, r io.Reader) error {
...
+			UncompressedLayerSizes: layerSize,
+		},
+		RawConfig:         string(rawConfig),
+		SumLayerSizeBytes: sumLayerSizeBytes(layers),
+		//TODO: figure out how to get/convert this data from crane
+		UncompressedTopLayerId: "placeholder",
+	}
+
+	// calling MarshalIndent so the json file written to disk is human-readable when opened
+	certImageJson, _ := json.MarshalIndent(certImage, "", "    ")
```
samira-barouti, line 339: Error handling is missing...

certification/internal/engine/engine.go
```diff
@@ -279,3 +298,126 @@ func untar(dst string, r io.Reader) error {
...
+
+		rpms = append(rpms, rpm)
+	}
+
+	rpmManifest := pyxis.RPMManifest{
+		RPMS: rpms,
+	}
+
+	// calling MarshalIndent so the json file written to disk is human-readable when opened
+	rpmManifestJson, _ := json.MarshalIndent(rpmManifest, "", "    ")
```
samira-barouti, line 389: Same here

certification/internal/engine/engine.go
```diff
@@ -279,3 +298,126 @@ func untar(dst string, r io.Reader) error {
...
+	fileName, err := artifacts.WriteFile(viper.GetString("rpm_manifest_file_name"), string(rpmManifestJson))
```
samira-barouti, line 390: It would be good to ensure `len(viper.GetString("rpm_manifest_file_name") >0` before writing to the file.

cmd/check_container.go
```diff
@@ -34,6 +39,19 @@ var checkContainerCmd = &cobra.Command{
 		if len(args) != 1 {
 			return fmt.Errorf("%w: A container image positional argument is required", errors.ErrInsufficientPosArguments)
 		}
+
+		if s, _ := cmd.Flags().GetBool("submit"); s {
+			if d, _ := cmd.Flags().GetString("docker-config"); len(d) == 0 {
+				return fmt.Errorf("%w: A docker configuration file must be present when calling --submit", errors.ErrNoDockerConfig)
+			}
```
samira-barouti, line 46: Out of curiosity: why do we need docker auth token to submit the result?

certification/internal/engine/engine.go
samira-barouti, line 113: I feel this block should come after the check execution. I am unsure of the use case: can a partner submit the preflight result even though some checks are failing?

samira-barouti (changes requested):

certification/internal/engine/engine.go
acornett21: Two things:
- It is here because `crane` has already pull/extracted the image on disk, if we do it later, we would have to pull/extract again. I wanted to make sure it was before any of the defer's were called.
- Yes, they can submit failed results, preflight is not validating the output, we submit the raw results to pyxis.

certification/internal/engine/engine.go
bcrochet: Manifest() and Layers() errors should not be ignored. It'd be safe to ignore RawConfigFile(), Digest() and Size() (they just return nil no matter what).

certification/internal/engine/engine.go
acornett21: This is coming from `default`, but I can add a check.

bcrochet: And it can't be overwritten with an envvar, right?

samira-barouti: @acornett21 if the value is required, we better do a check. There is a chance that the value is overwritten by empty string via env var (not sure if viper is smart enough to disregard the empty value and use the default in this case)

acornett21: @bcrochet Is there a way to check that to be certain?

cmd/check_container.go
```diff
@@ -99,13 +118,89 @@ var checkContainerCmd = &cobra.Command{
...
+				return err
+			}
+
+			testResultsJsonFile, err := os.Open(path.Join(artifacts.Path(), TestResultsFileName))
+			defer testResultsJsonFile.Close()
+			if err != nil {
+				return err
+			}
+
+			testResultsBytes, err := ioutil.ReadAll(testResultsJsonFile)
```
samira-barouti, line 165: `ioutil` is deprecated in Go 1.16 (https://go.dev/doc/go1.16#ioutil). We can use `io` or `os` packages instead.

certification/internal/engine/engine.go
```diff
@@ -279,3 +298,162 @@ func untar(dst string, r io.Reader) error {
...
+		UncompressedTopLayerId: "placeholder",
+	}
+
+	// calling MarshalIndent so the json file written to disk is human-readable when opened
+	certImageJson, err := json.MarshalIndent(certImage, "", "    ")
+	if err != nil {
+		return err
+	}
+
+	fileName, err := artifacts.WriteFile(viper.GetString("cert_image_file_name"), string(certImageJson))
```
samira-barouti, line 364: I am leaving a note here to keep track of checking for required values from viper (same as line 426)

samira-barouti (changes requested):

cmd/check_container.go
acornett21: I made this change, just an FYI they call `io.ReadAll()` directly now. See [here](https://github.com/golang/go/blob/master/src/io/ioutil/ioutil.go#L25)

certification/internal/engine/engine.go
bcrochet: As discussed, these are not expected to be changed by the end user, and finding a package to place these string constants in is the better approach.

acornett21: I created a new `defaults.go` file in the `certification` package. This doesn't create a circular dependency that I was trying to avoid by using viper, or the other `default.go` files in say `cmd`.

jomkz (approved):

cmd/check_container.go
```diff
@@ -99,13 +118,89 @@ var checkContainerCmd = &cobra.Command{
...
+	checkContainerCmd.Flags().BoolP("submit", "s", false, "submit check container results to red hat")
+	viper.BindPFlag("submit", checkContainerCmd.Flags().Lookup("submit"))
+
+	checkContainerCmd.Flags().StringP("docker-config", "d", "", "path to docker config.json file")
+	viper.BindPFlag("docker_config", checkContainerCmd.Flags().Lookup("docker-config"))
+
 	checkContainerCmd.Flags().String("pyxis-api-token", "", "API token for Pyxis authentication")
 	viper.BindPFlag("pyxis_api_token", checkContainerCmd.Flags().Lookup("pyxis-api-token"))
 
+	checkContainerCmd.Flags().String("pyxis-project-id", "", "Project ID for a Pyxis container project")
```
bcrochet, line 202: I thought this was going to be certification-project-id?

bcrochet (changes requested):

certification/internal/engine/engine.go
```diff
@@ -279,3 +298,162 @@ func untar(dst string, r io.Reader) error {
...
+
+	log.Tracef("rpm manifest written to disk: %s", fileName)
+
+	return nil
+}
+
+func sumLayerSizeBytes(layers []v1.Layer) int64 {
+	var sum int64
+	for _, layer := range layers {
+		size, _ := layer.Size()
```
samira-barouti, line 439: seems like we are ignoring error here. Do we want to ignore the error? or fail the submit?

samira-barouti (commented):
A question on the error handling

cmd/check_container.go
acornett21: It is, I had yet to rebase.

certification/internal/engine/engine.go
acornett21: Good catch, I forgot about this call. This has been fixed.

samira-barouti (approved):
/lgtm

bcrochet (approved):
/lgtm

PR #405, opened 2022-02-11
Add Unit tests for RunnableContainerCheck

### LINKED ISSUE #358: [Container Cert]  Start container and ensure it reaches a running state
## Summary

Preflight should start a container (e.g. with something like `podman run`), and
ensure that the container eventually reaches a running state.

This check is a part of the **Container** policy (i.e. the `preflight check
container` target).

For new contributors: [Implement a check](https://github.com/redhat-openshift-ecosystem/openshift-preflight/blob/main/docs/dev/IMPLEMENT_A_CHECK.md)

## To Pass the Check

A container, once started, should reach a Status value of `Up` within a
reasonable time frame.

For example, if we started the `docker.io/library/busybox:latest` container,
this is what success would look like for that container.

```shell
$ podman ps --all
CONTAINER ID  IMAGE                             COMMAND     CREATED        STATUS            PORTS       NAMES
69b0dc3c8f22  docker.io/library/busybox:latest  86400       2 seconds ago  Up 3 seconds ago              exciting_mirzakhani
```

Using `podman inspect` on the same container, we might want to look for the Status at the following path.

```shell
$ podman inspect 69b0dc3c8f22 | jq -r .[0].State.Status
running
```

Here's the full status block for context.

```json
podman inspect 69b0dc3c8f22 | jq -r .[0].State
{
  "OciVersion": "1.0.2-dev",
  "Status": "running",
  "Running": true,
  "Paused": false,
  "Restarting": false,
  "OOMKilled": false,
  "Dead": false,
  "Pid": 6561,
  "ConmonPid": 6557,
  "ExitCode": 0,
  "Error": "",
  "StartedAt": "2021-12-20T16:46:54.794870642Z",
  "FinishedAt": "0001-01-01T00:00:00Z",
  "Healthcheck": {
    "Status": "",
    "FailingStreak": 0,
    "Log": null
  }
}
```

## Implementation Notes

- The preflight codebase previously contained programmatic access to
  Podman/Docker that was deprecated for the existing use cases at the time. We
  may be able to leverage this code for these checks.
  
  - Original PR: [#50](https://github.com/redhat-openshift-ecosystem/openshift-preflight/pull/50)

- The implementation of this check re-introduces the need to run nested
  containerization (e.g. Podman inside of a container). This carries with it
  implications surrounding the need for elevated privileges within the top-level
  container engine (e.g. the ability to run as the root user in the preflight
  container.
  
  - Original issue, requesting that preflight reduce privilege requirements in order
    to execute [#165](https://github.com/redhat-openshift-ecosystem/openshift-preflight/issues/165)
  
  - Original PR, deprecating previous podman/docker dependencies to reduce
    privilege requirements
    [#174](https://github.com/redhat-openshift-ecosystem/openshift-preflight/pull/174)

## Items Pending Clarification / Questions

- Does this PR need to enable support for both Podman and Docker?
  
  - Pros:
  
    - The previous PR containing code enabling programmatic access to Podman
      already includes abstractions allowing access to Podman Docker

  - Cons:

    - Additional requirements surrounding testing of both engines, debugging
      of both engines, etc.

- Should we assume that nothing further is required _other_ than a `podman run`
  of the container?

  - Are there cases in which a container would not run if not provided parameters?
    E.g. `podman run -d --rm <image> --container-app-flag`

## Additional changes to resolve this issue

A non-exhaustive list of places that may need to be updated accompanying a pull
request that resolves this issue.

- [Developer Docs](https://github.com/redhat-openshift-ecosystem/openshift-preflight/tree/main/docs/dev#requirements).

  - Review this document to see if system dependencies have changes for developers.

- [Config Reference](https://github.com/redhat-openshift-ecosystem/openshift-preflight/blob/main/docs/CONFIG.md)

  - If any additional configuration is added, such as a user preference re:
    container tool (if supported), this document will likely need an update.

- [README.md requirements](https://github.com/redhat-openshift-ecosystem/openshift-preflight#requirements)

  - If the user needs a container engine to be locally installed, these
    requirements likely need to be updated.

samira-barouti:
* Fixes #358

### DIFF
--- a/certification/internal/policy/container/runnable_container_test.go
+++ b/certification/internal/policy/container/runnable_container_test.go
@@ -1,0 +1,43 @@
+package container
+
+import (
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/internal/cli"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification"
+)
+
+var _ = Describe("RunnableContainerCheck", func() {
+	var (
+		engine                 cli.PodmanEngine
+		runnableContainerCheck RunnableContainerCheck
+		imageRef               certification.ImageReference
+	)
+
+	BeforeEach(func() {
+		imageRef.ImageURI = "test-image:test-tag"
+	})
+
+	Describe("Checking that the container runs within a specified timeframe", func() {
+		Context("When container starts successfully", func() {
+			It("should pass Validate", func() {
+				engine = GoodPodmanEngine{}
+				runnableContainerCheck = *NewRunnableContainerCheck(&engine)
+				ok, err := runnableContainerCheck.Validate(imageRef)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(ok).To(BeTrue())
+			})
+		})
+		Context("When container does not run successfully", func() {
+			It("should not pass Validate", func() {
+				engine = BadPodmanEngine{}
+				runnableContainerCheck = *NewRunnableContainerCheck(&engine)
+				ok, err := runnableContainerCheck.Validate(imageRef)
+				Expect(err).To(HaveOccurred())
+				Expect(ok).To(BeFalse())
+			})
+		})
+	})
+
+})
--- a/certification/internal/policy/container/container_suite_test.go
+++ b/certification/internal/policy/container/container_suite_test.go
@@ -1,6 +1,7 @@
 package container
 
 import (
+	"errors"
 	"io"
 	"testing"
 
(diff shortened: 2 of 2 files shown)


certification/internal/policy/container/runnable_container_test.go
```diff
@@ -0,0 +1,43 @@
...
+		Context("When container starts successfully", func() {
+			It("should pass Validate", func() {
+				engine = GoodPodmanEngine{}
+				runnableContainerCheck = *NewRunnableContainerCheck(&engine)
+				ok, err := runnableContainerCheck.Validate(imageRef)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(ok).To(BeTrue())
+			})
+		})
+		Context("When container does not successfully", func() {
```
bcrochet, line 32: "does not start"

bcrochet (changes requested):

certification/internal/policy/container/runnable_container_test.go
samira-barouti: Whoops! Done.

bcrochet (approved):

PR #410, opened 2022-02-15
Implement RunAsSystemContainer Check

### LINKED ISSUE #374: [Container Cert] Container can run as a systemd service
https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux_atomic_host/7/html/managing_containers/running_containers_as_systemd_services_with_podman

`podman generate systemd`

samira-barouti:
This check ensures that a container can start as a Systemd service.
Ref: https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/building_running_and_managing_containers/assembly_porting-containers-to-systemd-using-podman_building-running-and-managing-containers

* Fixes #374

### DIFF
--- a/certification/internal/engine/podman.go
+++ b/certification/internal/engine/podman.go
@@ -4,11 +4,18 @@ import (
 	"bytes"
 	"context"
 	"fmt"
-	"os/exec"
+	"os"
+	exec "os/exec"
+	"regexp"
 	"strings"
 
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification"
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/internal/cli"
 	log "github.com/sirupsen/logrus"
+)
+
+const (
+	runAsPrivileged = true
 )
 
 type podmanEngine struct{}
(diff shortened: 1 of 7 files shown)


certification/internal/engine/podman.go
```diff
@@ -177,3 +149,122 @@ func (p *podmanEngine) WaitContainer(containerId string, waitOptions cli.WaitOpt
...
+			Stderr: output.Stderr,
+		}, err
+	}
+	if output, err = RunCommand("systemctl", []string{"start", serviceName}); err != nil {
+		log.Error(fmt.Sprintf("unable to start the service %s: ", serviceName), err)
+		log.Error("Stderr: ", output.Stderr)
+		return &cli.PodmanOutput{
+			Stderr: output.Stderr,
+		}, err
+	}
```
samira-barouti, line 201: @bcrochet @acornett21 @tonytcampbell @jomkz Since this check needs to copy unit file into systemd dir, and start/stop services, the preflight binary needs to run as root. Is it ok?

bcrochet: I think I'd rather us call sudo, rather than expect the binary to run as root. Then, it is up to the caller of preflight to ensure that sudo is set up correctly, but it also gives them the opportunity to limit what preflight has access to.

acornett21: I hadn't thought of @bcrochet 's approach, but that makes sense, and puts us more in control of when to execute `sudo` and when not to.

certification/internal/engine/podman.go
```diff
@@ -4,66 +4,39 @@ import (
...
 	"os/exec"
+	"regexp"
 	"strings"
 
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/internal/cli"
 	log "github.com/sirupsen/logrus"
 )
 
+const (
+	syetemdDir = "/etc/systemd/system"
```
acornett21, line 18: I think this is the `var` you were asking about, once my PR lands, you can place this in `defaults.go`

samira-barouti: Awesome! Thanks for making our life easier

certification/internal/engine/podman.go
samira-barouti: Great point! I updated the PR to only run with `sudo` if it is absolutely necessary. Please let me know if you have any feedbacks.

certification/internal/engine/podman.go
```diff
@@ -177,3 +149,122 @@ func (p *podmanEngine) WaitContainer(containerId string, waitOptions cli.WaitOpt
...
+	if err != nil {
+		log.Error("unable to fetch the name of the systemd service: ", err)
+		return &cli.PodmanOutput{
+			Stderr: err.Error(),
+		}, err
+	}
+	serviceName := r.FindStringSubmatch(serviceFilepath)[1]
+
+	// copy the unit file from working dir to systemd dir
+	_, err = copyFile(serviceFilepath, fmt.Sprintf("%s/%s", syetemdDir, serviceName))
```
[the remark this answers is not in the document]
samira-barouti: Done.

certification/internal/engine/podman.go
```diff
@@ -177,3 +148,127 @@ func (p *podmanEngine) WaitContainer(containerId string, waitOptions cli.WaitOpt
 	log.Info("container reached a running state")
 	return true, nil
 }
+
+func (p *podmanEngine) RunSystemContainer(containerName string) (*cli.PodmanOutput, error) {
+	log.Debug(fmt.Sprintf("Generating Systemd file for container %s", containerName))
+	cmdArgs := []string{"generate", "systemd", "--new", "--files", "--name", containerName}
+
+	output, err := RunCommand(false, "podman", cmdArgs)
```
bcrochet, line 156: You have a "magic boolean" here. One has to go and read the RunCommand function to see what that boolean means. It wouldn't be a bad idea to make a 'var privileged bool' at the top of this function, use it throughout, and just change it before each call.

bcrochet (commented):
Just a question about the privileged boolean.

certification/internal/engine/podman.go
samira-barouti: Sounds like a great idea! Will update the PR shortly :)

bcrochet (approved):
/lgtm

PR #412, opened 2022-02-15
Changes uses of make to be more idiomatic Go

bcrochet:
make can accept a second integer parameter that specifies the capacity
when making a slice. When used in this fashion, a slice can have the
benefits of being preallocated with capacity (not requiring to
continually be extended), but a 0 length (when passed 0 as the first).
Then, one can use append and not need the extra index var.


acornett21 (approved):
Thanks for making this change! Code is much more readable!

samira-barouti (approved):
/lgtm

PR #417, opened 2022-02-16
Move the defer statement to after error checks

bcrochet:
The defer statements are not necessary if an error occurs in the
os.Open call. Therefore, they should come after the error check.


acornett21 (approved):

samira-barouti (approved):
/lgtm

PR #421, opened 2022-02-17
Use the submit variable where possible

bcrochet:
One spot still used viper directly, instead of the submit variable
that was introduced in another PR.


acornett21 (approved):

samira-barouti (approved):
/lgtm

Issue #403, opened 2022-02-10
Clearly indicate the overall check results.


tonytcampbell:
## Is your feature request related to a problem? Please describe.
The JSON output of Preflight contains the pass/fail results for all checks and an overall `passed` attribute that indicates if all the checks passed or not. 

## Describe the solution you'd like.
The overall attribute named `passed` is not clear enough.  We should rename it to `overall result` or something similar. 

## Describe alternatives you've considered.
(A clear and concise description of any alternative solutions or features you've considered.)

## Additional context.
(Add any other context or screenshots about the feature request here.)

acornett21:
For `preflight` to implement this, we would need supporting changes in pyxis as well, since what `preflight` displays to the user is the payload that we send to `pyxis`.

jomkz:
After discussing this a bit, we will log a message to `stderr` that indicates the overall result to make it more clear to the end user.

PR #422, opened 2022-02-17
adding a temporary log and output message 

acornett21:
- adding temp logging and output message so we the user knows we actually submitted something

**note: ones pyxis api contract is finished we will display actually info to the user**

### DIFF
--- a/cmd/check_container.go
+++ b/cmd/check_container.go
@@ -197,6 +197,11 @@ var checkContainerCmd = &cobra.Command{
 			if err != nil {
 				return err
 			}
+
+			log.Info("results successfully submitted to Red Hat")
+			//TODO: displaying this to the user for now so they know submission has be complete
+			// once we get new API contract with actually info to give to the user we can update this
+			fmt.Println("results successfully submitted to Red Hat")
 		}
 
 		return nil


bcrochet:
Is this meant to fix #403 ?

cmd/check_container.go
```diff
@@ -197,6 +197,11 @@ var checkContainerCmd = &cobra.Command{
 			if err != nil {
 				return err
 			}
+
+			log.Info("results successfully submitted to Red Hat")
+			//TODO: displaying this to the user for now so they know submission has be complete
+			// once we get new API contract with actually info to give to the user we can update this
+			fmt.Println("results successfully submitted to Red Hat")
```
bcrochet, line 204: I'm not sure how I feel about this. We are outputting the test results to stdout as json. If we mix in other output, then that negates the ability to use a pipe to interpret said results. You would then have malformed json. This should either be a log message that always prints in our default loglevel, or it needs to just go to stderr.

bcrochet (changes requested):

acornett21:
> Is this meant to fix #403 ?

No. This was just meant as a placeholder to indicate we've submitted results. We don't have the `official` verbiage to display to the user. But maybe after out conversation this morning, we just write out the `_ID` to the user and refine later after a demo? Thoughts?

cmd/check_container.go
acornett21: This was just going to be temporary, but after this morning, we know what to display to the user.

bcrochet: So, we definitely do not want to print out to stdout. That should be reserved for the result json. Any output should come in the form of a log.Info. If our default log level is not log.Info, it should be made as such, and the log.Info log level should be reserved for actual messages to the user. It should not be used for error reporting, or debug-level messages.

PR #423, opened 2022-02-17
Add image_id to the test results

bcrochet:
When we create the image, pull out the image id and add it to the
test results when submitted.

### DIFF
--- a/certification/pyxis/pyxis.go
+++ b/certification/pyxis/pyxis.go
@@ -80,27 +80,23 @@ func (p *pyxisEngine) createImage(ctx context.Context, certImage *CertImage) (*C
 	return &newCertImage, nil
 }
 
-func (p *pyxisEngine) createRPMManifest(ctx context.Context, imageId string, rpms []RPM) error {
-	rpmManifest := RPMManifest{
-		ImageID: imageId,
-		RPMS:    rpms,
-	}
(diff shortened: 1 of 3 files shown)


bcrochet:
This needs more investigation.

/hold

certification/pyxis/pyxis.go
```diff
@@ -76,17 +76,13 @@ func (p *pyxisEngine) createImage(ctx context.Context, certImage *CertImage) (*C
...
-		RPMS:    rpms,
-	}
+func (p *pyxisEngine) createRPMManifest(ctx context.Context, rpmManifest *RPMManifest) error {
 	b, err := json.Marshal(rpmManifest)
 	if err != nil {
 		log.Error(err)
 		return err
 	}
-	req, err := newRequestWithApiToken(ctx, http.MethodPost, getPyxisUrl(fmt.Sprintf("images/id/%s/rpm-manifest", imageId)), p.ApiToken)
+	req, err := newRequestWithApiToken(ctx, http.MethodPost, getPyxisUrl(fmt.Sprintf("images/id/%s/rpm-manifest", rpmManifest.ImageID)), p.ApiToken)
```
acornett21, line 85: When we write rpm manifest to disk we do not have the image id, so we pass it somewhere.

acornett21 (approved):

jomkz (approved):

PR #427, opened 2022-02-21
Simplify the handling of the API token

bcrochet:
Pull the newRequestWithApiToken into the pyxisEngine struct. So,
rather than pass the apitoken to every call, we instead pull it
out of the struct, and take the body. If the body is not nil, we
add the application/json Content-type header.

### DIFF
--- a/certification/pyxis/pyxis.go
+++ b/certification/pyxis/pyxis.go
@@ -45,12 +45,11 @@ func (p *pyxisEngine) createImage(ctx context.Context, certImage *CertImage) (*C
 		log.Error(err)
 		return nil, err
 	}
-	req, err := newRequestWithApiToken(ctx, http.MethodPost, getPyxisUrl("images"), p.ApiToken)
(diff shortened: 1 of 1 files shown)


certification/pyxis/pyxis.go
```diff
@@ -122,7 +120,7 @@ func (p *pyxisEngine) createRPMManifest(ctx context.Context, rpmManifest *RPMMan
 }
 
 func (p *pyxisEngine) GetProject(ctx context.Context) (*CertProject, error) {
-	req, err := newRequestWithApiToken(ctx, http.MethodGet, getPyxisUrl(fmt.Sprintf("projects/certification/id/%s", p.ProjectId)), p.ApiToken)
+	req, err := p.newRequestWithApiToken(ctx, http.MethodGet, getPyxisUrl(fmt.Sprintf("projects/certification/id/%s", p.ProjectId)), nil)
```
acornett21, line 123: Is this `nil` by design?

bcrochet: Yes. It signifies that there is no body.

acornett21 (approved):

jomkz (approved):

PR #430, opened 2022-02-22
Add layer information to the results

bcrochet:
Get some of the layer ids and other information from Crane and pass
it on to Pyxis results.

### DIFF
--- a/certification/internal/engine/engine.go
+++ b/certification/internal/engine/engine.go
@@ -24,7 +24,6 @@ import (
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/errors"
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/pyxis"
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/runtime"
-	preflightRuntime "github.com/redhat-openshift-ecosystem/openshift-preflight/certification/runtime"
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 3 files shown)


certification/internal/engine/engine.go
```diff
@@ -342,22 +357,23 @@ func writeCertImage(img v1.Image) error {
...
+			Size:                   size,
+			UncompressedLayerSizes: layerSizes,
 		},
 		RawConfig:         string(rawConfig),
 		SumLayerSizeBytes: sumLayersSizeBytes,
-		// TODO: figure out how to get/convert this data from crane
-		UncompressedTopLayerId: "placeholder",
+		// This is an assumption that the DiffIDs are in order from base up.
+		// Need more evisdence that this is always the case.
+		UncompressedTopLayerId: config.RootFS.DiffIDs[0].String(),
```
acornett21, line 376: Non Blocking: If this is `top` level would this be the _last_ element in the array ie `config.RootFS.DiffIDs[len(config.RootFS.DiffIDs)-1]`?

acornett21 (approved):

certification/internal/engine/engine.go
bcrochet: I thought so, but in my findings it's the first in the list.

acornett21: Okay, that's kind of confusing for sure. Happy to test more on this when the time comes.

PR #431, opened 2022-02-22
Add a doc describing how to find uncompressed layer IDs

bcrochet:

### DIFF
--- a/docs/dev/LAYER_IDS.md
+++ b/docs/dev/LAYER_IDS.md
@@ -1,0 +1,29 @@
+# Using crane to find the uncompressed layer IDs
+
+`podman inspect` does not give the same layer IDs as `crane` does. `crane` should
+be used if one is looking for the uncompressed layer IDs.
+
+```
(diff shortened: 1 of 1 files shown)


docs/dev/LAYER_IDS.md
```diff
@@ -0,0 +1,29 @@
+# Using crane to find the uncompressed layer IDs
+
+`podman inspect` does not give the same layer IDs as `crane` does. `crane` should
+be used if one is looking for the uncompressed layer IDs.
+
+```
+❯ crane config registry.redhat.io/ubi8/ubi:latest | jq '.rootfs.diff_ids'
```
acornett21, line 7: We also need a `size` value of these layers. Do we then take those `sha` values and look them up in the `Layer` array?

acornett21: This was implemented differently than I thought, but how it's implemented makes sense.

acornett21 (approved):

PR #432, opened 2022-02-22
displaying response information to the user after results have been submitted

acornett21:
- fixes: #376 
- displaying the image id to the user and then showing them a URL to the project overview.

### DIFF
--- a/cmd/check_container.go
+++ b/cmd/check_container.go
@@ -191,11 +191,15 @@ var checkContainerCmd = &cobra.Command{
 				return err
 			}
 
-			// TODO: use the return values once we know what we need to display to the user
-			_, _, _, err = pyxisEngine.SubmitResults(certProject, certImage, rpmManifest, testResults)
+			_, certImage, _, err = pyxisEngine.SubmitResults(certProject, certImage, rpmManifest, testResults)
 			if err != nil {
 				return err
 			}
+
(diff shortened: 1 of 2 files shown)


acornett21:
@tonytcampbell Is along the lines of the message that we want to show to the user?

cmd/check_container.go
```diff
@@ -191,11 +191,15 @@ var checkContainerCmd = &cobra.Command{
...
 			}
 
-			// TODO: use the return values once we know what we need to display to the user
-			_, _, _, err = pyxisEngine.SubmitResults(certProject, certImage, rpmManifest, testResults)
+			_, certImage, _, err = pyxisEngine.SubmitResults(certProject, certImage, rpmManifest, testResults)
 			if err != nil {
 				return err
 			}
+
+			log.Infof("Test results have been submitted to Red Hat."+
```
bcrochet, line 199: These might be better as individual log statements?

acornett21: I was thinking that as well, or was wondering if there was a way to formate like a `paragraph` for a log message? But I don't think that's a thing, even `\n` shows up as a literal.

jomkz (approved):
/lgtm

bcrochet (approved):
/lgtm

PR #434, opened 2022-02-23
updating gitignore to include test file we create as part of runnable tests

acornett21:
- updating the gitignore to include files we create during runnable tests.
- we remove this file after the test is over, but during development/testing sometimes this file stays around and its better that it doesn't get committed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@ vuln.html
 bin
 preflight
 openshift-preflight
+container-podman-test.service
 
 # Config files
 config.yaml


bcrochet (changes requested):
I'd prefer to see us write this to a tmp dir rather than ignore it here.

acornett21:
> I'd prefer to see us write this to a tmp dir rather than ignore it here.

This is a file that `podman` writes and I don't see an argument to change where it's written to, but can look more later.

bcrochet:
> > I'd prefer to see us write this to a tmp dir rather than ignore it here.
> 
> This is a file that `podman` writes and I don't see an argument to change where it's written to, but can look more later.

You would need to not use the '--files' and read/write the file from stdout.

bcrochet:
Ignoring this for now, since these checks are no longer enabled.

Issue #436, opened 2022-02-24
The Build Release Action Does Not Run for Pre-release Tags


jomkz:
## Bug Description

When attempting to create a pre-release for preflight by creating the `1.1.0-alpha1` tag through the GitHub release UI, the "build release" action that is responsible for creating the container images was not run.  

## Version and Command Invocation

1.1.0-alpha1

## Steps to Reproduce:

1) Create a new pre-release using the GitHub UI for a tag other than `n.n.n`.

## Expected Result

Expecting the container images to be generated and pushed to quay.io for alpha (N.N.N-alphaN) or beta (N.N.N-betaN) tags on the repository, similar to what happens for production release tags (N.N.N)

## Actual Result

The container images were not created because the "build release" action was not run.

## Additional Context

The regex for matching tags does not account for release tags other than `n.n.n`.

PR #437, opened 2022-02-24
update gh release action so preflight can support alpha, beta and rc releases

acornett21:
- Fixes: #436 
- Updating the gh release action to support alpha, beta and rc using || conditions since gh actions do not lookahead/behind/grouping/etc for the regex 


bcrochet (approved):
/lgtm

jomkz (approved):

PR #444, opened 2022-02-28
Disable runnable checks for now

bcrochet:
Preflight will not, at this time, be running the runnable container
check and the run as system container check. So, this disables them
until we know for sure they will not be used, then we can remove the
code completely.


jomkz (approved):
/lgtm

Issue #442, opened 2022-02-25
currently if pyxis returns a error for one api call, preflight cannot run `--submit` again


acornett21:
## Bug Description
When running preflight if the image or rpm-mainfest have been created, but say the `/results` api errors out, preflight cannot execute `--submit` again since we are calling POST endpoints.

## Version and Command Invocation
main

## Steps to Reproduce:
1) run `--submit` with bad data field for the results 
2) see `/results call fail`
3) re run `-submit` see `create image` call fail due to `409` conflict

## Expected Result
I'd expect preflight to handle response from pyxis properly and be able to continue the process

## Actual Result
preflight can't get passed the `/image` POST call

## Additional Context
Any `POST` call that we make, we should handle the `409` response make  a `PATCH` to pyxis

PR #448, opened 2022-03-01
adding in a call to pyxis getImage so preflight can run multiple times

acornett21:
- Fixes: #442 
- adding in call to `projects/certification/id/%s/images?filter=docker_image_digest==%s` to see if the image already exists in pyxis
- Relates: #369 
- added call to `images/id/%s/rpm-manifest` to get an existing rpm manifest (**note: this route isn't fully functioning in the external pyxis api yet**)
- adding in tests to cover all new and existing methods in pyxis package

### DIFF
--- a/certification/pyxis/pyxis.go
+++ b/certification/pyxis/pyxis.go
@@ -51,6 +51,8 @@ func (p *pyxisEngine) createImage(ctx context.Context, certImage *CertImage) (*C
 		return nil, err
 	}
 
+	log.Debugf("URL is: %s", req.URL)
+
 	resp, err := p.Client.Do(req)
 	if err != nil {
 		log.Error(err)
@@ -63,6 +65,10 @@ func (p *pyxisEngine) createImage(ctx context.Context, certImage *CertImage) (*C
 	if err != nil {
 		log.Error(err)
 		return nil, err
+	}
+
+	if resp.StatusCode == 409 {
+		return nil, errors.Err409StatusCode
 	}
 
 	if !checkStatus(resp.StatusCode) {
@@ -79,17 +85,15 @@ func (p *pyxisEngine) createImage(ctx context.Context, certImage *CertImage) (*C
 	return &newCertImage, nil
 }
 
-func (p *pyxisEngine) createRPMManifest(ctx context.Context, rpmManifest *RPMManifest) (*RPMManifest, error) {
-	b, err := json.Marshal(rpmManifest)
+func (p *pyxisEngine) getImage(ctx context.Context, dockerImageDigest string) (*CertImage, error) {
+	req, err := p.newRequestWithApiToken(ctx, http.MethodGet,
+		getPyxisUrl(fmt.Sprintf("projects/certification/id/%s/images?filter=docker_image_digest==%s", p.ProjectId, dockerImageDigest)), nil)
 	if err != nil {
 		log.Error(err)
 		return nil, err
 	}
-	req, err := p.newRequestWithApiToken(ctx, http.MethodPost, getPyxisUrl(fmt.Sprintf("images/id/%s/rpm-manifest", rpmManifest.ImageID)), bytes.NewReader(b))
-	if err != nil {
-		log.Error(err)
-		return nil, err
-	}
+
+	log.Debugf("URL is: %s", req.URL)
 
 	resp, err := p.Client.Do(req)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


certification/pyxis/pyxis.go
```diff
@@ -79,6 +84,45 @@ func (p *pyxisEngine) createImage(ctx context.Context, certImage *CertImage) (*C
...
+	defer resp.Body.Close()
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		log.Error(err)
+		return nil, err
+	}
+
+	if !checkStatus(resp.StatusCode) {
+		log.Errorf("%s: %s", "received non 200 status code in createImage", string(body))
```
bcrochet, line 110: Copy-pasta

certification/pyxis/pyxis.go
```diff
@@ -79,6 +84,45 @@ func (p *pyxisEngine) createImage(ctx context.Context, certImage *CertImage) (*C
...
+		return nil, errors.ErrNon200StatusCode
+	}
+
+	// using gjson to get the certImage we request since the payload is wrapped in an additional `data` element
+	certImage := gjson.Get(string(body), "data.0")
```
bcrochet, line 115: Still unsure about introducing a new library for this.

certification/pyxis/submit.go
```diff
@@ -23,17 +25,23 @@ func (p *pyxisEngine) SubmitResults(certProject *CertProject, certImage *CertIma
...
 	}
 
+	dockerImageDigest := certImage.DockerImageDigest
+
 	certImage, err = p.createImage(ctx, certImage)
-	if err != nil {
+	if err != nil && err != errors.Err409StatusCode {
 		log.Error(err, "could not create image")
 		return nil, nil, nil, err
+	} else {
```
bcrochet, line 34: There is no need for the else is there?

certification/pyxis/types.go
```diff
@@ -18,6 +19,16 @@ type CertImage struct {
 	UncompressedTopLayerId string       `json:"uncompressed_top_layer_id,omitempty"` // TODO: figure out how to populate this, it is not required
 }
 
+type Links struct {
+	RpmManifest     Href `json:"rpm_manifest,omitempty"`
+	TestResults     Href `json:"test_results"`
```
bcrochet, line 24: Shouldn't these all be omitempty? We would not want to submit these ever.

certification/pyxis/types.go
```diff
@@ -4,6 +4,7 @@ import "github.com/redhat-openshift-ecosystem/openshift-preflight/certification/
 
 type CertImage struct {
 	ID                     string       `json:"_id,omitempty"`
+	Links                  Links        `json:"_links"`
```
bcrochet, line 7: This should be omitempty.

certification/pyxis/types.go
```diff
@@ -18,6 +19,16 @@ type CertImage struct {
...
 }
 
+type Links struct {
+	RpmManifest     Href `json:"rpm_manifest,omitempty"`
+	TestResults     Href `json:"test_results"`
+	Vulnerabilities Href `json:"vulnerabilities"`
+}
+
+type Href struct {
+	Href string `json:"href"`
```
bcrochet, line 29: omitempty?

bcrochet (changes requested):
Couple of changes

certification/pyxis/submit.go
bcrochet: Or rather... Perhaps you should check if the error IS a 409, and do the thing, otherwise, return.

certification/pyxis/types.go
acornett21: I'm going to remove this struct, it does not buy us anything. I thought that these values were populated _once_ say an `rpm manifest` or `results` were created, but that is not the case, these values are populated instantly even if `rpm manifest` or `results` haven't been created yet.

certification/pyxis/types.go
acornett21: same as above.

certification/pyxis/types.go
acornett21: same as above.

certification/pyxis/submit.go
acornett21: went with two `if's` approach, lmk what you think.

certification/pyxis/types.go
acornett21: no longer going to use these structs.

certification/pyxis/types.go
acornett21: no longer going to use this struct.

certification/pyxis/submit.go
```diff
@@ -23,17 +23,30 @@ func (p *pyxisEngine) SubmitResults(certProject *CertProject, certImage *CertIma
...
 	}
 
-	certImage, err = p.createImage(ctx, certImage)
+	dockerImageDigest := certImage.DockerImageDigest
+
+	certImage, err = p.getImage(ctx, dockerImageDigest)
 	if err != nil {
-		log.Error(err, "could not create image")
-		return nil, nil, nil, err
+		log.Warn(err, "could not create get image")
```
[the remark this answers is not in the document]
acornett21: fixed

certification/pyxis/pyxis.go
acornett21: I added a local struct and removed that library. LMK what you think.

jomkz (approved):
/lgtm

PR #450, opened 2022-03-01
removing --submit check when writing image config and rpm manifest to disk

acornett21:
- removing check for `--submit` flag when writing image config and rpm manifest to disk
- this information is helpful even if the user isn't doing a `--submit` so always write it to disk


bcrochet (approved):
/lgtm

jomkz (approved):

PR #451, opened 2022-03-01
Add a flag to specify an image as a scratch image

bcrochet:
We have some checks and auxillary methods that will fail if an image is a scratch image. Rather than reuse (and probably overload) the IsBundle flag, this adds a new IsScratch flag. And in the case of an operator check, both will be set to true.
    
### DIFF
--- a/certification/internal/engine/engine.go
+++ b/certification/internal/engine/engine.go
@@ -41,6 +41,9 @@ type CraneEngine struct {
 	// IsBundle is an indicator that the asset is a bundle.
 	IsBundle bool
 
+	// IsScratch is an indicator that the asset is a scratch image
+	IsScratch bool
+
 	imageRef certification.ImageReference
 	results  runtime.Results
 }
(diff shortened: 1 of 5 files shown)


certification/internal/engine/engine.go
```diff
@@ -96,12 +99,12 @@ func (c *CraneEngine) ExecuteChecks() error {
 		return fmt.Errorf("%w: %s", errors.ErrExtractingTarball, err)
 	}
 
-	// only write these files to disk for container checks
-	if !c.IsBundle {
-		if err := writeCertImage(img); err != nil {
-			return err
-		}
+	// write json files to disk to be used to send to pyxis after all checks are complete
```
acornett21, line 102: this comment is probably no longer valid, since we are going to write this file for bundles and non `container --submit` cases as well.

acornett21 (approved):

jomkz (approved):

PR #457, opened 2022-03-04
remove `options` from `crane.ListTags`

acornett21:
- removing `options` from `crane.ListTags` since crane ends up using options.auth in their implementation and then they point this back to the local `auth.json` file or more simply pout the `DefaultKeychain` on the system.

### DIFF
--- a/certification/internal/engine/crane.go
+++ b/certification/internal/engine/crane.go
@@ -1,6 +1,8 @@
 package engine
 
-import "github.com/google/go-containerregistry/pkg/crane"
+import (
+	"github.com/google/go-containerregistry/pkg/crane"
+)
 
 type craneEngine struct{}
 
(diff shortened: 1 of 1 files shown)


skattoju (approved):
lgtm

certification/internal/engine/crane.go
```diff
@@ -11,6 +14,7 @@ func NewCraneEngine() *craneEngine {
 func (c *craneEngine) ListTags(imageURI string) ([]string, error) {
 	// prepare crane runtime options
 	options := make([]crane.Option, 0)
```
bcrochet, line 16: Small nit, but would be more efficient to give it a length 0, capacity of 1.

acornett21: Yeah, I can change that. Was more focused on trying to test this.

tonytcampbell (approved):
/lgtm

certification/internal/engine/crane.go
```diff
@@ -10,7 +13,8 @@ func NewCraneEngine() *craneEngine {
 
 func (c *craneEngine) ListTags(imageURI string) ([]string, error) {
 	// prepare crane runtime options
-	options := make([]crane.Option, 0)
+	options := make([]crane.Option, 1)
```
bcrochet, line 16: This needs to be make([]crane.Option, 0, 1)

jomkz (approved):
/lgtm

PR #458, opened 2022-03-07
match catalog url and connect url

acornett21:
- match catalog url and connect url so we can support running preflight in lower environments
- this gives the user the proper connect URL for the pyxis environment that they connected to

### DIFF
--- a/cmd/check_container.go
+++ b/cmd/check_container.go
@@ -202,7 +202,7 @@ var checkContainerCmd = &cobra.Command{
 			log.Info("Test results have been submitted to Red Hat.")
 			log.Info("These results will be reviewed by Red Hat for final certification.")
 			log.Infof("The container's image id is: %s.", certImage.ImageID)
-			log.Infof("Please check https://connect.redhat.com/projects/%s/overview to monitior the progress.", projectId)
+			log.Infof(fmt.Sprintf("Please check %s to monitior the progress.", buildConnectURL(projectId)))
 		}
 
 		return nil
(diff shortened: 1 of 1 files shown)


cmd/check_container.go
```diff
@@ -229,3 +229,16 @@ func init() {
 
 	checkCmd.AddCommand(checkContainerCmd)
 }
+
+func buildConnectURL(projectID string) string {
+	connectURL := fmt.Sprintf("https://connect.redhat.com/projects/%s/overview", projectID)
```
itroyano, line 234: maybe it's worth moving urls and the like into a constants.go ?
would be easier to maintain later.

acornett21: This URL is only for display purposes to the user, we do not call this URL in the code. The `catalog` URL is in `defaults`, I'm not sure we need to add `connect` there as well.

cmd/check_container.go
```diff
@@ -229,3 +229,16 @@ func init() {
...
+	pyxisHost := viper.GetString("pyxis_host")
+
+	if len(pyxisHost) > 0 {
+		s := strings.Split(pyxisHost, ".")
```
bcrochet, line 238: I don't think this will work in all cases. pyxisHost should never be 0 length, I think. It has a default set. So this will always happen. If you didn't set PYXIS_HOST, the s[1] below would be 'redhat'. We might instead need to have a separate 'PFLT_PYXIS_ENV' or something.

bcrochet (changes requested):

jomkz (approved):
/lgtm

bcrochet (approved):
/lgtm

PR #459, opened 2022-03-07
adding in repositories information for create image pyxis call

acornett21:
- Relates: #369 
- Adding in repositories information to be sent to pyxis  
- Moved `writeCertImage` and `writeRPMManifest` to be after `certification.ImageReference` struct creation since `writeCertImage` will now use `certification.ImageReference` struct instead of `crane v1.Image` struct so that we can call `name.NewTag` with the proper information.

### DIFF
--- a/certification/internal/engine/engine.go
+++ b/certification/internal/engine/engine.go
@@ -18,7 +18,7 @@ import (
 	"time"
 
 	"github.com/google/go-containerregistry/pkg/crane"
-	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/name"
 	rpmdb "github.com/knqyf263/go-rpmdb/pkg"
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification"
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/artifacts"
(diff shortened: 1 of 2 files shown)


certification/internal/engine/engine.go
```diff
@@ -51,12 +52,21 @@ type CraneEngine struct {
 func (c *CraneEngine) ExecuteChecks() error {
 	log.Debug("target image: ", c.Image)
 
-	// prepare crane runtime options, if necessary
-	options := make([]crane.Option, 0)
+	// prepare crane runtime options for NewTag call, if necessary
+	tagOptions := make([]name.Option, 0)
+
+	log.Debug("getting tag info for image")
+	tag, err := name.NewTag(c.Image, tagOptions...)
```
bcrochet, line 59: If you don't ever anticipate us putting options there, just leave it off. I don't think we would need options here.

bcrochet (commented):
Just a small comment about the options. Feel free to ignore it.

certification/internal/engine/engine.go
```diff
@@ -100,7 +107,7 @@ func (c *CraneEngine) ExecuteChecks() error {
 		return fmt.Errorf("%w: %s", errors.ErrExtractingTarball, err)
 	}
 
-	if err := writeCertImage(img); err != nil {
+	if err := writeCertImage(img, tag); err != nil {
```
bcrochet, line 110: rather than change this signature, why not just create the tag in the writeCertImage func? It makes more sense there, since we aren't using the tag anywhere else in this func.

certification/internal/engine/engine.go
```diff
@@ -100,7 +100,15 @@ func (c *CraneEngine) ExecuteChecks() error {
 		return fmt.Errorf("%w: %s", errors.ErrExtractingTarball, err)
 	}
 
-	if err := writeCertImage(img); err != nil {
+	// store the image internals in the engine image reference to pass to validations.
+	// TODO: pass this to validations when the new check interface includes it.
```
bcrochet, line 104: This TODO can be removed

bcrochet (changes requested):
Just a small comment to remove a comment.

certification/internal/engine/engine.go
acornett21: I've been wondering that for awhile and keep forgetting to ask, will remove it.

bcrochet (approved):
/lgtm

bcrochet (approved):
/lgtm

Issue #461, opened 2022-03-08
Have `certImage.certified` field match `testResults.passed` Field


acornett21:
## Bug Description
Currently `preflight` sends a default value of `false` to pyxis when posting to `/images`, this was based on the assumption that the certification process still needs to run some backend tasks for the certification process to happen, however those process already are accounted for and set their own `true/false` value.

## Version and Command Invocation
`main`

## Steps to Reproduce:
1) Run `check container`
2) See that `certImage.certifed` is always set to false

## Expected Result
The value sent to pyxis should match the value from the results.

## Actual Result
The value is always false.

## Additional Context
N/A

PR #462, opened 2022-03-08
making certImage.certified match testResults.passed

acornett21:
- Fixes: #461 
- Moved`testResultsJsonFile` to be the first step since `testResults.passed` is needed in `certImage` creation. The diff makes it looked like more has changed, but it really is that test `testResultsJsonFile` moves from `last` to `first` in this code.


bcrochet (approved):
/lgtm

jomkz (approved):

PR #466, opened 2022-03-09
Pass docker config to the results upload

bcrochet:
The docker config was not being passed to the results. This adss them
to the results, and makes the env var the same as the check operator
command.

This also rearranges some of the initialization code so that the viper
configs are initialized in time to be able to shortcut whether
--docker-config is required or not. If the PFLT_DOCKERCONFIG envvar
is set, it does not require the flag.

### DIFF
--- a/cmd/check_container.go
+++ b/cmd/check_container.go
@@ -44,7 +44,7 @@ var checkContainerCmd = &cobra.Command{
 			return fmt.Errorf("%w: A container image positional argument is required", errors.ErrInsufficientPosArguments)
 		}
 
-		if submit {
+		if submit && !viper.IsSet("dockerConfig") {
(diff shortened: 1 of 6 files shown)


cmd/check_container.go
```diff
@@ -88,6 +89,21 @@ var checkContainerCmd = &cobra.Command{
 			cfg.Scratch = true
 		}
 
+		if submit {
```
acornett21, line 92: why another `if submit {` when we have that on line 157? This diff might not be showing the entire picture, or I'm missing something.

acornett21: This was done so this logic lives closer to the GetProject method call.

acornett21 (approved):

tonytcampbell (approved):
/lgtm

Issue #465, opened 2022-03-09
Preflight must output the created containerImage _id to the user


jpaulovic:
In order to enable partners in UAT to view test results in Connect, the partners will need to be able to construct a URL manually to get to specific image's test results in this format: `https://connect.uat.redhat.com/projects/{certProject._id}/images/{containerImage._id}/scan-results`

Preflight currently outputs a log such as
`time="2022-03-04T08:37:22Z" level=info msg="Please check https://connect.redhat.com/projects/6217d5887c6c194c4959ae10/overview to monitor the progress."`

We need to extend that to also output the containerImage._id so that the users in UAT can generate the URL mentioned above

cc @tonytcampbell

acornett21:
We will just add this to the existing logic and always displaying `/scan-results` for all environments, we also need to change the below:

`log.Infof("The container's image id is: %s.", certImage.ImageID)` to

`log.Infof("The container's image id is: %s.", certImage.ID)

PR #469, opened 2022-03-09
adding in scan results url to output and changing to the mongo db _id

acornett21:
- Fixes: #465 
- Displaying scan-results URL to the user
- Changing from displaying `image_id` to `_Id` to the user

### DIFF
--- a/cmd/check_container.go
+++ b/cmd/check_container.go
@@ -201,8 +201,9 @@ var checkContainerCmd = &cobra.Command{
 
 			log.Info("Test results have been submitted to Red Hat.")
 			log.Info("These results will be reviewed by Red Hat for final certification.")
-			log.Infof("The container's image id is: %s.", certImage.ImageID)
-			log.Infof(fmt.Sprintf("Please check %s to monitior the progress.", buildConnectURL(projectId)))
+			log.Infof("The container's image id is: %s.", certImage.ID)
+			log.Infof("Please check %s to view scan results.", buildScanResultsURL(projectId, certImage.ID))
+			log.Infof(fmt.Sprintf("Please check %s to monitor the progress.", buildOverviewURL(projectId)))
 		}
 
 		return nil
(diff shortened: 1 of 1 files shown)


cmd/check_container.go
```diff
@@ -201,8 +201,9 @@ var checkContainerCmd = &cobra.Command{
 
 			log.Info("Test results have been submitted to Red Hat.")
 			log.Info("These results will be reviewed by Red Hat for final certification.")
-			log.Infof("The container's image id is: %s.", certImage.ImageID)
-			log.Infof(fmt.Sprintf("Please check %s to monitior the progress.", buildConnectURL(projectId)))
+			log.Infof("The container's image id is: %s.", certImage.ID)
+			log.Infof("Please check %s to view scan results.", buildScanResultsURL(projectId, certImage.ID))
+			log.Infof(fmt.Sprintf("Please check %s to monitior the progress.", buildOverviewURL(projectId)))
```
itroyano, line 206: nit typo - "monitior" should be "monitor"

acornett21: Good catch!

jomkz (approved):
/lgtm

cmd/check_container.go
```diff
@@ -242,3 +243,16 @@ func buildConnectURL(projectID string) string {
 
 	return connectURL
 }
+
+func buildScanResultsURL(projectID string, imageID string) string {
+	connectURL := fmt.Sprintf("https://connect.redhat.com/projects/%s/images/%s/scan-results", projectID, imageID)
```
bcrochet, line 248: You could factor out a tunction that builds the "connect" URL... And just tack on the extra bit at the end for each.

bcrochet (changes requested):
Dedupe some code

cmd/check_container.go
bcrochet: To be addressed in another PR.

/lgtm

bcrochet (approved):

Issue #447, opened 2022-03-01
`preflight check container` Should Upload `preflight.log` to pyxis


acornett21:
## Is your feature request related to a problem? Please describe.
prflight is currently only submitting results to `/test-results` and does not give a full picture of everything that has happened during a given run. In order to provide more information to the user in connect, as well as more information to support if a support ticket is placed, preflight should also upload `preflight.log` to pyxis.

## Describe the solution you'd like.
preflight should take the `preflight.log` file that is on disk and send it to [pyxis](https://catalog.qa.redhat.com/api/containers/v1/ui/#/Certification%20projects/pyxis.rest.legacy.cert_project_artifacts.create_certification_artifact) we also need to make sure we include `image_id` from `certimage.image_id` so pyxis can link the image to the artifacts.

## Describe alternatives you've considered.
N/A

## Additional context.
N/A

PR #473, opened 2022-03-10
adding in createArtifacts method so we can send pyxis the preflight.log

acornett21:
- Fixes: #447 
- Added `POST` call to `/artifacts`
- Added code to read log file off of disk and build `Artifact` struct
- **Note** Existing tests already cover this method since the URL path is shared. Tests of pyxis code will be enhanced when we we switch to grahpql calls.

### DIFF
--- a/certification/pyxis/pyxis.go
+++ b/certification/pyxis/pyxis.go
@@ -328,6 +328,48 @@ func (p *pyxisEngine) createTestResults(ctx context.Context, testResults *TestRe
 	return &newTestResults, nil
 }
 
+func (p *pyxisEngine) createArtifacts(ctx context.Context, artifact *Artifact) (*Artifact, error) {
+	b, err := json.Marshal(artifact)
+	if err != nil {
+		log.Error(err)
+		return nil, err
+	}
+	req, err := p.newRequestWithApiToken(ctx, http.MethodPost, getPyxisUrl(fmt.Sprintf("projects/certification/id/%s/artifacts", p.ProjectId)), bytes.NewReader(b))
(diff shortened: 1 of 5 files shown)


certification/pyxis/pyxis.go
```diff
@@ -328,6 +328,48 @@ func (p *pyxisEngine) createTestResults(ctx context.Context, testResults *TestRe
 	return &newTestResults, nil
 }
 
+func (p *pyxisEngine) createArtifacts(ctx context.Context, artifact *Artifact) (*Artifact, error) {
```
bcrochet, line 331: This should be named 'createArtifact'

certification/pyxis/submit.go
```diff
@@ -57,6 +57,15 @@ func (p *pyxisEngine) SubmitResults(ctx context.Context, certProject *CertProjec
 		}
 	}
 
+	for _, artifact := range artifacts {
+		artifact.ImageID = certImage.ID
+		_, err = p.createArtifacts(ctx, &artifact)
```
bcrochet, line 62: If we aren't using the return value, collapse the if statement to this line.

bcrochet (approved):
/lgtm

Approving, but createArtifacts needs to be named createArtifact.

certification/pyxis/pyxis.go
acornett21: fixed

certification/pyxis/submit.go
acornett21: fixed

jomkz (approved):

Issue #472, opened 2022-03-10
Add support to send image layer `sha256` to pyxis


acornett21:
## Is your feature request related to a problem? Please describe.
In order for `Clair` to be able to scan for vulnerabilities preflight needs to send `certImage.parsed_data.layers[]` to pyxis. The layers array is a list of strings of the `sha256` of the layers. This data is in the `crane config <img>` (see below)

## Describe the solution you'd like.
We need to save off the layer `sha256` information when writing `cert-image.json` to disk, this way it can be mashalled and sent to pyxis later.

## Describe alternatives you've considered.
N/A

## Additional context.
```
 "rootfs": {
    "type": "layers",
    "diff_ids": [
      "sha256:1733579794674bb13464023e8b7839ad924216064387c3bc9f53c33a4b3cd3c1",
      "sha256:47f871cfc125200d9264414fdf234fb7c57beea52683c16fc73a115463ed1454",
      "sha256:8a4a82f5e7f8dc8bb753e8ce3af7a0a7dbc77d4424448c67e5e7d30bac25d4c6",
      "sha256:e28bbcda30e0c80b9ef5d7513d949cb778569d1b34944abd6f5574539d0469b6",
      "sha256:a1fc8c6dd0aa454c4b0a1b3e738754b01757e45db2f459e8de8f9d664518d09d",
      "sha256:047f99f97b37b3d9eba5e19908d83eae648712f313c3dbe02a93ef3d8b09c1b2",
      "sha256:fc0a24266cae3f1173d40c6f4e89835156cc49821fdbf5c520a1619ceb286a19",
      "sha256:9a24a693316b84090d7fc2e8d85b9de1529d1fda3544cc448143852fdc69ce4b",
      "sha256:46181b9f594036ce2587865e6aa9e6c92f1e7f6c5b64de055f627c81f1c949c3",
      "sha256:95873c7a93997c2b6ac43ed7fac3f3a5b4cf9887d7edb0a924a2a9bd51c90721",
      "sha256:0385b2f12a960b1c87135d74e83603606445e33719d08ecd37a20cdbe6238b5c",
      "sha256:aa0b8c3616cbcd0ecc13879bd1ac3da60acce06b7467a1f932a0c0de0e3fb777",
      "sha256:e7d768a64c6f0d1ffad9576960040d22966403af9b746848f2c15afafb365ba2",
[...]

PR #476, opened 2022-03-10
added in layers array into ParsedData struct since this is required

acornett21:
- Fixes: #472
- Adding in layers array since this info is required by Clair


bcrochet (approved):
/lgtm

jomkz (approved):

PR #481, opened 2022-03-14
added base64 encoded contents to artifacts

gautamkrishnar:
Added base64 encoded contents to artifacts

#### Context
- Pyxis API is expecting base64 encoded contents instead of plain text In the content field: https://catalog.qa.redhat.com/api/containers/v1/ui/#/Images/pyxis.rest.legacy.cert_project_artifacts.get_certification_artifacts_by_image_id
Screenshot 2022-03-14 at 2 03 33 PM

### DIFF
--- a/cmd/check_container.go
+++ b/cmd/check_container.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"context"
+	"encoding/base64"
 	"encoding/json"
 	"fmt"
 	"io"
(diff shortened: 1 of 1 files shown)


cmd/check_container.go
```diff
@@ -23,6 +23,7 @@ import (
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
+	b64 "encoding/base64"
```
acornett21, line 26: Since this is from the go library, there isn't a need to add an `alias` the package. It can be referenced as `base64` which is more clear. And this import can be moved to the rest of the `go` imports.

jomkz (approved):
/lgtm

acornett21:
@gautamkrishnar Can you squash the commits to only be one commit?

gautamkrishnar:
@acornett21 done ✅

bcrochet (approved):
/lgtm

PR #482, opened 2022-03-15
check image under test for a certified uncompressed top layer id

### LINKED ISSUE #25: Improve BasedOnUBI check to ensure reliable validation of UBI vs. RHEL images
The BasedOnUBi check currently checks to ensure that the `/etc/os-release` file contains the following keys:

```
ID="rhel"
NAME="Red Hat Enterprise Linux"
```

Technically, this passes with RHEL images:

```
podman run \
    -it --rm \
    --entrypoint cat \
    registry.access.redhat.com/rhel7@sha256:d7de4fd7b3850d03d2d432a3dce209e99db31d6a4a3d0017a195363208ade4f5 \
    /etc/os-release \
    -- | egrep "^ID=|^NAME="

NAME="Red Hat Enterprise Linux Server"
ID="rhel"
```

We'll need to decide:
* Should we be more explicit in identifying something is UBI, and how we do that, or if we want to pass RHEL-based images
* How we identify that something is specifically UBI and not RHEL

If we want to be more specific with this check, we'll need to
* Update `certification/internal/shell/based_on_ubi.go` to factor in the additional validation
* Update `certification/internal/shell/based_on_ubi_test.go` to test that validation

skattoju:
Addresses #25 this will make the current based on ubi check fail due to the query to pyxis timing out. This improvement to the based on ubi check attempts to emulate[1] There is an issue open to refactor the query on the pyxis side[2]

[1]https://gitlab.cee.redhat.com/cvp/pipeline/-/blob/master/src/org/cvp/pipeline/IsvContainerUtils.groovy#L17
[2]https://issues.redhat.com/browse/ISV-1843

Fixes #25

### DIFF
--- a/certification/engine/engine.go
+++ b/certification/engine/engine.go
@@ -3,13 +3,17 @@ package engine
 
 import (
 	"fmt"
+	"net/http"
+	"time"
 
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification"
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/errors"
 	internal "github.com/redhat-openshift-ecosystem/openshift-preflight/certification/internal/engine"
 	containerpol "github.com/redhat-openshift-ecosystem/openshift-preflight/certification/internal/policy/container"
 	operatorpol "github.com/redhat-openshift-ecosystem/openshift-preflight/certification/internal/policy/operator"
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/pyxis"
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/runtime"
+	"github.com/spf13/viper"
 )
 
 // CheckEngine defines the functionality necessary to run all checks for a policy,
@@ -89,8 +93,9 @@ var (
 	hasNoProhibitedCheck   certification.Check = &containerpol.HasNoProhibitedPackagesCheck{}
 	hasRequiredLabelsCheck certification.Check = &containerpol.HasRequiredLabelsCheck{}
 	runAsRootCheck         certification.Check = &containerpol.RunAsNonRootCheck{}
-	basedOnUbiCheck        certification.Check = &containerpol.BasedOnUBICheck{}
 	hasModifiedFilesCheck  certification.Check = &containerpol.HasModifiedFilesCheck{}
+	basedOnUbiCheck        certification.Check = containerpol.NewBasedOnUbiCheck(pyxis.NewPyxisEngine(viper.GetString("pyxis_api_token"),
+		viper.GetString("certification_project_id"), &http.Client{Timeout: 60 * time.Second}))
 	// runnableContainerCheck  certification.Check = containerpol.NewRunnableContainerCheck(internal.NewPodmanEngine())
 	// runSystemContainerCheck certification.Check = containerpol.NewRunSystemContainerCheck(internal.NewPodmanEngine())
 )
--- a/certification/internal/policy/container/base_on_ubi.go
+++ b/certification/internal/policy/container/base_on_ubi.go
@@ -1,73 +1,67 @@
 package container
 
 import (
-	"os"
-	"path/filepath"
-	"strings"
+	"context"
 
 	cranev1 "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification"
+	pyxis "github.com/redhat-openshift-ecosystem/openshift-preflight/certification/pyxis"
 	log "github.com/sirupsen/logrus"
 )
 
 // BasedOnUBICheck evaluates if the provided image is based on the Red Hat Universal Base Image
 // by inspecting the contents of the `/etc/os-release` and identifying if the ID is `rhel` and the
 // Name value is `Red Hat Enterprise Linux`
-type BasedOnUBICheck struct{}
-
-func (p *BasedOnUBICheck) Validate(imgRef certification.ImageReference) (bool, error) {
-	labels, err := p.getLabels(imgRef.ImageInfo)
-	if err != nil {
-		return false, err
-	}
-
-	osRelease, err := p.getOsReleaseContents(imgRef.ImageFSPath)
-	if err != nil {
-		log.Debugf("could not retrieve contents of os-release")
-		return false, err
-	}
-
-	return p.validate(labels, osRelease)
+type BasedOnUBICheck struct {
+	LayerHashCheckEngine layerHashChecker
 }
 
-func (p *BasedOnUBICheck) getLabels(image cranev1.Image) (map[string]string, error) {
+type layerHashChecker interface {
+	CheckRedHatLayers(ctx context.Context, layerHashes []cranev1.Hash) ([]pyxis.CertImage, error)
+}
+
+func NewBasedOnUbiCheck(layerHashChecker layerHashChecker) *BasedOnUBICheck {
+	return &BasedOnUBICheck{LayerHashCheckEngine: layerHashChecker}
+}
+
+func (p *BasedOnUBICheck) Validate(imgRef certification.ImageReference) (bool, error) {
+	layerHashes, err := p.getImageLayers(imgRef.ImageInfo)
+	if err != nil {
+		return false, err
+	}
+
+	return p.validate(layerHashes)
+}
+
+func (p *BasedOnUBICheck) getImageLayers(image cranev1.Image) ([]cranev1.Hash, error) {
 	configFile, err := image.ConfigFile()
 	if err != nil {
 		return nil, err
 	}
-	return configFile.Config.Labels, nil
+
+	return configFile.RootFS.DiffIDs, nil
 }
 
-func (p *BasedOnUBICheck) getOsReleaseContents(path string) ([]string, error) {
-	osrelease, err := os.ReadFile(filepath.Join(path, "etc", "os-release"))
+func (p *BasedOnUBICheck) checkRedHatLayers(ctx context.Context, layerHashes []cranev1.Hash) (bool, error) {
+	certImages, err := p.LayerHashCheckEngine.CheckRedHatLayers(ctx, layerHashes)
 	if err != nil {
-		log.Debug("could not open os-release file for reading")
-		return nil, err
+		log.Error("Error when querying pyxis for uncompressed top layer ids", err)
 	}
-
-	return strings.Split(string(osrelease), "\n"), nil
-}
-
-func (p *BasedOnUBICheck) validate(labels map[string]string, osRelease []string) (bool, error) {
-	var hasRHELID, hasRHELName bool
-	for _, value := range osRelease {
-		line := strings.Split(value, "=")
-		if line[0] == "ID" && line[1] == `"rhel"` {
-			log.Trace("Has RHEL ID")
-			hasRHELID = true
-			continue
-		}
-		if line[0] == "NAME" && strings.Contains(line[1], "Red Hat Enterprise Linux") {
-			log.Trace("Has RHEL Name")
-			hasRHELName = true
-			continue
-		}
-	}
-
-	if hasRHELID && hasRHELName {
+	if certImages != nil && len(certImages) >= 1 {
 		return true, nil
 	}
+	return false, nil
+}
 
+func (p *BasedOnUBICheck) validate(layerHashes []cranev1.Hash) (bool, error) {
+	ctx := context.Background()
+	hasUBIHash, err := p.checkRedHatLayers(ctx, layerHashes)
+	if err != nil {
+		log.Error("Unable to verify layer hashes", err)
+	}
+	if hasUBIHash {
+		return true, nil
+	}
 	return false, nil
 }
 
--- a/certification/internal/policy/container/base_on_ubi_test.go
+++ b/certification/internal/policy/container/base_on_ubi_test.go
@@ -1,56 +1,36 @@
 package container
 
 import (
+	"context"
 	"os"
-	"path/filepath"
 
 	cranev1 "github.com/google/go-containerregistry/pkg/v1"
 	fakecranev1 "github.com/google/go-containerregistry/pkg/v1/fake"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification"
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/pyxis"
 )
 
-func labelsForUbiCheck(label string, bad bool) map[string]string {
-	labels := map[string]string{
-		"name":                 "name",
-		"vendor":               "vendor",
-		"version":              "version",
-		"release":              "release",
-		"summary":              "summary",
-		"description":          "description",
-		"com.redhat.component": label,
-	}
-
-	if bad {
-		delete(labels, "com.redhat.component")
-	}
-
-	return labels
-}
-
-func goodLabelsConfigFile() (*cranev1.ConfigFile, error) {
+func ConfigFile() (*cranev1.ConfigFile, error) {
 	return &cranev1.ConfigFile{
-		Config: cranev1.Config{
-			Labels: labelsForUbiCheck("ubi8-container", false),
-		},
+		Config: cranev1.Config{},
 	}, nil
 }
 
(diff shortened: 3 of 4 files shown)


certification/pyxis/layers.go
```diff
@@ -0,0 +1,71 @@
...
+	"net/http"
+	"strings"
+
+	cranev1 "github.com/google/go-containerregistry/pkg/v1"
+
+	log "github.com/sirupsen/logrus"
+)
+
+func (p *pyxisEngine) CheckRedHatLayers(layerHashes []cranev1.Hash) (*CertImage, error) {
+	ctx := context.Background()
```
bcrochet, line 17: This should be passed in, not created here.

certification/pyxis/layers.go
```diff
@@ -0,0 +1,71 @@
...
+
+	layerIds := make([]string, 0, len(layerHashes))
+	for _, layer := range layerHashes {
+		layerIds = append(layerIds, layer.Hex)
```
bcrochet, line 21: Hex is the format that pyxis expects in the query?

certification/pyxis/layers.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		log.Error(err)
+		return nil, err
+	}
+
+	type imageList struct {
+		Page     int         `json:"page"`
```
bcrochet, line 54: If we aren't going to utilize Page, PageSize, or Total, there is no need to include them here.

certification/pyxis/layers.go
```diff
@@ -0,0 +1,71 @@
...
+	for _, layer := range layerHashes {
+		layerIds = append(layerIds, layer.Hex)
+	}
+
+	log.Tracef("the layerIds passed to pyxis are %s", layerIds)
+
+	req, err := p.newRequestWithApiToken(
+		ctx,
+		http.MethodGet,
+		getPyxisUrl(fmt.Sprintf("images?page_size=500&page=0&filter=repositories.registry=in=(registry.redhat.io,registry.access.redhat.com) and uncompressed_top_layer_id=in=(%s)", strings.Join(layerIds, ","))),
```
acornett21, line 29: Do we really need `page_size` and `page` here? isn't the default of `100` enough for our use case?

certification/pyxis/layers.go
```diff
@@ -0,0 +1,71 @@
...
+	if resp.Body != nil {
+		defer resp.Body.Close()
+	}
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		log.Error(err)
+		return nil, err
+	}
+
```
acornett21, line 52: Because go will unmarshall anything it gets back and not validate it, we should add a check here to make sure we are getting a `200` back before unmarshalling

```
if !checkStatus(resp.StatusCode) {

File: certification/pyxis/layers.go [hunk 6]
```diff
@@ -0,0 +1,71 @@
...
+	resp, err := p.Client.Do(req)
+	if err != nil {
+		log.Error(err)
+		return nil, err
+	}
+
+	if resp.Body != nil {
```
acornett21, line 43: This `if` statement is not needed, go guarantees a non-nil body.

https://pkg.go.dev/net/http#Response.Body

certification/pyxis/layers.go
```diff
@@ -0,0 +1,71 @@
...
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"strings"
+
+	cranev1 "github.com/google/go-containerregistry/pkg/v1"
+
```
acornett21, line 12: Nit: can we organize the imports?

certification/internal/policy/container/base_on_ubi.go
```diff
@@ -1,12 +1,18 @@
 package container
 
 import (
+	"net/http"
 	"os"
 	"path/filepath"
 	"strings"
+	"time"
+
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/errors"
```
acornett21, line 10: Nit: can we organize the imports?

acornett21 (changes requested):

certification/internal/policy/container/base_on_ubi.go
```diff
@@ -64,7 +83,24 @@ func (p *BasedOnUBICheck) validate(labels map[string]string, osRelease []string)
...
-	if hasRHELID && hasRHELName {
+	projectId := viper.GetString("certification_project_id")
+	if projectId == "" {
+		return false, errors.ErrEmptyProjectID
+	}
+	if strings.HasPrefix(projectId, "ospid-") {
+		projectId = strings.Split(projectId, "-")[1]
+	}
+	apiToken := viper.GetString("pyxis_api_token")
+	pyxisEngine := pyxis.NewPyxisEngine(apiToken, projectId, &http.Client{Timeout: 60 * time.Second})
```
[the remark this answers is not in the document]
skattoju: now passing in via NewBasedOnUBICheck

certification/pyxis/layers.go
skattoju: good point might have to do layer.Algorithm+":"+layer.Hex

certification/pyxis/layers.go
skattoju: can be removed

certification/pyxis/layers.go
skattoju: added

certification/pyxis/layers.go
skattoju: removed

certification/pyxis/layers.go
skattoju: ctrl + option + O didn't do anything 🤷  but I removed the spaces..

certification/internal/policy/container/base_on_ubi.go
skattoju: ctrl + option + O didn't do anything 🤷  but I removed the spaces..

skattoju: make fmt seems to add them back though

certification/engine/engine.go
```diff
@@ -70,6 +78,21 @@ func queryChecks(checkName string) certification.Check {
 	return nil
 }
 
+type PyxisLayerHashChecker struct{}
```
bcrochet, line 81: This is not needed.

certification/engine/engine.go
```diff
@@ -89,8 +112,8 @@ var (
 	hasNoProhibitedCheck   certification.Check = &containerpol.HasNoProhibitedPackagesCheck{}
 	hasRequiredLabelsCheck certification.Check = &containerpol.HasRequiredLabelsCheck{}
 	runAsRootCheck         certification.Check = &containerpol.RunAsNonRootCheck{}
-	basedOnUbiCheck        certification.Check = &containerpol.BasedOnUBICheck{}
 	hasModifiedFilesCheck  certification.Check = &containerpol.HasModifiedFilesCheck{}
+	basedOnUbiCheck        certification.Check = containerpol.NewBasedOnUbiCheck(&PyxisLayerHashChecker{})
```
bcrochet, line 116: This should still be passing in PyxisEngine, as it would satisfy the HashChecker interface.

certification/engine/engine.go
```diff
@@ -70,6 +78,21 @@ func queryChecks(checkName string) certification.Check {
 	return nil
 }
 
+type PyxisLayerHashChecker struct{}
+
+func (*PyxisLayerHashChecker) CheckRedHatLayers(layers []cranev1.Hash) ([]pyxis.CertImage, error) {
```
bcrochet, line 83: Not necessary to have this.

certification/internal/policy/container/base_on_ubi.go
```diff
@@ -7,13 +7,24 @@ import (
...
-type BasedOnUBICheck struct{}
+type BasedOnUBICheck struct {
+	LayerHashCheckEngine layerHashChecker
+}
+
+type layerHashChecker interface {
+	CheckRedHatLayers(layerHashes []cranev1.Hash) ([]pyxis.CertImage, error)
+}
+
+func NewBasedOnUbiCheck(layerHashChecker layerHashChecker) *BasedOnUBICheck {
```
bcrochet, line 25: When this is called you can just pass a PyxisEngine, since it satisfies the layerHashChecker interface.

rpm-manifest.json
```diff
@@ -0,0 +1,2119 @@
+{
```
bcrochet, line 1: I don't think this file should be here. Please remove it.

test/e2e/cert-image.json
```diff
@@ -0,0 +1,247 @@
+{
```
bcrochet, line 1: Remove this file

test/e2e/rpm-manifest.json
```diff
@@ -0,0 +1,2119 @@
+{
```
bcrochet, line 1: Please remove.

cert-image.json
```diff
@@ -0,0 +1,247 @@
+{
```
bcrochet, line 1: Remove this file

bcrochet (changes requested):
Comments about implementation and some extraneous files.

certification/internal/policy/container/base_on_ubi_test.go
```diff
@@ -81,7 +98,22 @@ NAME="Red Hat Enterprise Linux"
 		os.RemoveAll(imageRef.ImageFSPath)
```
bcrochet, line 98: You can also remove all this junk that was for the previous way of doing this check.

rpm-manifest.json
skattoju: git add . ask questions later 😂

skattoju: removed 😅

certification/engine/engine.go
skattoju: removed 👍

certification/engine/engine.go
skattoju: yeah anything that satisfies the interface 👍

certification/engine/engine.go
skattoju: removed 👍

certification/internal/policy/container/base_on_ubi_test.go
skattoju: this becomes redundant yeah but i guess CVP does both ?

certification/pyxis/layers.go
skattoju: creating in validate for now ..

certification/pyxis/layers.go
skattoju: removed

certification/engine/engine.go
```diff
@@ -70,6 +77,22 @@ func queryChecks(checkName string) certification.Check {
 	return nil
 }
 
+var (
+	pyxisProjectId string
+	pyxisApiToken  string
+)
+
+func getPyxisParams() {
+	pyxisProjectId = viper.GetString("certification_project_id")
```
bcrochet, line 86: This is unnecessary. I've added code to make sure that it is neither empty nor starts with a prefix. You can just call viper.GetString() where it is needed. This function can go away. See #492

certification/engine/engine.go
```diff
@@ -89,8 +112,8 @@ var (
 	hasNoProhibitedCheck   certification.Check = &containerpol.HasNoProhibitedPackagesCheck{}
 	hasRequiredLabelsCheck certification.Check = &containerpol.HasRequiredLabelsCheck{}
 	runAsRootCheck         certification.Check = &containerpol.RunAsNonRootCheck{}
-	basedOnUbiCheck        certification.Check = &containerpol.BasedOnUBICheck{}
 	hasModifiedFilesCheck  certification.Check = &containerpol.HasModifiedFilesCheck{}
+	basedOnUbiCheck        certification.Check = containerpol.NewBasedOnUbiCheck(pyxis.NewPyxisEngine(pyxisApiToken, pyxisProjectId, &http.Client{Timeout: 60 * time.Second}))
```
bcrochet, line 116: Just use viper here directly.

certification/internal/policy/container/base_on_ubi_test.go
```diff
@@ -29,28 +31,27 @@ func labelsForUbiCheck(label string, bad bool) map[string]string {
 	return labels
 }
 
-func goodLabelsConfigFile() (*cranev1.ConfigFile, error) {
+func LabelsConfigFile() (*cranev1.ConfigFile, error) {
 	return &cranev1.ConfigFile{
 		Config: cranev1.Config{
-			Labels: labelsForUbiCheck("ubi8-container", false),
+			Labels: labelsForUbiCheck("ubi-container", false),
```
bcrochet, line 37: This whole func is no longer needed.

certification/internal/policy/container/base_on_ubi_test.go
bcrochet: I was unaware of that. If that's the case, I guess we should leave it in, but... A better test would be to check only the top layer? Perhaps we add that back at a later date?

certification/internal/policy/container/base_on_ubi_test.go
```diff
@@ -59,45 +60,21 @@ var _ = Describe("BaseOnUBI", func() {
 		imageRef        certification.ImageReference
 	)
 
-	const osrelease = "os-release"
-
 	BeforeEach(func() {
 		fakeImage := fakecranev1.FakeImage{
```
bcrochet, line 64: You can just create the FakeImage without any params.

certification/pyxis/layers.go
```diff
@@ -0,0 +1,64 @@
...
+		return nil, err
+	}
+
+	resp, err := p.Client.Do(req)
+	if err != nil {
+		log.Error(err)
+		return nil, err
+	}
+
+	if resp.StatusCode != 200 {
```
bcrochet, line 41: Make this a range between 200 and 299. Anything outside of that is considered an error.

bcrochet (changes requested):
A few more comments.

certification/internal/policy/container/base_on_ubi_test.go
skattoju: yeah because this might break at some point as it is different between ubi-7 and ubi-8 .. so removed it ..

certification/internal/policy/container/base_on_ubi_test.go
skattoju: was getting a null pointer so left it... but all of that is probably overkill just need a subset will try to remove most of it..

certification/internal/policy/container/base_on_ubi_test.go
skattoju: it didnt like it .. was getting null pointer so put the config back but will see what exactly was missing..

certification/pyxis/layers.go
skattoju: gotcha

certification/pyxis/layers.go
bcrochet: Run 'make fmt'

certification/pyxis/layers.go
skattoju: technically 400-599 are errors no ? but for our purposes think we should error on anything other than 200 ? would we handle other 20x codes gracefully ?

certification/internal/policy/container/base_on_ubi_test.go
skattoju: removed

certification/internal/policy/container/base_on_ubi_test.go
skattoju: just passing an empty config now

certification/engine/engine.go
skattoju: yup makes it cleaner 👍

certification/pyxis/layers.go
acornett21: @bcrochet Is correct what you want is `layer.String`

bcrochet (approved):

certification/pyxis/layers.go
acornett21: @skattoju You still need the `defer resp.Body.Close()`, the `if` is what is not needed. Sorry this wasn't clear.

skattoju: re-added 😅

acornett21 (approved):

PR #484, opened 2022-03-15
changing parsedData.Layers to come from the imageInfo.Manifest

acornett21:
- Changing where `parsedData.Layers` is sourced from, obtaining it from imageInfo.Manifest.Layers instead of `config.RootFS.DiffIDs`


bcrochet (approved):
/lgtm

PR #489, opened 2022-03-16
Fail multi-arch-build on any failure

### LINKED ISSUE #488: preflight multi-arch build process should fail gracefully if all binaries are not produced
## Bug Description
Right now we are just do a `grep` to see if _any_ binary as built, but this leads to us missing the fact that any given binary might not be built.

`main`

## Steps to Reproduce:
1) Take PR #485 Into your local branch
2) Run `make build-multi-arch`
3) See that the power build fails, and does not produce a binary

## Expected Result
I'd expect our make/build process to fail if a given binary is not produced that we requested to be built.

## Actual Result
Preflight builds 3 of the 4 binary, but exits like it successfully built all 4 binaries.

## Additional Context
We should try to grep that all binaries are present and if they are not produce an exit(1) failure.

bcrochet:
Implement a template and a new foreach loop to create individual
targets for each architecture. There is currently no support for a
GOOS other than Linux. If that becomes necessary, we can add support
for other OSs.

FIxes #488

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,11 +15,16 @@ build:
 	@ls | grep preflight
 
 .PHONY: build-multi-arch
-build-multi-arch:
-	$(foreach GOOS, $(PLATFORMS),\
-	$(foreach GOARCH, $(ARCHITECTURES),\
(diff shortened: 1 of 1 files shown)


jomkz (approved):
/lgtm

acornett21 (changes requested):
For whatever reason, this will not work. Go still produces an exit code of `0` for this, since it built `one successful` binary. I have no idea why that is the case or if there is a flag we can add.

bcrochet:
> For whatever reason, this will not work. Go still produces an exit code of `0` for this, since it built `one successful` binary. I have no idea why that is the case or if there is a flag we can add.

It was make, not Go, that was just continuing on. I'm now checking the exit status after each forloop execution.

skattoju (approved):
nice

acornett21 (approved):

PR #490, opened 2022-03-16
TESTONLY: DNM: Test go-rpmdb with mattn/go-sqlite3

bcrochet:
This uses a CGO dropin replacement for Sqlite3 that does not fail the build. More testing would be needed.


bcrochet:
Concept proven.

PR #498, opened 2022-03-17
Simplify Pyxis Submit call

bcrochet:
There were too many input parameters and too many output returns.
This makes a type for each to contain each of them. This will allow
expansion of returns and inputs without having to modify the entire
Submit method, since there will only be 2 returns.


acornett21 (approved):
Nice work! I hated this too, and it was confusing to look at.

jomkz (approved):

Issue #500, opened 2022-03-17
[critical] Regression in 1.1.0-beta1: Preflight can't test images referred by sha anymore


tkrishtop:
## Bug Description
Starting from the 1.1.0-beta1 version, Preflight can't pull the images referred by digest (sha) anymore.
1.1.0-alpha2 is the last version working fine.

The regression was probably introduced in the fix for #467 because `:` could appear in three situations:
- registry with port
- image referred by digest
- image tag

[Here is PR that probably introduced the bug](https://github.com/redhat-openshift-ecosystem/openshift-preflight/pull/468/files)

## Version and Command Invocation
Starting from the 1.1.0-beta1 version.
1.1.0-alpha2 is the last version without this bug.

## Steps to Reproduce:
```
# preflight_1.1.0-beta3 == bug is already here
$ PFLT_JUNIT=true \
PFLT_ARTIFACTS=$PWD/artifacts \
PFLT_LOGFILE=$PWD/artifacts/preflight.log \
PFLT_LOGLEVEL=trace \
./preflight_1.1.0-beta3 \
check container \
quay.io/rh-nfv-int/testpmd-operator@sha256:3e8fc703c71a7ccaca24b7312f8fcb3495370c46e7abc12975757b76430addf5 \
--REMOVED \
--pyxis-api-token=REMOVED \
--docker-config=$PWD/repoconfig/config.json

time="2022-03-17T18:04:34+01:00" level=info msg="certification library version 0.0.0 <commit: 06a9fd1520a37dc5b39328655f6ceaabf6471a51>"
time="2022-03-17T18:04:34+01:00" level=debug msg="URL is: https://catalog.redhat.com/api/containers/v1/projects/certification/id/REMOVED"
time="2022-03-17T18:04:36+01:00" level=debug msg="Certification project name is: Test Project: certify simple-demo-operator"
time="2022-03-17T18:04:36+01:00" level=debug msg="target image: quay.io/rh-nfv-int/testpmd-operator@sha256:3e8fc703c71a7ccaca24b7312f8fcb3495370c46e7abc12975757b76430addf5"
time="2022-03-17T18:04:36+01:00" level=debug msg="pulling image from target registry"
time="2022-03-17T18:04:37+01:00" level=debug msg="temporary directory is /tmp/preflight-3498777854"
time="2022-03-17T18:04:37+01:00" level=debug msg="exporting and flattening image"
[...]

PR #501, opened 2022-03-17
using registry/repository/tag information from imageRef struct

acornett21:
- Fixes: #500 
- Using registry/repository/tag information from imageRef struct instead of creating a `name.NewTag` for repository.tags
- Not everything passed in is a `tag` so `NewTag` was failing for `sha` information, `imageRef` struct is populated with `name.ParseReference` which accepts `sha` or `tag`


skattoju (approved):
lgtm

jomkz (approved):

PR #511, opened 2022-03-24
unsupported image error in case rpm path isn't found

### LINKED ISSUE #502: [Doc or bug?] check container for non-UBI and non-RHEL based images
Should `check container` support non-UBI or non-RHEL images?

```sh
preflight check container docker.io/alpine --pyxis-api-token XX  --certification-project-id XX --pyxis-host catalog.uat.redhat.com/api/containers
time="2022-03-21T01:39:41-07:00" level=info msg="certification library version 0.0.0 <commit: 06a9fd1520a37dc5b39328655f6ceaabf6471a51>"
Error: open /tmp/preflight-3153558342/fs/var/lib/rpm/Packages: no such file or directory
time="2022-03-21T01:39:44-07:00" level=fatal msg="open /tmp/preflight-3153558342/fs/var/lib/rpm/Packages: no such file or directory"
```

## Expected Result

An error stating only UBI and RHEL images are supported.  potentially with a doc link.
Or skip the packages check in case non-RHEL is actually intended to be supported.

itroyano:
Fixes #502

### DIFF
--- a/certification/errors/errors.go
+++ b/certification/errors/errors.go
@@ -34,4 +34,5 @@ var (
 	Err409StatusCode                   = errors.New("remote API returned conflict")
 	ErrInvalidImageUri                 = errors.New("image uri could not be parsed")
 	ErrRetrievingLayers                = errors.New("could not get container layers")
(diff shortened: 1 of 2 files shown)


bcrochet (approved):
/lgtm

jomkz (approved):

PR #513, opened 2022-03-24
Set a non-root user

itroyano:
Setting a non-root user for the Preflight image, to make sure it passes the `run_as_nonroot` check && we aren't doing any sudoing during the checks.


jomkz (approved):
/lgtm

bcrochet (approved):
/lgtm

PR #514, opened 2022-03-24
log pyxis layer check request

skattoju:
added extra log to showing the generated query and endpoint used to check layers in pyxis

### DIFF
--- a/certification/pyxis/layers.go
+++ b/certification/pyxis/layers.go
@@ -34,6 +34,7 @@ func (p *pyxisEngine) CheckRedHatLayers(ctx context.Context, layerHashes []crane
 		return nil, err
 	}
 
+	log.Tracef("URL is %s", req.URL)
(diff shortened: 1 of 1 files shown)


certification/pyxis/layers.go
```diff
@@ -34,6 +34,7 @@ func (p *pyxisEngine) CheckRedHatLayers(ctx context.Context, layerHashes []crane
 		return nil, err
 	}
 
+	log.Tracef("request to pyxis is %s?filter=%s", getPyxisUrl("images"), pyxisQuery)
```
acornett21, line 37: We should probably log the URL in the request and not trust the code we built. Like the rest of the pyxis calls.
```suggestion
	log.Debugf("URL is: %s",  req.URL)
```

skattoju: agreed 👍

acornett21 (approved):

PR #520, opened 2022-03-25
Use the default pyxis host for BasedOnUBI

bcrochet:
Since the default pyxis host is prod, that is what we should pass
to the BasedOnUBI pyxis engine. It removes the "magic constant" from
the initialization.

### DIFF
--- a/certification/defaults.go
+++ b/certification/defaults.go
@@ -4,5 +4,6 @@ var (
 	DefaultCertImageFilename   = "cert-image.json"
 	DefaultRPMManifestFilename = "rpm-manifest.json"
 	DefaultTestResultsFilename = "results.json"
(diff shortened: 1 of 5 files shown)


acornett21 (approved):

certification/defaults.go
```diff
@@ -4,5 +4,6 @@ var (
 	DefaultCertImageFilename   = "cert-image.json"
 	DefaultRPMManifestFilename = "rpm-manifest.json"
 	DefaultTestResultsFilename = "results.json"
+	DefaultPyxisHost           = "catalog.redhat.com/api/containers"
```
skattoju, line 7: was going to say 👍

skattoju (approved):

jomkz (approved):

PR #525, opened 2022-03-28
Detect if layer 0 is non-zero length and shift to layer 1

### LINKED ISSUE #523: Base layer is not always non-zero length, causing HasModifiedFiles to fail
The base layer of an image is sometimes zero. Ubi-micro, for instance, has a zero-length layer at its base. This is most likely because ubi-micro is based on 'FROM scratch'.

HasModifiedFiles needs to take this into account.

If layer[0] is len(0), then layer[1] should be considered the base layer. And the modified files search should continue from layer[2]

komish:
In `preflight check container`, if the base image of the container under test has a zero length (e.g. `FROM scratch` in the Containerfile), then there will be no files to use to evaluate if something was changed in a subsequent layer.

This patch shifts the `layerFiles` over one layer if the first layer is deemed to be a zero-length file. I don't believe it's plausible to get additional layers with zero length, so we don't account for it.

This also adds comments throughout the files associated with the check, and also removes a deprecated comment in another, unrelated file in a separate commit.

Fixed #523

### DIFF
--- a/certification/internal/policy/container/has_modified_files.go
+++ b/certification/internal/policy/container/has_modified_files.go
@@ -14,11 +14,16 @@ import (
 )
 
 // HasModifiedFilesCheck evaluates that no files from the base layer have been modified by
-// subsequent layers.
+// subsequent layers by comparing the file list installed by Packages against the file list
+// modified in subsequent layers.
 type HasModifiedFilesCheck struct{}
 
 type packageFilesRef struct {
-	LayerFiles   [][]string
+	// LayerFiles contains a slice of files modified by layer.
+	LayerFiles [][]string
+	// PackagesFiles is a map of modified files. the anonymous struct
+	// here is just used to allow us to search package files by name
+	// as the key of this map, and doesn't have any value.
 	PackageFiles map[string]struct{}
 }
 
@@ -30,11 +35,19 @@ func (p *HasModifiedFilesCheck) Validate(ctx context.Context, imgRef certificati
 	return p.validate(packageFiles)
 }
 
+// getDataToValidate returns the list of files (packageFilesRef.PackageFiles)
+// installed via packages from the container image,and the list of files (packageFilesRef.LayerFiles)
+// modified/added via layers in the image.
 func (p *HasModifiedFilesCheck) getDataToValidate(ctx context.Context, imgRef certification.ImageReference) (*packageFilesRef, error) {
+	// Get a list of packages from the RPM database. This avoids having to rely on
+	// rpm, dnf, yum, etc. being installed in the image.
 	pkgList, err := rpm.GetPackageList(ctx, imgRef.ImageFSPath)
 	if err != nil {
 		return nil, err
 	}
+
+	// Get the files put in place on the filesystem by the
+	// installed packages.
 	packageFiles := make(map[string]struct{}, len(pkgList))
 	for _, pkg := range pkgList {
 		filenames, err := pkg.InstalledFiles()
@@ -42,6 +55,8 @@ func (p *HasModifiedFilesCheck) getDataToValidate(ctx context.Context, imgRef ce
 			return nil, err
 		}
 		for _, file := range filenames {
+			// A struct is used here, but it is unimportant and
+			// should not have value.
 			packageFiles[file] = struct{}{}
 		}
 	}
@@ -53,6 +68,8 @@ func (p *HasModifiedFilesCheck) getDataToValidate(ctx context.Context, imgRef ce
 
 	files := make([][]string, 0, len(layers))
 
+	// Uncompress each layer and build a slice containing the files
+	// modified by each layer.
 	for _, layer := range layers {
 		r, err := layer.Uncompressed()
 		if err != nil {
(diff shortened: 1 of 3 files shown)


certification/internal/policy/container/has_modified_files.go
```diff
@@ -61,12 +78,16 @@ func (p *HasModifiedFilesCheck) getDataToValidate(ctx context.Context, imgRef ce
 		pathChan := make(chan string)
 
 		go func() {
+			// For each path in the pathChan, add it to the layer's
+			// list of files.
 			layerFiles := make([]string, 0)
 			for path := range pathChan {
 				layerFiles = append(layerFiles, path)
 			}
+			// also add it to the overal list of files.
```
bcrochet, line 87: overall

certification/internal/policy/container/has_modified_files.go
```diff
@@ -76,9 +97,20 @@ func (p *HasModifiedFilesCheck) getDataToValidate(ctx context.Context, imgRef ce
...
+// validate compares the list of LayerFiles and PackageFiles to see what PackageFiles
+// have been modified within the additional layers.
 func (p *HasModifiedFilesCheck) validate(packageFilesRef *packageFilesRef) (bool, error) {
 	layerFiles := packageFilesRef.LayerFiles
 	packageFiles := packageFilesRef.PackageFiles
+
+	// In cases where the image was built FROM scratch,
+	// we drop the empty layer. This should be relatively rare,
+	// but it is the case in images such as ubi-micro.
+	if len(layerFiles[0]) == 0 {
```
bcrochet, line 109: Would it be better to leave this method as is, and perhaps drop the layer in the gather method? Since we have the cranev1.Layer? But now I see that layer.Size() is the *compressed* size... So that's not exactly what we want... is it possible to have a single zero-length file in a layer and have it's size be zero?

bcrochet (changes requested):
Couple of comments/questions.

certification/internal/policy/container/has_modified_files.go
bcrochet: It doesn't seem like that would be possible... But, I think we can rely on the Compressed() size to be 32. And that the hash would be 4f4fb700ef54461cfa02571ae0db9a0dc1e0cdb5577484a6d75e68dc38e8acc1. So, we could check for either of those things. Very magic-numbery, but I think they both can be relied upon. I just feel like validate shouldn't need to massage the data any further. That's the job of getDataToValidate. Feel free to ignore me.

komish: I moved it to the getDataToValidate method, though probably not in the cleanest way. I just mirrored the exact behavior, just in this function instead of in the Validate function - only because I can't seem to test against UAT at the moment. Local tests pass, and the logic should accomplish the same thing. The `packageFilesRef` goes untouched between the two functions.

certification/internal/policy/container/has_modified_files.go
komish: Fixed

certification/internal/policy/container/has_modified_files.go
```diff
@@ -30,18 +35,28 @@ func (p *HasModifiedFilesCheck) Validate(ctx context.Context, imgRef certificati
...
 func (p *HasModifiedFilesCheck) getDataToValidate(ctx context.Context, imgRef certification.ImageReference) (*packageFilesRef, error) {
+	// Get a list of packages from the RPM database. This avoids having to rely on
+	// rpm, dnf, yum, etc. being installed in the image.
 	pkgList, err := rpm.GetPackageList(ctx, imgRef.ImageFSPath)
 	if err != nil {
 		return nil, err
 	}
+
+	// Get the files put in place on the filesystem by the
+	// installed package.
```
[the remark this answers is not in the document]
komish: Fixed

certification/internal/policy/container/has_modified_files.go
komish: Also, for what it's worth, I kept the check as designed (check for slice for a non-zero length, indicating that files were populated in that layer) for the moment, only because we don't ever want to encounter zero files in the first layer (as opposed to relying on the compressed size or the hash, which may or may not be variable). 

The way I see it, if our logic iterates over the first value in that slice, `packageFilesRef.LayerFiles`, then we should just explicitly check to make sure that's non-zero.

I think it may be possible to find a case where this... doesn't quite work how I think it should, but I'll need to spend cycles testing on it and come back to it. For the moment, though, with our UBI images, it should cover the bases.

certification/internal/rpm/rpm.go
```diff
@@ -10,6 +10,9 @@ import (
 	preflightErrors "github.com/redhat-openshift-ecosystem/openshift-preflight/certification/errors"
 )
 
+// GetPackageList returns the list of packages in the rpm database from /var/lib/rpm,
```
bcrochet, line 13: This is sort of correct... It first looks for rpmdb.sqlite at /var/lib/rpm/rpmdb.sqlite (RHEL 9). If not found, it checks for /var/lib/rpm/Packages (RHEL 8 and prior).

komish: Fixed

bcrochet (approved):
/lgtm

acornett21 (approved):

PR #528, opened 2022-03-29
logging overall results status of pass or fail in both checks

acornett21:
- Fixes: #403 
- Logging overall results of `PASS/FAILED` in both check operator and check container checks.


jomkz (approved):
/lgtm

cmd/check_container.go
```diff
@@ -152,6 +152,8 @@ var checkContainerCmd = &cobra.Command{
 			return err
 		}
 
+		log.Infof("Preflight result: %s", covertPassedOverall(results.PassedOverall))
```
bcrochet, line 155: I think this needs to be at the very end. So that it's the absolute last thing logged.

cmd/utils.go
```diff
@@ -93,3 +93,12 @@ func buildOverviewURL(projectID string) string {
 func buildScanResultsURL(projectID string, imageID string) string {
 	return fmt.Sprintf("%s/images/%s/scan-results", buildConnectURL(projectID), imageID)
 }
+
+func covertPassedOverall(passedOverall bool) string {
+	preflightResult := "FAILED"
+	if passedOverall {
+		preflightResult = "PASSED"
```
bcrochet, line 100: Just return this, and at the end just return "FAILED"

cmd/utils.go
```diff
@@ -93,3 +93,12 @@ func buildOverviewURL(projectID string) string {
 func buildScanResultsURL(projectID string, imageID string) string {
 	return fmt.Sprintf("%s/images/%s/scan-results", buildConnectURL(projectID), imageID)
 }
+
+func covertPassedOverall(passedOverall bool) string {
```
bcrochet, line 97: convert

bcrochet (approved):
/lgtm

PR #532, opened 2022-03-30
Fix source rpm name parsing

bcrochet:
Some source rpm names are breaking the parsing logic we have (that was
very close to the CVP version). We couldn't use the CVP version, as the
golang regex engine does not support positive lookaheads. So, this patch
instead utilizes Split on '-', then puts the source name back together
with any '-', minus the last two entries. Unit tests provided.


acornett21 (approved):

acornett21:
I tested this PR with all RHEL9 beta images and there wasn't any issues.

jomkz (approved):

PR #538, opened 2022-04-01
Fix nonroot home directory permissions

bcrochet:
The permissions on the /home/nonroot WORKDIR were set to root:root.
This caused any commands to fail that tried to create directories
in the preflight image. This sets the ownership properly.


jomkz (approved):
/lgtm

acornett21 (approved):

PR #539, opened 2022-04-01
Use graphql for CheckRedHatLayers

bcrochet:
It was pointed out that if too many layer hashes were passed in the
URL for a REST request, it could cause a 500 for the length of the
URL being too long. This switches the CheckRedHatLayers pyxis query
to use the GraphQL endpoint, and send the query in the body, which
has no such limit.


jomkz (approved):

certification/pyxis/layers.go
```diff
@@ -2,68 +2,60 @@ package pyxis
...
+		"contImageLayers": layerIds,
+		"registries":      []graphql.String{"registry.access.redhat.com"},
 	}
 
-	log.Tracef("URL is %s", req.URL)
-	resp, err := p.Client.Do(req)
-	if err != nil {
-		log.Error(err)
-		return nil, err
+	httpClient, ok := p.Client.(*http.Client)
```
acornett21, line 40: I think we'd still want to set a timeout on this client, otherwise there is no timeout and this call could hang forever.

acornett21, line 40: This is the Client that's on the `pyxisClient` struct correct? Or is it a net new HTTP Client?

acornett21: That's what I thought but I wanted to make sure, since the other code where that is instantiated isn't changing was hard to tell.

acornett21 (approved):

PR #554, opened 2022-04-06
adding clarity to logs when GetPackageList is called

acornett21:
- Since this method is called in multiple places, adding an errors to the error files
- Removing wrap of `errors.ErrExtractingTarball` in `Validate` method of since thats not the only possible error

### DIFF
--- a/certification/internal/engine/engine.go
+++ b/certification/internal/engine/engine.go
@@ -439,7 +439,7 @@ func getBgName(srcrpm string) string {
 func writeRPMManifest(ctx context.Context, containerFSPath string) error {
 	pkgList, err := rpm.GetPackageList(ctx, containerFSPath)
 	if err != nil {
-		log.Error("could not get rpm list, continuing without it")
+		log.Error(fmt.Errorf("%w: continuing without it", errors.ErrRPMPackageList))
 	}
 
 	// covert rpm struct to pxyis struct
--- a/certification/internal/policy/container/has_modified_files.go
+++ b/certification/internal/policy/container/has_modified_files.go
@@ -30,7 +30,7 @@ type packageFilesRef struct {
 func (p *HasModifiedFilesCheck) Validate(ctx context.Context, imgRef certification.ImageReference) (bool, error) {
 	packageFiles, err := p.getDataToValidate(ctx, imgRef)
 	if err != nil {
-		return false, fmt.Errorf("%w: %s", errors.ErrExtractingTarball, err)
+		return false, fmt.Errorf("%w: %s", errors.ErrDetectingModifiedFiles, err)
 	}
 	return p.validate(packageFiles)
 }
@@ -43,7 +43,7 @@ func (p *HasModifiedFilesCheck) getDataToValidate(ctx context.Context, imgRef ce
 	// rpm, dnf, yum, etc. being installed in the image.
 	pkgList, err := rpm.GetPackageList(ctx, imgRef.ImageFSPath)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("%w: %s", errors.ErrRPMPackageList, err)
 	}
 
 	// Get the files put in place on the filesystem by the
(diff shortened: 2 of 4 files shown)


certification/internal/engine/engine.go
```diff
@@ -437,7 +437,7 @@ func getBgName(srcrpm string) string {
 func writeRPMManifest(ctx context.Context, containerFSPath string) error {
 	pkgList, err := rpm.GetPackageList(ctx, containerFSPath)
 	if err != nil {
-		log.Error("could not get rpm list, continuing without it")
+		log.Error(fmt.Errorf("%w: %s", errors.ErrRPMPackageList, ", continuing without it"))
```
komish, line 440: Can you  include the value of `err` in your message? You can probably take "continuing without it" and add it straight into your string formatting, as well.

certification/internal/policy/container/has_modified_files.go
```diff
@@ -30,7 +30,7 @@ type packageFilesRef struct {
 func (p *HasModifiedFilesCheck) Validate(ctx context.Context, imgRef certification.ImageReference) (bool, error) {
 	packageFiles, err := p.getDataToValidate(ctx, imgRef)
 	if err != nil {
-		return false, fmt.Errorf("%w: %s", errors.ErrExtractingTarball, err)
+		return false, fmt.Errorf("%s", err)
```
komish, line 33: This looks like it needs a wrapping error.

certification/internal/policy/container/has_modified_files.go
```diff
@@ -43,7 +43,7 @@ func (p *HasModifiedFilesCheck) getDataToValidate(ctx context.Context, imgRef ce
 	// rpm, dnf, yum, etc. being installed in the image.
 	pkgList, err := rpm.GetPackageList(ctx, imgRef.ImageFSPath)
 	if err != nil {
-		return nil, err
+		return nil, errors.ErrRPMPackageList
```
komish, line 46: Can the preflight error wrap `err` here?

komish (changes requested):
Few small changes ⬆️

certification/internal/engine/engine.go
acornett21: Since we continue anyways I think adding the `err` adds unnecessary noise in this case.

certification/internal/policy/container/has_modified_files.go
acornett21: I'm trying to take away the wrapping, because the error is not always an `ErrExtractingTarball` error. Maybe it just makes more sense to return err here.

certification/internal/policy/container/has_modified_files.go
acornett21: I choose not to wrap this error here and in `has_prohibited_packages.go` to reduce the noise, but maybe for those two cases it makes sense to add wrapping.

certification/internal/engine/engine.go
bcrochet: That's fair, but it still would be easier to read if it's in the format string itself.

komish: It does add noise, but we would have nothing to use for debug purposes if we ran into failure cases (where we can't find the RPM list, but are expecting it).

certification/internal/policy/container/has_modified_files.go
komish: I think the right thing to do here is to wrap with an appropriate error, relevant to the type of check (e.g. ErrDetectingModifiedFiles).

bcrochet (approved):
/lgtm

certification/internal/engine/engine.go
acornett21: As a compromise, I wrapped the error in both checks that call `GetPackageList` but do not think the added noise here adds any value to us when trying to  debug.

certification/internal/policy/container/has_modified_files.go
acornett21: I added an error and wrapped the error that bubbles up.

PR #555, opened 2022-04-07
Document workflow for submitting container check results to red hat

komish:
### Motivation

The submit workflow of the container check (i.e. `if submit {` and beyond) felt like it would benefit from some clarification on exactly what's happening and why. 

This PR aims to improve the readability and flow of `check_container.go`'s results submission logic. This will probably be an iterative process, but for the moment I did not warrant extracting too many things into their own functions quite yet (we don't have enough reuse opportunities for this logic from what I can see).

The core behavior remains as of right now. A majority of the changes take place in `cmd/check_container.go`, but other files were touched as they related to this particular section of `check_container.go`.

The target `make test` is working as expected. If there are other ways to test, I'm all ears.

### Change summary by file

**cmd/check_container.go**

- Commented the code throughout to help developers follow the flow.
- Added a few notes in places where the variables/contructs being used were establish in a workflow outside of our submission logic.  Specifically, references to `certProject`, or the `pyxisClient` are used here, but are established earlier in the code base. Because of the distance between initialization of these variables and their use here for this workflow, I feel like it's possible that they may contain values that we don't expect. An example might be if their values are changed somewhere between where they're established, and where we use them here. No change, but at the very least it warranted a note from my perspective.
- I replaced several uses of `os.Open` and `io.Read` with a single `os.ReadFile` to reduce the `if err != nil` checks that were spread throughout.
- Wrapped errors in user-friendly error messages, and formatted that across multiple lines (for readability - totally subjective).

**certification/errors/errors.go**

- In an effort to make some of our errors more clear, I've added a net-new error for the submission workflow, and relabeled the HTTP 409 error we use for Pyxis-focused interactions.

**certification/internal/engine/engine.go**

- Just added a comment to explain the motivation behind `writeCertImage`

**certification/pyxis/pyxis.go**

- Updated the HTTP 409 error name for clarity.

**certification/pyxis/submit.go**

- Commented the SubmitResults function
- Updated HTTP 409 function names
- removed one extraneous variable declaration (`dockerImageDigest`)

**cmd/utils.go**

- Added a utility function that allows us to read an entire file on disk and also get its size in one operation. This just aims to reduce the error checking that was happening inline in our submit workflow.

---


cmd/check_container.go
```diff
@@ -152,111 +152,117 @@ var checkContainerCmd = &cobra.Command{
 			return err
 		}
 
-		// submitting results to pxysis if submit flag is set
+		// assemble artifacts and submit results to pyxis if user provided the submit flag.
 		if submit {
 			log.Info("preparing results that will be submitted to Red Hat")
-			log.Tracef("CertProject: %+v", certProject)
+			log.Tracef("CertProject: %+v", certProject) // TODO() This depends on Pyxis calls made outside of this block.
```
bcrochet, line 158: My proposed changes to the UX would move those calls to here.

komish: 💯 Indeed. When that merges, these notes should go away.

certification/errors/errors.go
```diff
@@ -31,11 +31,12 @@ var (
 	ErrUnsupportedGoType               = errors.New("go type unsupported")
 	ErrSaveFileFailed                  = errors.New("failed to save file to artifacts directory")
 	ErrNon200StatusCode                = errors.New("error calling remote API")
-	Err409StatusCode                   = errors.New("remote API returned conflict")
+	ErrPyxis409StatusCode              = errors.New("remote API returned conflict")
```
acornett21, line 34: Non Blocking: I get the change of this for clarity, but if `preflight` ever has to call another backend service, having the more generic error might make sense.

certification/pyxis/submit.go
```diff
@@ -37,36 +45,38 @@ func (p *pyxisClient) SubmitResults(ctx context.Context, certInput *Certificatio
 		}
 	}
 
-	dockerImageDigest := certImage.DockerImageDigest
```
acornett21, line 40: good catch, this was used in two places before, now only one. Smart to remove it.

cmd/check_container.go
```diff
@@ -152,111 +152,117 @@ var checkContainerCmd = &cobra.Command{
...
 
-			testResultsJsonFile, err := os.Open(path.Join(artifacts.Path(), certification.DefaultTestResultsFilename))
+			// The engine writes the certified image config to disk in a Pyxis-specific format.
+			// Include that data in our submission.
+			certImageFilepath := path.Join(artifacts.Path(), certification.DefaultCertImageFilename)
+			log.Tracef("Loading certImage data from path %s", certImageFilepath)
+			certImageBytes, err := os.ReadFile(certImageFilepath)
```
acornett21, line 164: I get that this reduces the complexity of the code and verbosity, and results in easier to read code, but don't we lose granularity of error handling in this case?

cmd/check_container.go
```diff
@@ -152,111 +152,117 @@ var checkContainerCmd = &cobra.Command{
...
+					"rpm manifest",
+					err,
+				)
 			}
 
-			certImage := new(pyxis.CertImage)
-			err = json.Unmarshal(certImageBytes, &certImage)
+			// Include Preflight's test results in our submission. pyxis.TestResults embeds them.
+			var testResults pyxis.TestResults
+			err = json.Unmarshal(formattedResults, &testResults)
```
acornett21, line 212: Since formattedResults isn't really the struct that testResults is built on, and does not have an `json` values, does this actually work as expected? ie has the payload sent to pyxis been validated?

acornett21 (commented):
Just a few things that need clarity, but overall I think this provides clarity to others.

cmd/check_container.go
```diff
@@ -152,111 +152,117 @@ var checkContainerCmd = &cobra.Command{
...
 			if err != nil {
-				return err
+				return fmt.Errorf("%w: %s", errors.ErrSubmittingToPyxis, err)
 			}
 
 			log.Info("Test results have been submitted to Red Hat.")
 			log.Info("These results will be reviewed by Red Hat for final certification.")
 			log.Infof("The container's image id is: %s.", certResults.CertImage.ID)
 			log.Infof("Please check %s to view scan results.", buildScanResultsURL(projectId, certResults.CertImage.ID))
-			log.Infof(fmt.Sprintf("Please check %s to monitor the progress.", buildOverviewURL(projectId)))
```
bcrochet, line 259: Thank you for that. I forgot to go back and fix that one. :)

cmd/check_container.go
```diff
@@ -152,111 +152,117 @@ var checkContainerCmd = &cobra.Command{
...
 			certInput := &pyxis.CertificationInput{
 				CertProject: certProject,
-				CertImage:   certImage,
-				RpmManifest: rpmManifest,
-				TestResults: testResults,
-				Artifacts:   artifacts,
+				CertImage:   &certImage,
+				RpmManifest: &rpmManifest,
+				TestResults: &testResults,
+				Artifacts:   []pyxis.Artifact{logFileArtifact},
```
bcrochet, line 253: Not sure about this change. There will be more artifacts (In fact, I think we should already be sending up the files that we've put on disk, for comparison purposes), so I think having this defined as above should remain.

bcrochet (changes requested):
Overall, looks good. Just one comment about the artifacts slice.

certification/errors/errors.go
skattoju: nit: .. or a more specific message like "pyxis remote API returned conflict" (in case we would have ErrOtherExtService409StatusCode in the future)

cmd/check_container.go
acornett21: I agree with @bcrochet on this, a list was intended for us to send more files on disk as they are needed.

certification/errors/errors.go
komish: I meant to change the message to include a note about Pyxis. I think @acornett21 has a valid point, though, so if we decide to add more remote API calls, I would vote instead to have a generic HTTP Status Code that gets wrapped with the status code itself (so we can reuse it anywhere and report an error as we need to). This will be fixed.

cmd/check_container.go
komish: Generally, I would agree, but we didn't handle `os.Open` and `io.Read` errors with any additional verbosity anyway. From my perspective, the operation we care about is "open and read" - if we can't do that, then `os.ReadFile`'s error should get report back to the user and should offer enough context.

cmd/check_container.go
komish: @acornett21 If I understood correctly, the file we read from disk was results.json. We write formattedResults to that file, no? 

In theory, this is not a very safe/reliable operation either way because, while we don't really expose it, the code supports other formatters for the UserResponse. Either way, I believe the data we were reading in using the previous implementation is the same.

Correct me if I'm wrong!

Alternatively, maybe we find an appropriate way to expose [formatters.getResponse](https://github.com/redhat-openshift-ecosystem/openshift-preflight/blob/main/certification/formatters/util.go#L10)'s behavior?

cmd/check_container.go
komish: Gotcha. We were instantiating an empty slice and then appending logFileArtifact, which we already had at the time (and could have just initialized with the value already in place). I can revert this.

certification/errors/errors.go
komish: This is resolved. I'll let @acornett21 resolve the comment thread when you have time.

cmd/check_container.go
komish: I reverted this for the moment, though I still instantiate the slice with our existing value and avoid an extra resize operation (append)

certification/errors/errors.go
acornett21: This makes more sense now, and I agree if the time comes where we call other systems, we wrap a generic message.

certification/pyxis/submit.go
acornett21: This actually needs to stay, I forgot we do not get any useful information back on conflicts, so we save this off first for calls to `GET`

cmd/check_container.go
acornett21: Looking more, I was missing that this uses `genericJSONFormatter` method which in turn uses the `formatters.UserResponse` which is in `pyxis.TestResults`, so under the covers they are the same, just getting there is different and hard to see just from this one change.

cmd/check_container.go
```diff
@@ -152,111 +152,120 @@ var checkContainerCmd = &cobra.Command{
...
 				ContentType: http.DetectContentType(logFileBytes),
 				Filename:    logFileName,
-				FileSize:    logFileInfo.Size(),
+				FileSize:    logFileSize,
 			}
 
-			artifacts := make([]pyxis.Artifact, 0, 1)
-			artifacts = append(artifacts, logFileArtifact)
+			// additional artifacts to send to Pyxis.
+			artifacts := []pyxis.Artifact{logFileArtifact}
```
bcrochet, line 247: I'm fine with this for now. We can go back to the prior method when we start adding more artifacts, and we know ahead of time how many there are.

cmd/check_container.go
bcrochet: No strong feeling there. But leaving the creation of the slice outside the creation of the results is ideal.

bcrochet (approved):

Issue #552, opened 2022-04-06
Proposal - Make pyxis credentials optional unless submitting


bcrochet:
# Proposal - Make Pyxis credentials optional unless Submitting

## Current State

The current state of `preflight check container` requires a user to provide Pyxis credentials, via the `--certification-project-id` and `--pyxis-api-token`. This requires that a project be set up in Pyxis prior to utilizing the preflight tool.

The reason for these credentials is that a call to Pyxis is required to get the project information, and from that project info, check the `OsContentType` field. If it is set to 'scratch', then a subset of container checks are run. Currently, that means three checks are not performed:

* BasedOnUBI
* HasModifiedFiles
* HasNoProhibitedPackages

## Proposal

In order to make the UX for the vast majority of containers simpler and easier, it is being proposed to make the above credentials optional, and only checking for a scratch image when those are provided. They would become required should the `--submit` flag also be given.

The flow would become:

* Check if credentials are provided.
* If not, run all checks
* If they are, check for scratch, and act accordingly.

This could mean that if a container is indeed being based on scratch, without credentials, the three checks mentioned above will not pass. So, a partner using a scratch image has two options:

1. Pass the credentials, with a project set OsContentType: scratch, and have the same experience as on submit.
2. Don't pass credentials, knowing that the experience upon submit will differ (but only in that certain checks will NOT be run)

The positive of this change would be that the experience for a non-scratch (again, likely the vast majority of containers), would result in an easier experience, as the lack of credentials would not change the experience of running checks, both with and without `--submit`.

### Other necessary changes

Additional information should be added to the three checks that shouldn't be run on a scratch image to indicate as such, and that if they are attempted against a scratch image, they are expected to fail.

komish:
For context, I believe this relates to https://github.com/redhat-openshift-ecosystem/openshift-preflight/issues/549.

komish:
> Making the parameters optional together with the scratch option, makes the above assumption incorrect.

To clarify, are you specifically referring to the fact that we remove checks for scratch images? Because success/failure should be identical otherwise, regardless of the value of the submit flag.

itroyano:
> To clarify, are you specifically referring to the fact that we remove checks for scratch images? Because success/failure should be identical otherwise, regardless of the value of the submit flag.

One specific user-unfriendly case of the above equation, is a user running his X image twice -
- first time G is skipped (without necessarily the user knowing it should run), resulting in a successful flow.
[...]

PR #558, opened 2022-04-11
Make the pulling of the project to check for scratch optional

### LINKED ISSUE #549: Is that possible to run Preflight tests (without --submit) without providing Pyxis project ID?
For Preflight 1.0.8, parameters `certification-project-id` and `pyxis-api-token` were not required to simply run the tests. Starting from 1.1.0-beta1 pre-release, these parameters become mandatory.

We face quite often a situation when the partners want first to simply run the tests, to estimate the work to do. Now, in such situations, we provide them with sort of fake or test credentials but that's additional work on our side. 

Would it be possible to make these parameters non-mandatory? I know that some tests, like BasedOnUbi, now require performing a Pyxis call. Is it possible to use some sort of default values internally in Preflight, just to make such calls? 

To make `certification-project-id` and `pyxis-api-token` to be optional when running without `--submit` option. 
If Pyxis calls are needed internally by Preflight, for example, to retrieve layerIds in the BasedOnUbi test, Preflight could use some sort of internal hardcoded credentials, with read-only permissions.

Now, in such situations, we provide our partners with sort of fake or test credentials but that's additional work on our side.

bcrochet:
In order to facilitate the running of preflight without the requirement
of a certification ID and API key, we will check if the certification
ID is provided. If it is not, we will skip checking the project before
running the checks. If --submit is given, the id and API key will become
required again.

Refers to #552
Fixes #549

### DIFF
--- a/cmd/check_container.go
+++ b/cmd/check_container.go
@@ -45,8 +45,18 @@ var checkContainerCmd = &cobra.Command{
 			return fmt.Errorf("%w: A container image positional argument is required", errors.ErrInsufficientPosArguments)
 		}
 
-		if submit && !viper.IsSet("dockerConfig") {
-			cmd.MarkFlagRequired("docker-config")
+		if submit {
+			if !viper.IsSet("dockerConfig") {
+				cmd.MarkFlagRequired("docker-config")
+			}
+
+			if !viper.IsSet("certification_project_id") {
+				cmd.MarkFlagRequired("certification-project-id")
+			}
+
+			if !viper.IsSet("pyxis_api_token") {
+				cmd.MarkFlagRequired("pyxis-api-token")
+			}
 		}
 
 		return nil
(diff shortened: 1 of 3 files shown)


acornett21 (commented):
Should the `recipes` be updated as part of this PR? If it's after the fact thats fine too. We also probably want to raise a PR [here](https://github.com/redhat-openshift-ecosystem/certification-releases/blob/main/containers/beta-workflow.md) as well.

cmd/check_container.go
```diff
@@ -66,41 +76,25 @@ var checkContainerCmd = &cobra.Command{
...
+				// Since we want the modified version, write it back
+				// to viper so that subsequent calls don't need to check
+				viper.Set("certification_project_id", projectId)
 			}
-			defer dockerConfigJsonFile.Close()
-
-			dockerConfigJsonBytes, err := io.ReadAll(dockerConfigJsonFile)
+			apiToken := viper.GetString("pyxis_api_token")
+			pyxisClient := pyxis.NewPyxisClient(viper.GetString("pyxis_host"), apiToken, projectId, &http.Client{Timeout: 60 * time.Second})
+			certProject, err := pyxisClient.GetProject(ctx)
```
komish, line 88: Does this also need to be wrapped in an `if submit { .. } ` block?

komish (changes requested):
Just a quick question ⬇️

cmd/check_container.go
bcrochet: No. This is by design. If you don't supply a projectId, it won't run. The only reason this block is running, is to check if it is a scratch container. So it doesn't have anything to do with submit.

acornett21 (approved):

cmd/check_container.go
komish: Might be an edge case, but if a user provides a project ID (for whatever reason), but doesn't pass the submit flag, we'd hit this block. If that same user provided nothing else (as in, no api token, docker config, etc., because they didn't want to submit), we'd hit error cases here (unable to hit Pyxis because the user's token wasn't provided). Am I misreading? Edge case too small?

I'm good either way.

bcrochet: That would happen, yes. It just means that GetProject will error out with a 401 when the API token is empty.
I could make it so that if certification-project-id is supplied, then api-token *must* be supplied. Then it would be caught before even getting here. As for docker config, that's moving to within the if submit at the end of the command.

acornett21: I don't think there is a way for `cobra` to say `flag-a` depends on `flag-b` but if that's a thing that would be nice to add upfront for this edge case.

bcrochet: Well, sort of. You have to do it yourself. Look at the Args section of the cmd. We're already doing it with --submit and docker-config.

komish: So this lets us check our container against scratch checks without requiring submit. Okay, I think that's a better benefit than the edge error case. LGTM

komish (approved):
/lgtm

Adam mentioned that our recipes might need updating, but that can be done as a follow up.

cmd/check_container.go
komish: So this lets us check our container against scratch checks without requiring submit. Okay, I think that's a better benefit than the edge error case. LGTM

jomkz (approved):

PR #559, opened 2022-04-11
Clarify the graphql layer checking logic

komish:
## Motivation

The BasedOnUBI check had a few opportunities where clarification could help future developers make changes, so this PR intends to clean some of those things.

The logic itself doesn't change much, but there are some method name changes for very specific reasons (detailed before). Otherwise, the code is functionally the same.

I think there's room to make further distinction between just "layers" and the uncompressed layer IDs that we're working with here, but I haven't figured out what makes the most sense there yet, so I'll have to revisit.

I mostly just wanted to change `CheckRedHatLayers` (see below).

## Change summary, by file

**certification/internal/policy/container/based_on_ubi.go**

- The comment describing the check was no longer correct. We perform a more granular check now than simply checking OS files to see if we're using the Red Hat UBI
- Several other method/function comments were put in place.
- The `CheckRedHatLayers` method has been renamed (starting at the interface, and all implementations I found). The `CheckRedHatLayers` name wasn't clear as to what we expected to get out of the method, so the new name `CertifiedImagesContainingLayers`, while longer, does include the inputs (Layers) and the output (CertifiedImages containing... well you get the idea).
- StaticCheck complained that checking `certImages` for nil before checking length is greater than 1 was redundant, so I complied.

**certification/pyxis/layers.go**

- Mostly comments here, and making sure the method name matched the interface's new definition.

**certification/pyxis/layers_test.go**
**certification/internal/policy/container/based_on_ubi_test.go**

- Nothing to note, other than making sure Fakes are in line.

### DIFF
--- a/certification/internal/policy/container/base_on_ubi.go
+++ b/certification/internal/policy/container/base_on_ubi.go
@@ -9,15 +9,13 @@ import (
 	log "github.com/sirupsen/logrus"
 )
 
-// BasedOnUBICheck evaluates if the provided image is based on the Red Hat Universal Base Image
-// by inspecting the contents of the `/etc/os-release` and identifying if the ID is `rhel` and the
-// Name value is `Red Hat Enterprise Linux`
+// BasedOnUBICheck evaluates if the provided image is based on the Red Hat Universal Base Image.
 type BasedOnUBICheck struct {
 	LayerHashCheckEngine layerHashChecker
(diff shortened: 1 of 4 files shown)


bcrochet (approved):
Doesn't really have much to do with GraphQL, but does make the code easier to read.

/lgtm

PR #564, opened 2022-04-14
Refactor and simplify some of the file reading

bcrochet:
This simplifies and refactors some of our file reading.

* Extracts all of the "read file, unmarshal" in builder to a common
  function
* Uses os.ReadFile() to get the docker config instead of open, read,
  close
* Use os.ReadFile() with Artifacts, and just get the length from the
  bytes read, instead of using io.Stat()

### DIFF
--- a/certification/pyxis/builder.go
+++ b/certification/pyxis/builder.go
@@ -4,7 +4,6 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"fmt"
-	"io"
 	"net/http"
 	"os"
 	"path"
@@ -105,20 +104,7 @@ func (b *certificationInputBuilder) WithRPMManifestFromFile(filepath string) *ce
 // in the submission. Multiple calls to this will append artifacts. Errors are logged,
 // but will not halt execution.
 func (b *certificationInputBuilder) WithArtifactFromFile(filepath string) *certificationInputBuilder {
-	file, err := os.Open(filepath)
-	if err != nil {
-		log.Error(err)
-		return b
-	}
-	defer file.Close()
-
-	bytes, err := io.ReadAll(file)
-	if err != nil {
-		log.Error(err)
-		return b
-	}
-
-	info, err := file.Stat()
+	bytes, err := os.ReadFile(filepath)
 	if err != nil {
 		log.Error(err)
 		return b
@@ -129,7 +115,7 @@ func (b *certificationInputBuilder) WithArtifactFromFile(filepath string) *certi
 		Content:     base64.StdEncoding.EncodeToString(bytes),
 		ContentType: http.DetectContentType(bytes),
 		Filename:    path.Base(filepath),
-		FileSize:    info.Size(),
+		FileSize:    int64(len(bytes)),
 	}
 
 	b.Artifacts = append(b.Artifacts, newArtifact)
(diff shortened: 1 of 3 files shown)


certification/pyxis/builder.go
```diff
@@ -150,43 +134,39 @@ func (b *certificationInputBuilder) storeRPMManifest(filepath string) error {
...
 
-	var manifest RPMManifest
-	err = json.Unmarshal(bytes, &manifest)
+	err = json.Unmarshal(bytes, &submission)
 	if err != nil {
 		return fmt.Errorf(
 			"%w: data for the %s appears to be malformed: %s",
 			errors.ErrSubmittingToPyxis,
-			"rpm manifest",
+			"rpm manifest", // USE REFLECTION - Worth it?
```
komish, line 142: Nope! 😆

bcrochet: Glad I put that in there! I'll take that out.

certification/pyxis/builder.go
```diff
@@ -129,17 +115,15 @@ func (b *certificationInputBuilder) WithArtifactFromFile(filepath string) *certi
 		Content:     base64.StdEncoding.EncodeToString(bytes),
 		ContentType: http.DetectContentType(bytes),
 		Filename:    path.Base(filepath),
-		FileSize:    info.Size(),
+		FileSize:    int64(len(bytes)),
```
komish, line 118: Perfect. I'm sure you tested as well, but documenting a self-contained test in case anyone else wants to try.

```
package main

import (
	"fmt"
	"os"
)

func main() {
	f := "file.txt"
	os.WriteFile(f, []byte("hello world\n"), 0755)

	file, _ := os.Open(f)
	info, _ := file.Stat()

	fileBytes, _ := os.ReadFile(f)
	fmt.Println("len bytes: ", int64(len(fileBytes)))
	fmt.Println("size from stat: ", info.Size())
}
```

certification/pyxis/builder.go
```diff
@@ -150,43 +134,38 @@ func (b *certificationInputBuilder) storeRPMManifest(filepath string) error {
...
+func (b *certificationInputBuilder) storeRPMManifest(filepath string) error {
+	var manifest RPMManifest
+	err := readAndUnmarshal(filepath, &manifest)
+	if err != nil {
+		return err
+	}
+
 	b.RpmManifest = &manifest
 	return nil
 }
```
komish, line 160: I'm not sure If I wanted to remove the string representing the asset we were working with here. The only reason I like it is because we do this same kind of logic multiple times, but with different files. In other words, there's a high likelihood that we'll see similar error messages if any of the instances that perform this kind of logic fail, so differentiating might be difficult.

Did you see somewhere else where we would be giving that little bit of context to a debugging developer that justifies not having it here?

komish (changes requested):
discussion points!

certification/pyxis/builder.go
bcrochet: Happy to write a unit test for this.

certification/pyxis/builder.go
bcrochet: Perhaps we should have the more elaborate error being creating in each of the store methods, rather than in the extracted logic. They would have that context. Or, we just need to add the context here... Let me play with that a bit.

bcrochet: Let me know what you think of the solution in the latest commit.

acornett21 (approved):

certification/pyxis/builder.go
komish: Hey! That works nicely. 👍

komish (approved):
/lgtm

PR #566, opened 2022-04-14
Prefer the usage of pyxis env instead of pyxis host

bcrochet:
Create a map, and use that map to lookup what the pyxis host and
path should be, rather than forcing the user to know that information.

If pyxis host is set, it will override the env. It is assumed that the
user knows what they are doing.

### DIFF
--- a/cmd/check_container.go
+++ b/cmd/check_container.go
@@ -73,6 +73,8 @@ var checkContainerCmd = &cobra.Command{
 			ResponseFormat: DefaultOutputFormat,
 		}
 
+		pyxisHost := pyxisHostLookup(viper.GetString("pyxis_env"), viper.GetString("pyxis_host"))
+
 		projectId := viper.GetString("certification_project_id")
 		if projectId != "" {
 			if strings.HasPrefix(projectId, "ospid-") {
@@ -82,7 +84,7 @@ var checkContainerCmd = &cobra.Command{
 				viper.Set("certification_project_id", projectId)
 			}
 			apiToken := viper.GetString("pyxis_api_token")
-			pyxisClient := pyxis.NewPyxisClient(viper.GetString("pyxis_host"), apiToken, projectId, &http.Client{Timeout: 60 * time.Second})
+			pyxisClient := pyxis.NewPyxisClient(pyxisHost, apiToken, projectId, &http.Client{Timeout: 60 * time.Second})
 			certProject, err := pyxisClient.GetProject(ctx)
 			if err != nil {
 				log.Error(fmt.Errorf("%w: %s", errors.ErrRetrievingProject, err))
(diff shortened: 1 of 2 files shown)


cmd/check_container.go
```diff
@@ -226,6 +228,19 @@ var checkContainerCmd = &cobra.Command{
...
+		"prod": "catalog.redhat.com/api/containers",
+		"uat":  "catalog.uat.redhat.com/api/containers",
+		"qa":   "catalog.qa.redhat.com/api/containers",
+	}
+	if hostOverride != "" {
+		return hostOverride
+	}
+
+	return envs[pyxisEnv]
+}
```
komish, line 242: I think `pyxisEnv` needs validation here, right? 

If the user providers an invalid value for pyxisEnv, I think we panic on the return because the key wouldn't exist in our map.

cmd/check_container.go
```diff
@@ -236,9 +251,12 @@ func init() {
 	checkContainerCmd.Flags().String("pyxis-api-token", "", "API token for Pyxis authentication")
 	viper.BindPFlag("pyxis_api_token", checkContainerCmd.Flags().Lookup("pyxis-api-token"))
 
-	checkContainerCmd.Flags().String("pyxis-host", certification.DefaultPyxisHost, "Host to use for Pyxis submissions.")
+	checkContainerCmd.Flags().String("pyxis-host", "", fmt.Sprintf("Host to use for Pyxis submissions.\nThis will override Pyxis Env."))
```
komish, line 254: - Should we add an example format for this value here?
- Should we extend the message to discourage use of this flag for most use cases?

komish (changes requested):
Few small things!

cmd/check_container.go
bcrochet: You're right. I wasn't sure what to do with that case. I guess just return default?

cmd/check_container.go
bcrochet: I almost want to just say: "Only set this if you know what you're doing"

jomkz (approved):

cmd/check_container.go
```diff
@@ -226,6 +228,24 @@ var checkContainerCmd = &cobra.Command{
...
+	}
+	if hostOverride != "" {
+		return hostOverride
+	}
+
+	pyxisHost, ok := envs[pyxisEnv]
+	if !ok {
+		pyxisHost = envs["prod"]
+	}
+	return pyxisHost
```
komish, line 246: /lgtm

Sorry, I totally didn't think about this before. I know the default behavior is to use prod, but if the user is overriding this, then in theory they don't want prod. Should we default to a non-prod environment of some kind in cases where the user specified this flag, but with an invalid value? 

Don't know if I like it (vs. throwing an error), but anyway. Marking this LGTM otherwise.

komish (approved):
/lgtm

One small comment, but you can ignore if you'd like.

cmd/check_container.go
acornett21: In practice, I would agree with you @komish ie throw the error. For this use-case where there is `internal` and `external` hosts, I like having this override so as a dev team we can see if the issues is with Pyxis itself or just the `external` instance we call.

acornett21 (approved):

PR #567, opened 2022-04-14
Simplify the install of go tools

bcrochet:
This strips out some of the hacky 'make a temp module' stuff that was
necessary prior to Go 1.17. This still wraps in a function, but only
does 'go install' and only when it needs to.


jomkz (approved):

komish (approved):
/lgtm

acornett21 (approved):

Issue #568, opened 2022-04-14
Run `RunAsNonRoot` Check Should Be Conditionally Applied


acornett21:
## Is your feature request related to a problem? Please describe.
The `RunAsNonRoot` check should be conditionally applied based on the `certProject.container.privileged` that comes from pyxis.

## Describe the solution you'd like.
We exclude this check if `certProject.container.privileged = true` when we call pyxis during `submit`

## Describe alternatives you've considered.
N/A

## Additional context.
This was the ask from the business, so we will go with this direction for milestone 1.

PR #569, opened 2022-04-14
adding in RootExceptionContainerPolicy check list

acornett21:
- Fixes: #568 
- Adding in RootExceptionContainerPolicy which is to be used to exclude RunAsNonRoot check for projects that have Host level access set to `Privileged` in connect project setup
- The above value is determined from pyxis `certProject.container.privileged` value

## Connect Setup
-  https://connect.uat.redhat.com/projects/6258a3c3170e7478b8eab726/settings
image

## Test Results
- https://connect.uat.redhat.com/projects/6258a3c3170e7478b8eab726/images/6258a52a170e7478b8eab727/scan-results (The `BasedOnUBI` error can be ignored, Pyxis is currently having issues and will retest on Monday)
image

You will notice no `RunAsNonRootCheck`

### DIFF
--- a/certification/engine/engine.go
+++ b/certification/engine/engine.go
@@ -137,6 +137,18 @@ var scratchContainerPolicy = map[string]certification.Check{
 	// runSystemContainerCheck.Name(): runSystemContainerCheck,
 }
 
+var rootExceptionContainerPolicy = map[string]certification.Check{
+	hasLicenseCheck.Name():        hasLicenseCheck,
+	hasUniqueTagCheck.Name():      hasUniqueTagCheck,
+	maxLayersCheck.Name():         maxLayersCheck,
+	hasNoProhibitedCheck.Name():   hasNoProhibitedCheck,
+	hasRequiredLabelsCheck.Name(): hasRequiredLabelsCheck,
+	basedOnUbiCheck.Name():        basedOnUbiCheck,
+	hasModifiedFilesCheck.Name():  hasModifiedFilesCheck,
+	// runnableContainerCheck.Name():  runnableContainerCheck,
+	// runSystemContainerCheck.Name(): runSystemContainerCheck,
+}
+
 func makeCheckList(checkMap map[string]certification.Check) []string {
 	checks := make([]string, 0, len(checkMap))
 
@@ -157,4 +169,8 @@ func ContainerPolicy() []string {
 
 func ScratchContainerPolicy() []string {
 	return makeCheckList(scratchContainerPolicy)
+}
+
+func RootExceptionContainerPolicy() []string {
+	return makeCheckList(rootExceptionContainerPolicy)
 }
--- a/cmd/check_container.go
+++ b/cmd/check_container.go
@@ -95,6 +95,11 @@ var checkContainerCmd = &cobra.Command{
 				cfg.EnabledChecks = engine.ScratchContainerPolicy()
 				cfg.Scratch = true
 			}
+
+			// if a partner sets `Host Level Access` in connect to `Privileged`, enable RootExceptionContainerPolicy checks
+			if certProject.Container.Privileged {
+				cfg.EnabledChecks = engine.RootExceptionContainerPolicy()
+			}
 		}
 
 		engine, err := engine.NewForConfig(cfg)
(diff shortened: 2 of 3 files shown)


test/containerfiles/container-passes-privileged.Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM registry.access.redhat.com/ubi8/ubi:latest
```
bcrochet, line 1: Everything else looks good, but I'm not sure calling this this 'container-passes' is necessarily correct. Because it won't pass without the Pyxis call and removing RunsAsRoot, right?

certification/engine/engine.go
```diff
@@ -158,3 +170,7 @@ func ContainerPolicy() []string {
 func ScratchContainerPolicy() []string {
 	return makeCheckList(scratchContainerPolicy)
 }
+
+func PrivilegedContainerPolicy() []string {
```
bcrochet, line 174: Also, let's not name this "Privileged". They are overloading it in Pyxis, but this has a completely different connotation. We are actually checking running as root, not for --privileged. They have different meanings. I think just RootExceptionContainerPolicy should do the trick.

test/containerfiles/container-passes-privileged.Dockerfile
skattoju4: I guess its to test that RunAsRoot was skipped ?

acornett21: @bcrochet That's true, I wasn't really sure what to call it. I just wanted to show the container that I used to test that went along with the screen shot shared. Thoughts on naming it similar to your suggestion below `container-passes-with-root-exception`? Or is it better to not include this container in the codebase?

bcrochet: I don't think we need to add this file at all, as you say. This case can be covered by the container-fails image.

acornett21: I ended up removing this, I don't think anyone else will use this container in the future, but was needed to show the tests I did.

certification/engine/engine.go
acornett21: Change the name to the one suggested above.

certification/runtime/runtime.go
```diff
@@ -15,6 +15,7 @@ type Config struct {
 	Mounted        bool
 	Bundle         bool
 	Scratch        bool
+	Privileged     bool
```
komish, line 18: I might be missing it; what does this do for us?

I don't mind it being here... because at some point I think we should migrate back to a single-point-of-configuration type workflow, which doesn't exist. But we haven't had conversations about that, and I'm not sure I'm seeing where this is used.

cmd/check_container.go
```diff
@@ -93,6 +93,12 @@ var checkContainerCmd = &cobra.Command{
 				cfg.EnabledChecks = engine.ScratchContainerPolicy()
 				cfg.Scratch = true
 			}
+
+			// if a partner sets `Host Level Access` in connect to `Privileged`, enable RootExceptionContainerPolicy checks
+			if certProject.Container.Privileged {
+				cfg.EnabledChecks = engine.RootExceptionContainerPolicy()
+				cfg.Privileged = true
```
komish, line 100: Related to my other comment, I see this is being set, but I'm not sure where it's being used.

Should we be setting this configuration and then letting the engine.NewForConfig function give us the right policy?

komish (approved):
Few questions. Nothing to change urgently, so I'm going to mark this as approved.

certification/runtime/runtime.go
acornett21: Right now we aren't using this, I feel its better to add it now incase we need to use it. But if others feel it can be removed, that's fine as well.

cmd/check_container.go
acornett21: We aren't using this value today, but maybe in the future it's need.

certification/runtime/runtime.go
bcrochet: I would say leave it out. We did this for scratch, because we needed to skip the rpmmanifest. We don't need to know about it in the engine.

cmd/check_container.go
bcrochet: Leave it out.

certification/runtime/runtime.go
acornett21: I removed this.

bcrochet (approved):
/lgtm

itroyano:
Ltgm.  Maybe a bit of extra logging to indicate that's the set of checks running.

PR #574, opened 2022-04-20
updating recipes file to include a way to run `check container`

acornett21:
- Adding in a recipe for a user to be able to run `check container` via podman
- Also updated `check operator` recipe so to more closely match `check container1


bcrochet (approved):
/lgtm

jomkz (approved):

Issue #519, opened 2022-03-25
Update all `check container` KnowledgeBaseURL and CheckURL in every Check


acornett21:
## Is your feature request related to a problem? Please describe.
Before `check container` goes GA, the documentation is moving to a new location, so we need to update all of our URL's to point to the new location.

## Describe the solution you'd like.
Update both `KnowledgeBaseURL` and `CheckURL` in all of the `check container` checks.

## Describe alternatives you've considered.
N/A

## Additional context.

`KnowledgeBaseURL` is a URL detailing how to resolve a check failure.
`CheckURL` is a URL pointing to the official policy documentation from Red Hat, containing information on exactly what is being tested and why.

acornett21:
@taylorjordanNC Please update this issue once you have the production URL and documentation posted. Thanks!

jomkz:
This should use the current policy guide:

https://access.redhat.com/documentation/en-us/red_hat_openshift_certification/4.9/html-single/red_hat_openshift_software_certification_policy_guide/index#assembly-requirements-for-container-images_openshift-sw-cert-policy-introduction

acornett21:
@jomkz Is that what we want to go with? I thought something more detailed was getting written with our checknames and such?

taylorjordanNC:
That's the URL but the changes just haven't been pushed yet to show the increased detail - but let's use this URL.

jomkz:
In addition, they will likely be changing to a URL that is not OpenShift release specific so that we don’t have to change it for every minor release but we can use this for now.

acornett21:
@jomkz Sounds good, PR for that above is open.

PR #581, opened 2022-04-22
updating check container check url in all checks

acornett21:
- Fixes: #519 
- Updating all url's in all checks to point to cert policy docs


bcrochet (changes requested):
Can we put that in a constant somewhere?

jomkz:
> Can we put that in a constant somewhere?

+1 perhaps use that as the default for in the future when a check specific URL/anchor isn’t available

acornett21:
> should we deprecate one of those since both point to the same thing ?

There are different URL's for someone of the `check operator` check. Also, these values get sent to pyxis for the partner to view, so to not break anything in pyxis or connect, it's best that we keep both.

acornett21:
@bcrochet Moved this to package level constant.

bcrochet (approved):
/lgtm

jomkz (approved):

PR #587, opened 2022-04-26
updating vagrant file to change the port

acornett21:
- updating the vagrant file so that prelfight vm and CRC can run at the same time. CRC runs on the default port of `2222` and so does `vagrant` since CRC is harder to configure, updating here.


jomkz (approved):

bcrochet (approved):
/lgtm

PR #588, opened 2022-04-27
Use --docker-config for crane if specified

bcrochet:
Currently, preflight only uses the standard Docker DOCKER_CONFIG,
docker file (.docker/config.json) or podman's
$XDG_RUNTIME_DIR/containers/auth.json when pulling the image under
test. Now, if --docker-config is specified (or PFLT_DOCKERCONFIG),
preflight will attempt to use that file for authentication. It will
not fall back to the others. An authfile must be explicit.

### DIFF
--- a/certification/internal/authn/keychain.go
+++ b/certification/internal/authn/keychain.go
@@ -1,0 +1,77 @@
+package authn
+
+import (
+	"os"
+
+	"github.com/docker/cli/cli/config"
+	"github.com/docker/cli/cli/config/types"
+	craneauthn "github.com/google/go-containerregistry/pkg/authn"
+	"github.com/google/go-containerregistry/pkg/name"
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/viper"
+)
+
+type preflightKeychain struct{}
+
+var PreflightKeychain craneauthn.Keychain = &preflightKeychain{}
+
+// Resolve returns an Authenticator with credentials, or Anonymous if no suitable credentials
+// are found for the target. This implements the Keychain interface from go-containerregistry,
+// and will be passed to crane,.
+//
+// If the viper config is empty, assume Anonymous.
+// If the file cannot be found or read, that constitues an error.
+func (k *preflightKeychain) Resolve(target craneauthn.Resource) (craneauthn.Authenticator, error) {
+	log.Trace("entering preflight keychain Resolve")
+
+	configFile := viper.GetString("dockerConfig")
+	if configFile == "" {
+		// No file specified. No auth expected
+		return craneauthn.Anonymous, nil
+	}
+
+	r, err := os.Open(configFile)
+	if os.IsNotExist(err) {
+		log.Errorf("could not find authfile: %s", configFile)
+		return nil, err
+	}
+	if err != nil {
+		log.Errorf("Could not open authfile: %s", configFile)
+		return nil, err
+	}
+	defer r.Close()
+	cf, err := config.LoadFromReader(r)
+	if err != nil {
+		log.Errorf("Could not load authfile: %s", configFile)
+		return nil, err
+	}
+
+	var cfg, empty types.AuthConfig
+	for _, key := range []string{
+		target.String(),
+		target.RegistryStr(),
+	} {
+		if key == name.DefaultRegistry {
+			key = craneauthn.DefaultAuthKey
+		}
+
+		cfg, err = cf.GetAuthConfig(key)
+		if err != nil {
+			return nil, err
+		}
+		if cfg != empty {
+			break
+		}
+	}
+	if cfg == empty {
+		return craneauthn.Anonymous, nil
+	}
+
+	return craneauthn.FromConfig(craneauthn.AuthConfig{
+		Username:      cfg.Username,
+		Password:      cfg.Password,
+		Auth:          cfg.Auth,
+		IdentityToken: cfg.IdentityToken,
+		RegistryToken: cfg.RegistryToken,
+	}), nil
+}
--- a/certification/internal/authn/keychain_test.go
+++ b/certification/internal/authn/keychain_test.go
@@ -1,0 +1,209 @@
+// Copyright 2018 Google LLC All Rights Reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package authn
+
+import (
+	"encoding/base64"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"reflect"
+	"testing"
+
+	craneauthn "github.com/google/go-containerregistry/pkg/authn"
+	"github.com/google/go-containerregistry/pkg/name"
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/viper"
+)
+
+var (
+	fresh              = 0
+	testRegistry, _    = name.NewRegistry("test.io", name.WeakValidation)
+	testRepo, _        = name.NewRepository("test.io/my-repo", name.WeakValidation)
+	defaultRegistry, _ = name.NewRegistry(name.DefaultRegistry, name.WeakValidation)
+)
+
+// setupConfigDir sets up an isolated configDir() for this test.
+func setupConfigDir(t *testing.T) string {
+	tmpdir := os.Getenv("TEST_TMPDIR")
+	if tmpdir == "" {
+		var err error
+		tmpdir, err = ioutil.TempDir("", "keychain_test")
+		if err != nil {
+			t.Fatalf("creating temp dir: %v", err)
+		}
+	}
+
+	fresh++
+	p := filepath.Join(tmpdir, fmt.Sprintf("%d", fresh))
+	if err := os.Mkdir(p, 0o777); err != nil {
+		t.Fatalf("mkdir %q: %v", p, err)
+	}
+	return p
+}
+
+func setupConfigFile(t *testing.T, content string) string {
+	cd := setupConfigDir(t)
+	p := filepath.Join(cd, "config.json")
+	if err := ioutil.WriteFile(p, []byte(content), 0o600); err != nil {
+		t.Fatalf("write %q: %v", p, err)
+	}
+	os.Setenv("PFLT_DOCKERCONFIG", p)
+	// return the config dir so we can clean up
+	return cd
+}
+
+func TestNoConfig(t *testing.T) {
+	cd := setupConfigDir(t)
+	defer os.RemoveAll(filepath.Dir(cd))
+
+	auth, err := PreflightKeychain.Resolve(testRegistry)
+	if err != nil {
+		t.Fatalf("Resolve() = %v", err)
+	}
+
+	if auth != craneauthn.Anonymous {
+		t.Errorf("expected Anonymous, got %v", auth)
+	}
+}
+
+func encode(user, pass string) string {
+	delimited := fmt.Sprintf("%s:%s", user, pass)
+	return base64.StdEncoding.EncodeToString([]byte(delimited))
+}
+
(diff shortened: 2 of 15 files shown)


docs/RECIPES.md
```diff
@@ -155,6 +155,10 @@ You will also need:
 - A Partner Connect API Key
   - An API Key can be created in Red Hat Partner Connect at the following [URL](https://connect.redhat.com/account/api-keys)
   - Required for submit
+- A Docker Auth file, can be specified multiple ways
+  - If PFLT_DOCKERCONFIG envvar is set, or --docker-config is passed on the command line, this will be used to pull images.
```
acornett21, line 159: Nothing is changing how we read in the file we send to `pyxis` correct? Do we need to note that for submit this must be set? Or do you think that's already implied by the cmd checking for the flag/env and erroring on `--submit` if not provided? I'm fine either way, just want clarity.

certification/internal/authn/keychain.go
```diff
@@ -0,0 +1,68 @@
...
+	"github.com/spf13/viper"
+)
+
+type preflightKeychain struct {
+	mu sync.Mutex
+}
+
+var PreflightKeychain craneauthn.Keychain = &preflightKeychain{}
+
+func (k *preflightKeychain) Resolve(target craneauthn.Resource) (craneauthn.Authenticator, error) {
```
acornett21, line 20: I see that this resembles `main` of `crane` which I think is smart, however in the version we are referencing as a dependency the code looks a little different, see [here](https://github.com/google/go-containerregistry/blob/v0.8.0/pkg/authn/keychain.go#L119-L122).

Specifically, I've experience this weird issue (for a lack of a better term) with preflight and docker registry. Do you think we need to account for the above in our code as well? Or do you feel since `main` got ride of this code that it's handled elsewhere in their dependencies and we should also be fine? Or should we _also_ update preflight to use `main` so it's more of a one-to-one? If this is unclear please let me know.

acornett21 (commented):
At face value I think the code makes sense to me, and see no specific changes needed, however I have some questions below.

docs/RECIPES.md
bcrochet: Obviously this comment is out of date.

acornett21: Yeah this was from when both env's were listed I will resolve this comment.

certification/internal/authn/keychain.go
bcrochet: Added.

acornett21:
The `Recipes` file will need to be updated as part of this change as well. Specifically the `podman` example I believe.

jomkz (approved):

tkrishtop:
Hi @bcrochet the change is failing in disconnected env for the "private" operator (simple-demo-operator in the quay.io with restricted access): https://www.distributed-ci.io/jobs/960ad924-8404-4d89-b45b-81a23be4d505/jobStates#2590f59b-c27b-4bb7-8353-537d8c60e275:file87 

Here are the logs:
```
time="2022-04-27T22:20:36Z" level=debug msg="config file not found, proceeding without it"
time="2022-04-27T22:20:36Z" level=info msg="certification library version pr588 <commit: 0c3ed31319b0868a87fc3258e608296b7df9e08a>"
time="2022-04-27T22:20:36Z" level=debug msg="target image: quay.io/telcoci/simple-demo-operator-bundle@sha256:6cfbca9b14a51143cfc5d0d56494e7f26ad1cd3e662eedd2bcbebf207af59c86"
time="2022-04-27T22:20:36Z" level=debug msg="pulling image from target registry"
time="2022-04-27T22:20:36Z" level=trace msg="entering preflight keychain Resolve"
time="2022-04-27T22:20:36Z" level=fatal msg="failed to pull remote container: GET https://quay.io/v2/telcoci/simple-demo-operator-bundle/manifests/sha256:6cfbca9b14a51143cfc5d0d56494e7f26ad1cd3e662eedd2bcbebf207af59c86: UNAUTHORIZED: access to the requested resource is not authorized; map[]"
``` 

I'm sorry that failed status was not displayed in the last job, I'm going to add a failure if results.json is empty.

bcrochet:
> Hi @bcrochet the change is failing in disconnected env for the "private" operator (simple-demo-operator in the quay.io with restricted access): https://www.distributed-ci.io/jobs/960ad924-8404-4d89-b45b-81a23be4d505/jobStates#2590f59b-c27b-4bb7-8353-537d8c60e275:file87
> 
> Here are the logs:
> 
> ```
> time="2022-04-27T22:20:36Z" level=debug msg="config file not found, proceeding without it"
> time="2022-04-27T22:20:36Z" level=info msg="certification library version pr588 <commit: 0c3ed31319b0868a87fc3258e608296b7df9e08a>"
> time="2022-04-27T22:20:36Z" level=debug msg="target image: quay.io/telcoci/simple-demo-operator-bundle@sha256:6cfbca9b14a51143cfc5d0d56494e7f26ad1cd3e662eedd2bcbebf207af59c86"
> time="2022-04-27T22:20:36Z" level=debug msg="pulling image from target registry"
> time="2022-04-27T22:20:36Z" level=trace msg="entering preflight keychain Resolve"
> time="2022-04-27T22:20:36Z" level=fatal msg="failed to pull remote container: GET https://quay.io/v2/telcoci/simple-demo-operator-bundle/manifests/sha256:6cfbca9b14a51143cfc5d0d56494e7f26ad1cd3e662eedd2bcbebf207af59c86: UNAUTHORIZED: access to the requested resource is not authorized; map[]"
> ```
> 
> I'm sorry that failed status was not displayed in the last job, I'm going to add a failure if results.json is empty.

Thanks for that. This does have the side-effect on the operator side. I'll need to address that.

docs/RECIPES.md
```diff
@@ -198,7 +202,6 @@ $CONTAINER_TOOL run \
   --env PFLT_CERTIFICATION_PROJECT_ID=1234567890a987654321bcde \
   --env PFLT_PYXIS_API_TOKEN=abcdefghijklmnopqrstuvwxyz123456 \
   --env PFLT_DOCKERCONFIG=/temp-authfile.json \
-  --env DOCKER_CONFIG=/tmp/docker \
   -v /some/path/on/your/host/artifacts:/artifacts \
   -v ./temp-authfile.json:/temp-authfile.json:ro \
   -v ./temp-authfile.json:/tmp/docker/config.json:ro \
```
acornett21, line 207: This line can also be removed since we are no longer taking in the `DOCKER-CONFIG` arg.

tkrishtop:
> @tkrishtop This PR also means that if you are using a private registry that requires authentication, PFLT_DOCKERCONFIG/--docker-config will be mandatory. preflight will no longer utilize any of the "standard" authfile locations. DOCKER_CONFIG is no longer used. Your test is going to have to change.

Hi @bcrochet, thank you for the explanation and debug! I modified the tests following your recommendations, the job is now successful: https://www.distributed-ci.io/jobs/bfc4ca15-7577-48bd-8401-0db600d5e7be/tests 
I will merge and deploy the changes in DCI/Preflight integration tomorrow.

certification/internal/authn/keychain.go
```diff
@@ -0,0 +1,81 @@
...
+	"github.com/docker/cli/cli/config"
+	"github.com/docker/cli/cli/config/types"
+	craneauthn "github.com/google/go-containerregistry/pkg/authn"
+	"github.com/google/go-containerregistry/pkg/name"
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/viper"
+)
+
+type preflightKeychain struct {
+	mu sync.Mutex
```
komish, line 16: Pardon my ignorance, but why do we need mutual exclusion here?

The only reason I ask is because this feels like a _lot_ of code for what is otherwise just "read creds or assume anonymous". 

We don't run checks concurrently, but even if we did, I'm of the opinion that the credentials in use should _never_ change within the context of check execution, and should only ever change at the outset of execution (e.g. initializing everything before we start our core logic).

certification/internal/authn/keychain.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+var PreflightKeychain craneauthn.Keychain = &preflightKeychain{}
```
komish, line 19: Should we stuffing the authentication context into `context` using `WithValue()?`, which is already getting passed around everywhere? As opposed to using a global here?

komish (commented):
I'm still doing some review here, but here's some starting points for discussion.

certification/internal/authn/keychain.go
```diff
@@ -0,0 +1,81 @@
...
+
+func (k *preflightKeychain) Resolve(target craneauthn.Resource) (craneauthn.Authenticator, error) {
```
komish, line 21: This needs a comment detailing the "why" and "how".

certification/internal/authn/keychain.go
```diff
@@ -0,0 +1,81 @@
...
+
+	if _, err := os.Stat(configFile); err != nil {
+		log.Errorf("Could not find file: %s", configFile)
+		return nil, err
+	}
+
+	r, err := os.Open(configFile)
+	if os.IsNotExist(err) {
+		return craneauthn.Anonymous, nil
+	}
```
komish, line 41: Is `os.Open` resulting in an `IsNotExist` error cover roughly the same goal as running `os.Stat`? In other words, can we scrap the Stat call?

Otherwise, why do the two error cases for each call return a different outcome? Trying to put together why Stat would find it, but os.Open not finding it means we get Anonymous auth.

certification/internal/authn/keychain.go
bcrochet: Complete honesty: This was cargo-culted from the DefaultKeychain in go-containerregistry.

certification/internal/authn/keychain.go
bcrochet: I actually had that thought. I think it probably is the best way to handle it. If certain commands require different keychains, it would simplify it.

certification/internal/authn/keychain.go
bcrochet: Let me think about this one a bit more. Your intuition is probably correct.

certification/internal/authn/keychain.go
bcrochet: Let's punt on this one for a more comprehensive PR on just this subject.

certification/internal/authn/keychain.go
bcrochet: Removed.

certification/internal/authn/keychain.go
bcrochet: Definitely correct. Fixed.

certification/internal/authn/keychain.go
bcrochet: Done. LMKWYT

certification/internal/authn/keychain_test.go
```diff
@@ -0,0 +1,223 @@
...
+	tmp, err := ioutil.TempDir("", "keychain_test_home")
+	if err != nil {
+		log.Fatal(err)
+	}
+	os.Setenv("HOME", tmp)
+	os.Exit(func() int {
+		defer os.RemoveAll(tmp)
+		return m.Run()
+	}())
+}
```
komish, line 51: `TestMain` seems odd. I would read this as you're testing the `Main()`, which I don't believe is what's happening here. General convention maps gotest functions to look something like `Test<FunctionUnderTest><Permutation>`. I won't say we should strictly follow that convention, but maybe something other than Main here would be sufficient.

komish: Is resetting HOME here necessary? The doc for WithAuthFromKeychain seems to imply that the default keychain gets overridden with either your keychain or anonymous just by invoking this option https://github.com/google/go-containerregistry/blob/main/pkg/crane/options.go#L87

komish (changes requested):
Just one more thing here (related to the test functions)

Aside: I strongly dislike the various places where we have to pass the same crane options. We should probably add that to the short list of things to refactor in the future.

certification/internal/authn/keychain_test.go
bcrochet: Yeah, I can change TestMain to something more meaningful. This file was copied directly from go-containerregistry. As for the $HOME thing, no, it doesn't need to be there, since we aren't using that anyway.

bcrochet: After removing the HOME reassignment, TestMain was no longer necessary.

bcrochet:
> Aside: I strongly dislike the various places where we have to pass the same crane options. We should probably add that to the short list of things to refactor in the future.

Absolutely. It's next on my list.

komish (approved):
/lgtm

PR #591, opened 2022-04-28
Copy release images to redhat-isv on quay

komish:
This PR introduces an action that's invoked via workflow call. 

It waits for build-release and build-multiarch to complete, and then copy all of those images to the redhat-isv organization in Quay.

This should push a little more than we want (e.g. rc, betas, etc), but we don't automatically use these images in pipelines. To use it in pipelines, we must submit a PR to the pipeline to point to the new images in redhat-isv. From that perspective, adding extra images to the org should cause no problems.

### DIFF
--- a/.github/workflows/build-release.yml
+++ b/.github/workflows/build-release.yml
@@ -77,3 +77,16 @@ jobs:
     secrets:
       registry: ${{ secrets.IMAGE_REGISTRY }}
       token: ${{ secrets.GITHUB_TOKEN }}
+
+  copy-to-rhisv:
+    needs: [build-release, build-multiarch]
+    uses: ./.github/workflows/copy-to-rhisv.yml
+    with:
+      sourceImageName: ${{ needs.build-release.outputs.imageName }}
+      sourceImageTag: ${{ needs.build-release.outputs.imageVersion }}
+      destImageName: preflight-test
+    secrets:
+      sourceImageRegistry: ${{ secrets.IMAGE_REGISTRY }}
+      destImageRegistry: ${{ secrets.RHISV_IMAGE_REGISTRY }}
+      destRegistryUser: ${{ secrets.RHISV_REGISTRY_USER }}
+      destRegistryPassword: ${{ secrets.RHISV_REGISTRY_PASSWORD }}
--- a/.github/workflows/copy-to-rhisv.yml
+++ b/.github/workflows/copy-to-rhisv.yml
@@ -1,0 +1,56 @@
+name: Mirror images to RHISV Quay Org
+
+on:
+  workflow_call:
+    inputs:
+      sourceImageName:
+        required: true
+        type: string
+      sourceImageTag:
+        required: true
+        type: string
+      destImageName:
+        required: true
+        type: string
+    secrets:
+      sourceImageRegistry:
+        required: true
+      destImageRegistry:
+        required: true
+      destRegistryUser:
+        required: true
+      destRegistryPassword:
+        required: true
+
+jobs:
+  mirror-images-to-rhisv:
+    name: Mirror Images
+    runs-on: ubuntu-latest
+    steps:
+    - name: Set up Go
+      uses: actions/setup-go@v2
+      with:
+        go-version: 1.17
+
+    - name: Install Skopeo
+      run: |
+        sudo apt-get update && sudo apt-get install libgpgme-dev libdevmapper-dev libbtrfs-dev -y
+        go install github.com/containers/skopeo/cmd/skopeo@v1.7.0
+        skopeo -v
+
+    - name: Podman Login
+      uses: redhat-actions/podman-login@v1
+      with:
+        registry: ${{ secrets.destImageRegistry }}
+        username: ${{ secrets.destRegistryUser }}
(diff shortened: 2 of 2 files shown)


.github/workflows/build-release.yml
```diff
@@ -77,3 +77,12 @@ jobs:
     secrets:
       registry: ${{ secrets.IMAGE_REGISTRY }}
       token: ${{ secrets.GITHUB_TOKEN }}
+
+  copy-to-rhisv:
+    needs: [build-release, build-multiarch]
```
bcrochet, line 82: Should only need build-multiarch here. Why both?

.github/workflows/copy-to-rhisv.yml
```diff
@@ -0,0 +1,39 @@
...
+jobs:
+  mirror-images-to-rhisv:
+    needs: []
+    name: Mirror Images
+    runs-on: ubuntu-latest
+    steps:
+    - name: Podman Login
+      uses: redhat-actions/podman-login@v1
+      with:
+        registry: ${{ secrets.RHISV_IMAGE_REGISTRY }}
```
bcrochet, line 28: Probably best to pass these in as an input secret? Rather than refer to secrets directly?

bcrochet (changes requested):
Couple of comments.

.github/workflows/build-release.yml
komish: I need the inputs from build-release. The build-multiarch target does not produce outputs.

bcrochet: Ah. Hadn't run into that scenario before.

.github/workflows/copy-to-rhisv.yml
bcrochet: I think I prefer to see them as an input. Otherwise, you'd have to look in 2 places to know which secrets to set. As parameters, then you can see in the main workflow what needs to be set. Not a showstopper if you feel otherwise.

bcrochet: Oh, and it makes this more reusable, should we need to add further locations.

.github/workflows/build-release.yml
```diff
@@ -77,3 +77,12 @@ jobs:
...
+    uses: ./.github/workflows/copy-to-rhisv.yml
+    with:
+      sourceImageName: ${{ needs.build-release.outputs.imageName }}
+      sourceImageTag: ${{ needs.build-release.outputs.imageVersion }}
+    secrets:
+      sourceImageRegistry: ${{ secrets.IMAGE_REGISTRY }}
```
acornett21, line 88: New line please and thanks!

.github/workflows/copy-to-rhisv.yml
```diff
@@ -0,0 +1,39 @@
...
+        skopeo copy --all --preserve-digests ${{ secrets.sourceImageRegistry }}/${{ inputs.sourceImageName }}:${{ inputs.sourceImageTag }}-linux-ppc64le ${{ secrets.RHISV_IMAGE_REGISTRY }}/${{ env.DEST_IMAGE_NAME }}:${{ inputs.sourceImageTag }}-linux-ppc64le
+        skopeo copy --all --preserve-digests ${{ secrets.sourceImageRegistry }}/${{ inputs.sourceImageName }}:${{ inputs.sourceImageTag }}-linux-arm64 ${{ secrets.RHISV_IMAGE_REGISTRY }}/${{ env.DEST_IMAGE_NAME }}:${{ inputs.sourceImageTag }}-linux-arm64
+        skopeo copy --all --preserve-digests ${{ secrets.sourceImageRegistry }}/${{ inputs.sourceImageName }}:${{ inputs.sourceImageTag }}-linux-s390x ${{ secrets.RHISV_IMAGE_REGISTRY }}/${{ env.DEST_IMAGE_NAME }}:${{ inputs.sourceImageTag }}-linux-s390x
+        skopeo copy --all --preserve-digests ${{ secrets.sourceImageRegistry }}/${{ inputs.sourceImageName }}:${{ inputs.sourceImageTag }} ${{ secrets.RHISV_IMAGE_REGISTRY }}/${{ env.DEST_IMAGE_NAME }}:${{ inputs.sourceImageTag }}
```
acornett21, line 39: New line here as well.

.github/workflows/copy-to-rhisv.yml
komish: Fair enough, reconfigured this to use inputs almost entirely

.github/workflows/build-release.yml
komish: Fixed

.github/workflows/copy-to-rhisv.yml
komish: fixed

bcrochet (approved):
/lgtm

jomkz (approved):

komish:
Alright folks, here's the corresponding pipeline run that tests this PR in my fork:

https://github.com/komish/openshift-preflight/actions/runs/2246117575

The corresponding repositories are here
https://quay.io/repository/komish/preflight-test (destination)
https://quay.io/repository/komish/preflight (source)

PTAL

PR #592, opened 2022-05-02
Fix a data race exposed by race checker

bcrochet:
When adding the -race flag to go test along with -cover, a couple of
data races were exposed by the tool. This fixes those data races, and
introduces 'make cover' as a target in the Makefile.

The wg.Wait() was being executed before anything as added to the
WaitGroup. Therefore, it was possible that the channel might close
before any work was done. This makes sure that the WaitGroup has
something to wait on before calling Wait.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,6 +48,10 @@ image-push:
 test:
 	go test -v $$(go list ./... | grep -v e2e)
 
+.PHONY: cover
+cover:
+	go test -v $$(go list ./... | grep -v e2e) -race -cover -coverprofile=coverage.out
+
 .PHONY: vet
 vet:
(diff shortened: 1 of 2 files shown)


acornett21 (approved):

komish (approved):
/lgtm

komish:
Not sure if we care to commit the coverage.out, but if we don't, add it to the gitignore.

acornett21:
> Not sure if we care to commit the coverage.out, but if we don't, add it to the gitignore.

It's already in the ignore file [here](https://github.com/redhat-openshift-ecosystem/openshift-preflight/blob/main/.gitignore#L5)

PR #595, opened 2022-05-02
Run tests on release branches

bcrochet:
Since we have release branches now, it is a good idea to run tests
on those branches so that when we cherry-pick to them, it will run
these tests.

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -2,7 +2,9 @@ name: Go
 
 on:
   pull_request:
-    branches: [ main ]
+    branches:
+    - main
+    - release-*
 
 jobs:
   build:


acornett21 (approved):

.github/workflows/build-main.yml
```diff
@@ -2,7 +2,10 @@ name: Build Main
 
 on:
   push:
-    branches: [ main, ghactions ]
+    branches:
+    - main
+    - ghactions
+    - release-*
```
itroyano, line 8: I'm not sure if the tagging here - alpha sha - would work for a release branch.

jomkz (approved):

komish:
We probably also want to enable E2E tests for the release branch. 

Here's what we run against main:

https://github.com/openshift/release/blob/master/ci-operator/config/redhat-openshift-ecosystem/openshift-preflight/redhat-openshift-ecosystem-openshift-preflight-main__4.8.yaml

I can update this unless you want to take a stab at it.

komish:
Also, do we want to run `build-main.yaml`? Do we want published images of the release branch?

bcrochet:
> We probably also want to enable E2E tests for the release branch.
> 
> Here's what we run against main:
> 
> https://github.com/openshift/release/blob/master/ci-operator/config/redhat-openshift-ecosystem/openshift-preflight/redhat-openshift-ecosystem-openshift-preflight-main__4.8.yaml
> 
> I can update this unless you want to take a stab at it.

Yes we do. Go for it.

PR #598, opened 2022-05-04
Fix docker-config flag to come from cobra.PersistentFlags, not cobra.Flags.

naemono:
The `docker-config` Flags in the check command is a persistent flag,  not a flag.  Passing anything to this flag, and then subsequently running `viper.GetString("dockerConfig")` returns nothing.  

see
https://github.com/spf13/viper/issues/292#issuecomment-327996518

Test added showing that this flag is coming from persistentflags, not flags.


bcrochet (approved):
Nice catch.

/lgtm

acornett21 (approved):

PR #599, opened 2022-05-04
Remove extra return and add test coverage to pyxis input builder

komish:
This PR adds tests to the pyxis.CertificationInputBuilder which had 0% coverage, and removes a few extra returns that were not needed (because nothing happened after the conditional blocks that called it.

```
Ran 32 of 32 Specs in 0.013 seconds
SUCCESS! -- 32 Passed | 0 Failed | 0 Pending | 0 Skipped
--- PASS: TestPyxis (0.01s)
PASS
coverage: 58.0% of statements
ok      github.com/redhat-openshift-ecosystem/openshift-preflight/certification/pyxis   0.226s  coverage: 58.0% of statements
```

This removes _most_ of the "red" (untested) code from coverage reports for this source file. The remaining red code is such because it does not technically return an error. I tested the code downstream that actually returns the error, so in practice, this covers the use cases to my eye.

### DIFF
--- a/certification/pyxis/builder_test.go
+++ b/certification/pyxis/builder_test.go
@@ -1,11 +1,15 @@
 package pyxis
 
 import (
+	"encoding/json"
 	"os"
+	"path"
 	"path/filepath"
 
 	. "github.com/onsi/ginkgo/v2/dsl/core"
 	. "github.com/onsi/gomega"
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/formatters"
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/version"
 )
 
 var _ = Describe("Pyxis Builder tests", func() {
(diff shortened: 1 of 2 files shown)


certification/pyxis/builder_test.go
```diff
@@ -39,4 +43,218 @@ var _ = Describe("Pyxis Builder tests", func() {
...
+			It("should return an error", func() {
+				_, err := NewCertificationInput(nil)
+				Expect(err).To(HaveOccurred())
+			})
+		})
+
+		Context("with a non-nil CertProject value as input", func() {
+			p := &CertProject{
+				ID:                  "myID",
+				CertificationStatus: "Pending",
```
bcrochet, line 66: Is this a valid status? Do we care?

acornett21 (approved):

certification/pyxis/builder_test.go
```diff
@@ -39,4 +43,218 @@ var _ = Describe("Pyxis Builder tests", func() {
...
+				err := os.WriteFile(f, []byte("\tHello world!\n"), 0o0755)
+				Expect(err).ToNot(HaveOccurred())
+
+				err = builder.storeCertImage(f)
+				Expect(err).To(HaveOccurred())
+			})
+
+			It("should fail to finalize with no test results", func() {
+				_, err := builder.Finalize()
+				Expect(err).To(HaveOccurred())
```
bcrochet, line 140: Are you assuming that these are being run in any particular order? Doesn't ginkgo randomize tests?

bcrochet (commented):
Couple of questions

certification/pyxis/builder_test.go
komish: We don't validate the status value within preflight, so I would say for the moment that we do not care.

certification/pyxis/builder_test.go
komish: Looks like ginkgo randomizes the top-level specs, but everything within a container will be run in order. 

https://onsi.github.io/ginkgo/#spec-randomization

That said, randomize all is a recommended practice. I'm not sure if that breaks our testing today. I'm guessing wrapping this in before/after each would be the way to go if we want to prepare for --randomize-all? Bah.

certification/pyxis/builder_test.go
acornett21: I believe that `Started`, `In Progress` and `Published` are the only valid statuses.

certification/pyxis/builder_test.go
bcrochet: I'm ok within a spec to rely on it, I just wasn't sure what the default behaviour was.

PR #601, opened 2022-05-04
[Backport] Run tests on release branches

bcrochet:
Since we have release branches now, it is a good idea to run tests
on these branches so that when we cherry-pick to them, it will run
these. tests.


komish (approved):
/lgtm

acornett21 (approved):

PR #602, opened 2022-05-04
add preflight-specific labels to the resulting container image

komish:
We pass through UBI's labels, so I figured we should add some of our own.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,6 @@ ARG release_tag=0.0.0
 ARG ARCH=amd64
 ARG OS=linux
 
-# golang:1.16 image created 2021-06-24T00:31:06.02014601Z 
 FROM docker.io/golang:1.17 AS builder
 ARG quay_expiration
 ARG release_tag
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -18,9 +17,23 @@ RUN make build RELEASE_TAG=${release_tag}
...
+ARG release_tag
+ARG preflight_commit
 ARG ARCH
 ARG OS
 
+# Metadata
+LABEL name="Preflight" \
+      vendor="Red Hat, Inc." \
+      maintainer="Red Hat OpenShift Ecosystem" \
+      version="1" \
```
acornett21, line 29: Non-Blocking: What's the difference between `version` and `release_tag` in context of `metadata`? ie when should version here change?

acornett21 (approved):

Dockerfile
bcrochet: If I'm seeing this correctly, I would think when the Dockerfile changes. But @komish would be better to answer this.

Makefile
```diff
@@ -38,7 +38,7 @@ tidy:
 
 .PHONY: image-build
 image-build:
-	$(IMAGE_BUILDER) build -t $(IMAGE_REPO)/preflight:$(VERSION) .
+	$(IMAGE_BUILDER) build --build-arg release_tag=$(RELEASE_TAG) --build-arg preflight_commit=$(VERSION) -t $(IMAGE_REPO)/preflight:$(VERSION) .
```
bcrochet, line 41: Non-blocking: I'd like to see these maybe passed in as a make envvar? Then they can be modified, added to, overridden from elsewhere (or even not present in a local build).

bcrochet (approved):
/lgtm

Just a comment for thought.

Dockerfile
acornett21: That makes sense to me.

PR #604, opened 2022-05-04
updating a few of the help texts

acornett21:
- Updating a few of the help cmd text values since a few partners have had confusion around these values.

Below is the new output
```
Flags:
      --certification-project-id string   Certification Project ID from connect.redhat.com/projects/{certification-project-id}/overview
                                          URL paramater. This value may differ from the PID on the overview page. (env: PFLT_CERTIFICATION_PROJECT_ID)
  -h, --help                              help for container
      --pyxis-api-token string            API token for Pyxis authentication (env: PFLT_PYXIS_API_TOKEN)
      --pyxis-env string                  Env to use for Pyxis submissions. (default "prod")
      --pyxis-host string                 Host to use for Pyxis submissions. This will override Pyxis Env. Only set this if you know what you are doing.
                                          If you do set it, it should include just the host, and the URI path. (env: PFLT_PYXIS_HOST)
  -s, --submit                            submit check container results to red hat

Global Flags:
  -d, --docker-config string   Path to docker config.json file. This value is optional for publicly accessible images.
                               However, it is strongly encouraged for public Docker Hub images,
                               due to the rate limit imposed for unauthenticated requests. (env: PFLT_DOCKERCONFIG)
  -l, --list-checks            lists all the checks run for a given check

```

### DIFF
--- a/cmd/check_container.go
+++ b/cmd/check_container.go
@@ -248,13 +248,15 @@ func init() {
 	checkContainerCmd.Flags().String("pyxis-api-token", "", "API token for Pyxis authentication (env: PFLT_PYXIS_API_TOKEN)")
 	viper.BindPFlag("pyxis_api_token", checkContainerCmd.Flags().Lookup("pyxis-api-token"))
 
-	checkContainerCmd.Flags().String("pyxis-host", "", fmt.Sprintf("Host to use for Pyxis submissions.\nThis will override Pyxis Env.\nOnly set this if you know what you are doing.\nIf you do set it, it should include just the host, and the URI path.\n(env: PFLT_PYXIS_HOST)"))
+	checkContainerCmd.Flags().String("pyxis-host", "", fmt.Sprintf("Host to use for Pyxis submissions. This will override Pyxis Env. Only set this if you know what you are doing.\n"+
+		"If you do set it, it should include just the host, and the URI path. (env: PFLT_PYXIS_HOST)"))
 	viper.BindPFlag("pyxis_host", checkContainerCmd.Flags().Lookup("pyxis-host"))
 
 	checkContainerCmd.Flags().String("pyxis-env", certification.DefaultPyxisEnv, "Env to use for Pyxis submissions.")
 	viper.BindPFlag("pyxis_env", checkContainerCmd.Flags().Lookup("pyxis-env"))
 
-	checkContainerCmd.Flags().String("certification-project-id", "", fmt.Sprintf("Certification Project ID from connect.redhat.com.\nShould be supplied without the ospid- prefix.\n(env: PFLT_CERTIFICATION_PROJECT_ID)"))
+	checkContainerCmd.Flags().String("certification-project-id", "", fmt.Sprintf("Certification Project ID from connect.redhat.com/projects/{certification-project-id}/overview\n"+
+		"URL paramater. This value may differ from the PID on the overview page. (env: PFLT_CERTIFICATION_PROJECT_ID)"))
 	viper.BindPFlag("certification_project_id", checkContainerCmd.Flags().Lookup("certification-project-id"))
 
 	checkCmd.AddCommand(checkContainerCmd)
--- a/cmd/check.go
+++ b/cmd/check.go
@@ -22,7 +22,9 @@ func init() {
 
 	checkCmd.PersistentFlags().BoolP("list-checks", "l", false, "lists all the checks run for a given check")
 
-	checkCmd.PersistentFlags().StringP("docker-config", "d", "", "path to docker config.json file (env: PFLT_DOCKERCONFIG)")
+	checkCmd.PersistentFlags().StringP("docker-config", "d", "", "Path to docker config.json file. This value is optional for publicly accessible images.\n"+
+		"However, it is strongly encouraged for public Docker Hub images,\n"+
+		"due to the rate limit imposed for unauthenticated requests. (env: PFLT_DOCKERCONFIG)")
 	viper.BindPFlag("dockerConfig", checkCmd.PersistentFlags().Lookup("docker-config"))
 
 	rootCmd.AddCommand(checkCmd)


cmd/check_container.go
```diff
@@ -248,13 +248,15 @@ func init() {
 	checkContainerCmd.Flags().String("pyxis-api-token", "", "API token for Pyxis authentication (env: PFLT_PYXIS_API_TOKEN)")
 	viper.BindPFlag("pyxis_api_token", checkContainerCmd.Flags().Lookup("pyxis-api-token"))
 
-	checkContainerCmd.Flags().String("pyxis-host", "", fmt.Sprintf("Host to use for Pyxis submissions.\nThis will override Pyxis Env.\nOnly set this if you know what you are doing.\nIf you do set it, it should include just the host, and the URI path.\n(env: PFLT_PYXIS_HOST)"))
+	checkContainerCmd.Flags().String("pyxis-host", "", fmt.Sprintf("Host to use for Pyxis submissions.\nThis will override Pyxis Env."+
+		"\nOnly set this if you know what you are doing.\nIf you do set it, it should include just the host, and the URI path.\n(env: PFLT_PYXIS_HOST)"))
```
bcrochet, line 252: Can you format this like the other? But with the newlines at the end? Convince me why newlines would go at the beginning?

cmd/check_container.go
```diff
@@ -248,13 +248,15 @@ func init() {
...
 	viper.BindPFlag("pyxis_host", checkContainerCmd.Flags().Lookup("pyxis-host"))
 
 	checkContainerCmd.Flags().String("pyxis-env", certification.DefaultPyxisEnv, "Env to use for Pyxis submissions.")
 	viper.BindPFlag("pyxis_env", checkContainerCmd.Flags().Lookup("pyxis-env"))
 
-	checkContainerCmd.Flags().String("certification-project-id", "", fmt.Sprintf("Certification Project ID from connect.redhat.com.\nShould be supplied without the ospid- prefix.\n(env: PFLT_CERTIFICATION_PROJECT_ID)"))
+	checkContainerCmd.Flags().String("certification-project-id", "", fmt.Sprintf("Certification Project ID from connect.redhat.com/projects/{certification-project-id}/overview"+
+		" \nURL paramater. This value may differ from the PID on the overview page.\n(env: PFLT_CERTIFICATION_PROJECT_ID)"))
```
bcrochet, line 259: Same with format.

bcrochet (changes requested):
Couple of comments about formatting.

cmd/check_container.go
acornett21: Done.

cmd/check_container.go
acornett21: Done.

cmd/check_container.go
```diff
@@ -248,13 +248,15 @@ func init() {
...
+	checkContainerCmd.Flags().String("pyxis-host", "", fmt.Sprintf("Host to use for Pyxis submissions.\nThis will override Pyxis Env.\n"+
+		"Only set this if you know what you are doing.\nIf you do set it, it should include just the host, and the URI path.\n(env: PFLT_PYXIS_HOST)"))
 	viper.BindPFlag("pyxis_host", checkContainerCmd.Flags().Lookup("pyxis-host"))
 
 	checkContainerCmd.Flags().String("pyxis-env", certification.DefaultPyxisEnv, "Env to use for Pyxis submissions.")
 	viper.BindPFlag("pyxis_env", checkContainerCmd.Flags().Lookup("pyxis-env"))
 
-	checkContainerCmd.Flags().String("certification-project-id", "", fmt.Sprintf("Certification Project ID from connect.redhat.com.\nShould be supplied without the ospid- prefix.\n(env: PFLT_CERTIFICATION_PROJECT_ID)"))
+	checkContainerCmd.Flags().String("certification-project-id", "", fmt.Sprintf("Certification Project ID from connect.redhat.com/projects/{certification-project-id}/overview\n"+
```
komish, line 258: Is there really no other place to get this value than from the URL?

If provided the PID from the overview page, could we instead look up the ospid?

komish (commented):
Few comments

cmd/check_container.go
acornett21: @komish For current projects, the URL value _is_ the PID value, but for legacy projects they do not match and partners give us the value in the PID field which is also the ospid value for those legacy projects. As far as I know all pyxis api's take the URL value, so not sure we can cross reference, nor would I really be in favor for another call to pyxis.

cmd/check_container.go
```diff
@@ -248,13 +248,15 @@ func init() {
 	checkContainerCmd.Flags().String("pyxis-api-token", "", "API token for Pyxis authentication (env: PFLT_PYXIS_API_TOKEN)")
 	viper.BindPFlag("pyxis_api_token", checkContainerCmd.Flags().Lookup("pyxis-api-token"))
 
-	checkContainerCmd.Flags().String("pyxis-host", "", fmt.Sprintf("Host to use for Pyxis submissions.\nThis will override Pyxis Env.\nOnly set this if you know what you are doing.\nIf you do set it, it should include just the host, and the URI path.\n(env: PFLT_PYXIS_HOST)"))
+	checkContainerCmd.Flags().String("pyxis-host", "", fmt.Sprintf("Host to use for Pyxis submissions.\nThis will override Pyxis Env.\n"+
```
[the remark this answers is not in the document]
komish: Four line breaks just seem like a lot. Totally a nit. Feel free to disregard.

--

As it pertains to adding an example value: 

If we're really concerned about partners using this when they shouldn't (as in, when they mean to use the other flag), then perhaps we need to massage how we indicate to the user that this is unlikely to be the flag they want. Our current verbiage (`Only set this if you know what you are doing.) doesn't really help in this regard. 

Not including an example value to prevent folks from using it when they shouldn't doesn't make sense.

If we really want to go that route, just mark the flag hidden. I'd prefer not to do this.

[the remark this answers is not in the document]
acornett21: I didn't change any of the verbiage here, or come up with it, it's existing. I just changed the formatting so the code was readable and not 800 characters of scroll city. For this flag I'd be fine with it being hidden, since it would be for us as developers to use to test potential issues between the internal/external instances or say a pyxis route change.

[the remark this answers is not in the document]
komish: I'll let @bcrochet chime in. I'll toss in the LGTM if Brad's good with it.

komish (approved):
/lgtm

jomkz (approved):

PR #605, opened 2022-05-06
Increase Pyxis test coverage

bcrochet:
* Increase the Pyxis test coverage
* Fix some inconsistencies with the Pyxis test harness
* Fix some log messages that were poorly formatted
* Change some magic number to use constants from http package

### DIFF
--- a/certification/pyxis/pyxis.go
+++ b/certification/pyxis/pyxis.go
@@ -60,7 +60,7 @@ func (p *pyxisClient) createImage(ctx context.Context, certImage *CertImage) (*C
 
 	resp, err := p.Client.Do(req)
 	if err != nil {
-		log.Error(err)
+		log.Error(fmt.Errorf("%w: cannot create image", err))
 		return nil, err
 	}
 
@@ -72,7 +72,7 @@ func (p *pyxisClient) createImage(ctx context.Context, certImage *CertImage) (*C
 		return nil, err
 	}
 
-	if resp.StatusCode == 409 {
+	if resp.StatusCode == http.StatusConflict {
 		return nil, errors.ErrPyxis409StatusCode
 	}
 
@@ -121,7 +121,7 @@ func (p *pyxisClient) getImage(ctx context.Context, dockerImageDigest string) (*
 
 	// using an inline struct since this api's response is in a different format
 	data := struct {
-		Data []CertImage `json:"data,omitempty"`
+		Data []CertImage `json:"data"`
 	}{}
 
 	if err := json.Unmarshal(body, &data); err != nil {
--- a/certification/pyxis/pyxis_suite_test.go
+++ b/certification/pyxis/pyxis_suite_test.go
@@ -1,10 +1,10 @@
 package pyxis
 
 import (
+	"fmt"
 	"io"
 	"net/http"
 	"net/http/httptest"
-	"strings"
 	"testing"
 
 	. "github.com/onsi/ginkgo/v2/dsl/core"
@@ -50,6 +50,11 @@ type (
 	pyxisGraphqlLayerHandler struct{}
 )
 
+// For each of these ServeHTTP methods, there is a main switch statement that controls what response will
+// be sent back. Most are switched on the API Key that is provided. Depending on the request.Method, it
+// will determine whether to send back an error, or a valid response. This can probably be improved a bit
+// to dedupe. Acknowledged that it is a bit fragile. -bpc
+
 func (p *pyxisProjectHandler) ServeHTTP(response http.ResponseWriter, request *http.Request) {
 	log.Trace("In the Project ServeHTTP")
 	response.Header().Set("Content-Type", "application/json")
@@ -57,21 +62,18 @@ func (p *pyxisProjectHandler) ServeHTTP(response http.ResponseWriter, request *h
 		defer request.Body.Close()
 	}
 	switch {
+	case request.Method == http.MethodGet && request.Header["X-Api-Key"][0] == "my-401-project-api-token":
+		response.WriteHeader(http.StatusUnauthorized)
 	case request.Header["X-Api-Key"][0] == "my-bad-project-api-token":
-		response.WriteHeader(401)
-	case request.Header["X-Api-Key"][0] == "my-update-project-api-token":
-		if request.Method == http.MethodGet {
-			mustWrite(response, `{"_id":"deadb33f","certification_status":"Started","name":"My Spiffy Project","project_status":"Foo","type":"Containers","container":{"docker_config_json":"{}","type":"Containers"}}`)
-			break
-		}
-
-		response.WriteHeader(500)
+		response.WriteHeader(http.StatusUnauthorized)
 	case request.Header["X-Api-Key"][0] == "my-index-docker-io-project-api-token":
 		mustWrite(response, `{"_id":"deadb33f","certification_status":"Started","name":"My Index Docker IO Project","project_status":"Foo","type":"Containers","container":{"docker_config_json":"{}","type":"Containers","registry":"docker.io", "repository":"my/repo"}}`)
+	case request.Method == http.MethodPatch && request.Header["X-Api-Key"][0] == "my-error-project-api-token":
+		response.WriteHeader(http.StatusInternalServerError)
 	case request.Method == http.MethodPost:
 		body, err := io.ReadAll(request.Body)
 		if err != nil {
-			response.WriteHeader(400)
+			response.WriteHeader(http.StatusBadRequest)
 		}
 		mustWrite(response, string(body))
 	default:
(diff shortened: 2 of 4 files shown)


certification/pyxis/pyxis.go
```diff
@@ -60,7 +60,7 @@ func (p *pyxisClient) createImage(ctx context.Context, certImage *CertImage) (*C
 
 	resp, err := p.Client.Do(req)
 	if err != nil {
-		log.Error(err)
+		log.Error(fmt.Errorf("%w: cannot create image in createImage", err))
```
komish, line 63: nit: `cannot create image in createImage` --> `cannot create image` for the log.Error log level? Seems redundant otherwise.

certification/pyxis/pyxis_suite_test.go
```diff
@@ -77,13 +81,26 @@ func (p *pyxisImageHandler) ServeHTTP(response http.ResponseWriter, request *htt
...
+	case request.Method == http.MethodGet && request.Header["X-Api-Key"][0] == "my-bad-401-image-api-token":
+		log.Trace("get with 401")
+		response.WriteHeader(http.StatusUnauthorized)
 	case request.Header["X-Api-Key"][0] == "my-bad-image-api-token":
-		response.WriteHeader(401)
+		response.WriteHeader(http.StatusUnauthorized)
+	case request.Method == http.MethodPost && request.Header["X-Api-Key"][0] == "my-bad-500-image-api-token":
+		response.WriteHeader(http.StatusInternalServerError)
+	case request.Method == http.MethodPost:
+		mustWrite(response, responseString)
```
komish, line 101: This looks like it intends to test a variety of scenarios. I'd benefit from a few words in a comment going over the matrix.

certification/pyxis/pyxis_suite_test.go
```diff
@@ -95,11 +112,15 @@ func (p *pyxisRPMManifestHandler) ServeHTTP(response http.ResponseWriter, reques
...
+	case request.Method == http.MethodPost && request.Header["X-Api-Key"][0] == "my-bad-rpmmanifest-409-api-token":
+		response.WriteHeader(http.StatusConflict)
 		mustWrite(response, `{"_id":"foo"}`)
+	case request.Method == http.MethodPost && request.Header["X-Api-Key"][0] == "my-bad-rpmmanifest-401-api-token":
+		response.WriteHeader(http.StatusConflict)
+	case request.Method == http.MethodGet && request.Header["X-Api-Key"][0] == "my-bad-rpmmanifest-401-api-token":
+		response.WriteHeader(http.StatusUnauthorized)
 	case request.Header["X-Api-Key"][0] == "my-bad-rpmmanifest-api-token":
-		response.WriteHeader(401)
+		response.WriteHeader(http.StatusUnauthorized)
```
komish, line 123: Same thing here, re: comments

certification/pyxis/submit.go
```diff
@@ -51,31 +51,30 @@ func (p *pyxisClient) SubmitResults(ctx context.Context, certInput *certificatio
...
 		if err != nil {
-			log.Error(err, "could not get image")
+			log.Error(fmt.Errorf("%w: could not get image", err))
 			return nil, err
 		}
 	}
+	if err != nil && err != errors.ErrPyxis409StatusCode {
+		log.Error(fmt.Errorf("%w: could not create image", err))
+		return nil, err
+	}
```
komish, line 64: I know we (and Go) typically recommends keeping the flow as close to the left-most indentation as possible (as in, nest less), but this seems like a prime opportunity where nesting would fit. 

```Go
if err != nil {
    if err == errors.ErrPyxis409StatusCode { // or a case statement if we want to match more than one type of error
        ...
    }

    log...
    return...
}

File: certification/pyxis/submit.go [hunk 5]
```diff
@@ -51,31 +51,30 @@ func (p *pyxisClient) SubmitResults(ctx context.Context, certInput *certificatio
...
 		return nil, err
 	}
 	if err != nil && err == errors.ErrPyxis409StatusCode {
 		_, err = p.getRPMManifest(ctx, rpmManifest.ImageID)
 		if err != nil {
-			log.Error(err, "could not get rpm manifest")
+			log.Error(fmt.Errorf("%w: could not get rpm manifest", err))
 			return nil, err
 		}
 	}
```
komish, line 80: This might also benefit from the previous comment on similar code, though I know this didn't really change here.

certification/pyxis/submit.go
```diff
@@ -85,7 +84,7 @@ func (p *pyxisClient) SubmitResults(ctx context.Context, certInput *certificatio
 	for _, artifact := range artifacts {
 		artifact.ImageID = certImage.ID
 		if _, err := p.createArtifact(ctx, &artifact); err != nil {
-			log.Errorf("%s: could not create artifact: %s", err, artifact.Filename)
+			log.Error(fmt.Errorf("%w: could not create artifact: %s", err, artifact.Filename))
```
komish, line 87: great catch

certification/pyxis/submit_test.go
```diff
@@ -14,264 +14,210 @@ var _ = Describe("Pyxis Submit", func() {
...
+	var certInput certificationInput
 	mux := http.NewServeMux()
 	mux.Handle("/api/v1/projects/certification/id/", &pyxisProjectHandler{})
 	mux.Handle("/api/v1/images", &pyxisImageHandler{})
+	mux.Handle("/api/v1/projects/certification/id/my-image-project-id/images", &pyxisImageHandler{})
 	mux.Handle("/api/v1/images/id/blah/", &pyxisRPMManifestHandler{})
 	mux.Handle("/api/v1/projects/certification/id/my-awesome-project-id/test-results", &pyxisTestResultsHandler{})
 
+	BeforeEach(func() {
+		pyxisClient = NewPyxisClient("my.pyxis.host/api", "my-spiffy-api-token", "my-awesome-project-id", &http.Client{Transport: localRoundTripper{handler: mux}})
```
komish, line 26: nit: multiline this?

certification/pyxis/pyxis.go
bcrochet: Fixed.

certification/pyxis/submit.go
bcrochet: This change actually wasn't supposed to be here. :) Only the log fix. I had re-ordered it. I'll play with it.

certification/pyxis/pyxis_suite_test.go
bcrochet: How do you feel about nested switch? I think it might help the clarity here. I'm thinking switching on request.Method as the top-level.

certification/pyxis/submit_test.go
```diff
@@ -14,54 +14,83 @@ var _ = Describe("Pyxis Submit", func() {
...
+			})
+		})
+		Context("and the client sends a bad token", func() {
+			JustBeforeEach(func() {
+				pyxisClient.ApiToken = "my-bad-project-api-token"
+			})
+			It("should get an unauthorized", func() {
+				certResults, err := pyxisClient.SubmitResults(ctx, &certInput)
+				Expect(err).To(HaveOccurred())
+				Expect(err).To(MatchError(fmt.Errorf("%w: %s", errors.New("error calling remote API"), "could not retrieve project")))
```
komish, line 84: I mentioned this in Adam's PR as well, but the MatchError feels too granular in that we have the entire error definition written here, and I'm assuming, in the calling code as well.

Assuming these errors are our errors (and not errors coming from the remote API), should we be extracting these into local/package error variables? I don't think they need to go to the top level errors package, but wouldn't it be better to have the string definition of this error in one place? 

I'm not personally offended by checking that an error occurred without asserting the value of it... but I can see it either way. We're going through the effort to wrap, so if you want to test that we get the proper value, that's fine. 

Just my 2c. Feel free to disregard.

certification/pyxis/pyxis_suite_test.go
komish: I'm on the fence. But I'm also somewhat on the fence with the general need for a switch here.

I kind of get the need to do multiple scenarios. I think what I'm wondering is, would there be a way to restructure the test that would require less switching?

komish (changes requested):
Couple of comments in line. Only submitting a review again because GH says my comments are pending.

certification/pyxis/pyxis_suite_test.go
bcrochet: I take it back. I think I want to stick with a flat switch. Then I don't end up making duplicates for ones that are the same for Get and Post.

certification/pyxis/pyxis_suite_test.go
bcrochet: I'm making a "top-level" comment about these.

certification/pyxis/pyxis_suite_test.go
acornett21: FWIW, when I was adding tests to this, I thought about a nested switch as well when I thought I might need to marshal the request payload for an edge case, but when I coded that I hated how it read. I ended up just going with an `if` for the one case, and default return for everything else. Hopefully we never have a case were need more then two options within a given switch (ie GET vs POST).

certification/pyxis/submit_test.go
bcrochet: I'm not a fan of matching the error either. I'm just going to remove.

bcrochet:
@komish I'm not sure why it won't let me respond directly... Anyway, about the switch in the handlers. Basically, I'm trying to minimize the number of handlers. Otherwise, I think we'd end up with a bunch of small handlers. I do admit that it's rather fragile. If you have any suggestions for improving this mocked up http server, I'm all ears.

certification/pyxis/submit_test.go
acornett21: Why don't you all like matching the error? To me it makes sense to match the error so we know that the code is covered and hitting the proper flow we expect. We match values on `happy` path testing, ie `registry = docker.io` as an example to make sure we did something properly. I really don't understand not wanting to do the same for exceptions.

komish: @acornett21 for me, it's because the message we're testing for is a message that we set. I think it would be one thing if we were testing the external systems error messages (as in, make sure _this code path_ results in _that error_), but I don't think that's what's happening here.

komish:
> @komish I'm not sure why it won't let me respond directly... Anyway, about the switch in the handlers. Basically, I'm trying to minimize the number of handlers. Otherwise, I think we'd end up with a bunch of small handlers. I do admit that it's rather fragile. If you have any suggestions for improving this mocked up http server, I'm all ears.

I think I need to spend some time digging further into HTTP client/server testing in Go to really be able to offer an alternative. Which is to say that I don't have one, and I may as well just find myself reaching the same conclusion/implementation you have here after I spend time with it.

to that end, I think I'm comfortable with the way it's being tested even though it does somewhat feel/read a bit wild. The amount of running code (vs. test code) being touched is also minimal. Overall: 👍

komish (approved):
/lgtm

acornett21 (approved):

PR #607, opened 2022-05-06
[BACKPORT] move update logic into `started` since we should only care about upda…

komish:
…ting net new projects all other updates are taken care of by backend systems (#603)


acornett21 (approved):

komish:
/close

no longer backporting

PR #608, opened 2022-05-06
[BACKPORT] updating a few of the help texts based on a few issues that partners have had with the check container cmd (#604)

komish:
#604


komish:
/close

no longer backporting

PR #609, opened 2022-05-09
Deprecate release-1.1 branch

bcrochet:
Add a note to README.md that this branch should not be used going
forward.


bcrochet:
> 1. Do we need to delete everything? Would a note in the README.md, and an unplug of .github/ not be sufficient?
> 

In my opinion, yes. My thought is it is plainly obvious if literally nothing is there but a README.

>     2. We would need @tkrishtop to fix DCI? (sorry @tkrishtop!)

Probably nothing to fix. We won't be using a release branch anymore, so as long as it still works on main...

komish:
I supposed that's fine, then. Did you also remove the release-1.1 branch's .github? I assume we also need a subsequent PR to remove references to release-1.1 from our main branch's actions as well.

For @tkrishtop I just mean removing any hooks that look for the branch. I I know we won't be sending anything to them from that branch because we won't be doing anything with it, but from a pipeline cleanliness perspective, they're free to remove any references.

tkrishtop:
Hi @komish the DCI error is because there is no Dockerfile on the release-1.1 branch anymore and the build of the preflight image fails. I agree with @bcrochet that there is nothing to fix if you're not going to use a release branch anymore.

jomkz (approved):

komish (approved):
/lgtm

PR #612, opened 2022-05-09
Refactor Pyxis builder to use an io.Reader

### LINKED ISSUE #611: Refactor Pyxis builder to use io.Reader
To enhance the testability and improve code coverage of the pyxis.Builder, the methods should accept an io.Reader instead of a filename.

bcrochet:
Instead of passing filenames through the various layers, it is better
to pass an io.Reader instead. This increases the testability of the
builder itself.

Fixes #611


certification/pyxis/builder.go
```diff
@@ -67,68 +66,62 @@ func (b *certificationInputBuilder) Finalize() (*certificationInput, error) {
...
+func (b *certificationInputBuilder) WithRPMManifest(r io.Reader) *certificationInputBuilder {
+	if err := b.storeRPMManifest(r); err != nil {
 		log.Error(err)
 	}
 
 	return b
 }
 
-// WithArtifactFromFile reads a file at path and binds it as an artifact to include
+// WithArtifactFromFile reads from the io.Reader and binds it as an artifact to include
```
komish, line 99: The method name in this comment needs to change to match what it is on line 102.

certification/pyxis/builder.go
```diff
@@ -67,68 +66,62 @@ func (b *certificationInputBuilder) Finalize() (*certificationInput, error) {
...
 // in the submission. Multiple calls to this will append artifacts. Errors are logged,
-// but will not halt execution.
-func (b *certificationInputBuilder) WithArtifactFromFile(filepath string) *certificationInputBuilder {
-	bytes, err := os.ReadFile(filepath)
+// but will not halt execution. filename is the base filename.
```
komish, line 101: grammar nit: "filename is the..." --> "Filename is the..."

certification/pyxis/builder.go
```diff
@@ -67,68 +66,62 @@ func (b *certificationInputBuilder) Finalize() (*certificationInput, error) {
...
 
 	newArtifact := Artifact{
 		CertProject: b.CertProject.ID,
-		Content:     base64.StdEncoding.EncodeToString(bytes),
-		ContentType: http.DetectContentType(bytes),
-		Filename:    path.Base(filepath),
-		FileSize:    int64(len(bytes)),
+		Content:     base64.StdEncoding.EncodeToString(bts),
+		ContentType: http.DetectContentType(bts),
+		FileSize:    int64(len(bts)),
```
komish, line 113: What do you mean when you say the filename is the base filename? I see we're not passing a filename value here to the Artifact struct.

certification/pyxis/pyxis_suite_test.go
```diff
@@ -139,3 +140,9 @@ func (p *pyxisGraphqlLayerHandler) ServeHTTP(response http.ResponseWriter, reque
 	mustWrite(response, `{"data":{"find_images":{"error":null,"total":1,"page":0,"data":[{"uncompressed_top_layer_id":"good_top_layer","_id":"deadb33f"}]}}}`)
 	return
 }
+
+type errReader int
```
komish, line 144: Small comment here would help. Just something that indicates we're using this to facilitate throwing errors when passing this into io.Reader.

cmd/check_container.go
```diff
@@ -186,15 +186,51 @@ var checkContainerCmd = &cobra.Command{
...
+				)
+			}
+			logfile, err := os.Open(viper.GetString("logfile"))
+			if err != nil {
+				return fmt.Errorf(
+					"%w: could not open file for submission: %s",
+					err,
+					viper.GetString("logfile"),
+				)
+			}
```
komish, line 220: In fairness, passing the file-reading logic to the builder meant this check_container was easier to read because it didn't have these 4 opens, and associated error handling. Ah well. the testing is better. I'll take that.

That said, are we missing the `defer (*File).close()` calls here?

komish (changes requested):
Nice work. Couple of small things!

cmd/check_container.go
```diff
@@ -186,15 +186,51 @@ var checkContainerCmd = &cobra.Command{
...
+				WithPreflightResults(preflightResults).
 				// The certification engine writes the rpmManifest for images not based on scratch.
-				WithRPMManifestFromFile(path.Join(artifacts.Path(), certification.DefaultRPMManifestFilename)).
+				WithRPMManifest(rpmManifest).
 				// Include the preflight execution log file.
-				WithArtifactFromFile(viper.GetString("logfile"))
+				WithArtifact(logfile)
+
+			if len(submission.Artifacts) > 0 {
+				submission.Artifacts[len(submission.Artifacts)-1].Filename = viper.GetString("logfile")
```
acornett21, line 232: This is going to be really confusing to remember in the future. Part of the logic is building all the attributes of the `Artifacts` struct (size, projectId, contentType, etc), then here we are going to add the filename to that struct? No one will remember that's how this works after this PR is merged. Can't we just have the filename be set in the builder?

acornett21 (changes requested):

certification/pyxis/builder.go
bcrochet: That's not even relevant anymore.

certification/pyxis/builder.go
bcrochet: That was there before. I've removed that.

cmd/check_container.go
bcrochet: As I read this, I realized that was missing.

cmd/check_container.go
bcrochet: That's how I originally had it. I'm open to adding it back.

certification/pyxis/builder.go
komish: @bcrochet Do we need to pass in a filename? Is that not relevant to the receiving side (Pyxis)?

komish: Disregard. I see you did this already.

komish (approved):
/lgtm

jomkz (approved):

PR #624, opened 2022-05-10
Increase test coverage of the version package

### LINKED ISSUE #621: Test coverage improvement: github.com/redhat-openshift-ecosystem/openshift-preflight/version
This is a trivial package, but adding test coverage to it should be simple enough.

komish:
Version is tiny, so fairly easy to get 100% here. I added a few extra tests relevant to how we use version.

- Checks to make sure we have JSON struct tags, since we marshal the VersionContext into Preflight Results IIRC
- ensures that ldflags can properly replace these values. Updated the Makefile to make sure we pass these flags during tests (with prepared values)
- Makes sure that the implemented Stringer interface includes the commit and version.

Fixes #621 

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,11 +46,16 @@ image-push:
 
 .PHONY: test
 test:
-	go test -v $$(go list ./... | grep -v e2e)
+	go test -v $$(go list ./... | grep -v e2e) \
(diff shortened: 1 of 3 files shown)


bcrochet (approved):
/lgtm

acornett21 (approved):

PR #630, opened 2022-05-12
Refactor hasUniqueTagCheck to get rid of CraneEngine

bcrochet:
The extra "CraneEngine" that should have never been name that was
problematic. It was introduced merely to test crane calls. A better
way to test crane is available, so let's get rid of some unnecessary
abstractions.


acornett21 (approved):

komish (approved):
/lgtm

Issue #617, opened 2022-05-09
Test coverage improvement: github.com/redhat-openshift-ecosystem/openshift-preflight/certification/internal/engine


komish:
This has minimal coverage (0.3%) and has various functions, structs, and methods that need testing. Some of this code may need to be refactored to enable test cases.

PR #631, opened 2022-05-12
Enhance test coverage for internal/engine package

bcrochet:
* Remove the Podman engine, and the Checks that use it. It is unused
* Introduce test registry to enable testing with crane directly
* Will address each engine in separate PRs to keep the size down

Part of #617


certification/internal/engine/engine.go
```diff
@@ -105,12 +108,14 @@ func (c *CraneEngine) ExecuteChecks(ctx context.Context) error {
...
 			log.Error("unable to export and flatten container filesystem:", err)
 		}
+		wg.Done()
 	}()
 
 	log.Debug("extracting container filesystem to ", containerFSPath)
 	if err := untar(containerFSPath, r); err != nil {
 		return fmt.Errorf("%w: %s", errors.ErrExtractingTarball, err)
 	}
+	wg.Wait()
```
komish, line 118: What are doing with this waitgroup? Prepping for more concurrent executions? Fixing a bug?

certification/internal/engine/engine_test.go
```diff
@@ -1,10 +1,111 @@
...
+			Expect(err).ToNot(HaveOccurred())
+			Expect(engine.results.Passed).To(HaveLen(1))
+			Expect(engine.results.Failed).To(HaveLen(1))
+			Expect(engine.results.Errors).To(HaveLen(1))
+			Expect(engine.results.CertificationHash).To(BeEmpty())
+		})
+		Context("it is a bundle", func() {
+			JustBeforeEach(func() {
+				engine.IsBundle = true
+			})
```
komish, line 90: No change required with this comment. Just something I'm noting to commit to my own memory (and sharing if it helps).

This use of `JustBeforeEach` would be unnecessary because we only have a single `It` block below it. OTOH, if we had multiples, and they ran the chance of changing the value of `engine.IsBundle`, this would reset state for us.

I think this probably _could_ be restructured so that there are separate specs that both live within the top-level context having access to the configured `engine`, but this seems more than sufficient for this.

certification/internal/engine/engine_test.go
```diff
@@ -1,10 +1,111 @@
...
+	BeforeEach(func() {
+		// Set up a fake registry.
+		s := httptest.NewServer(registry.New())
+		DeferCleanup(func() {
+			s.Close()
+		})
+		u, err := url.Parse(s.URL)
+		Expect(err).ToNot(HaveOccurred())
+
+		src = fmt.Sprintf("%s/test/crane", u.Host)
```
komish, line 30: Is this a built-in "image" that the register package makes available for use in tests? I would benefit from a comment here.

certification/internal/engine/engine_test.go
```diff
@@ -1,10 +1,111 @@
...
+			})
+			It("should succeed and generate a bundle hash", func() {
+				err := engine.ExecuteChecks(context.TODO())
+				Expect(err).ToNot(HaveOccurred())
+				Expect(engine.results.CertificationHash).ToNot(BeEmpty())
+			})
+		})
+		Context("the image is invalid", func() {
+			JustBeforeEach(func() {
+				engine.Image = "does.not/exist/anywhere:ever"
```
komish, line 99: In relation to the use of `%s/test/crane` earlier, I'm assuming this replaces the image value with a path which the register server cannot serve, hence, why this is a pull failure case?

komish: Resolve this comment if I understood it correctly

certification/internal/engine/engine_test.go
acornett21: Based on his other PR's, I understand that you can place any image into the registry then look for it at a later time. So I see this as create fake registry, give the image a src, create the random image, then push it to the fake registry, so the fake registry can have N number of images. That's my understanding at least. I might be wrong.

komish:
@bcrochet I added the link to #617 so it'll close when this merges. Disconnect them if you have further plans for this package before you close it.

bcrochet:
> @bcrochet I added the link to #617 so it'll close when this merges. Disconnect them if you have further plans for this package before you close it.

This only addresses a single file in this package. The other files may end up moving, but until then, #617 should remain open. And, this only gets engine.go to around 70%. I'd like to see it above 80% before closing.

certification/internal/engine/engine.go
bcrochet: That is correct. Once I enabled tests on this, I immediately got a race condition warning from -race. This fixes it.

certification/internal/engine/engine_test.go
bcrochet: You are correct. This is probably a misuse of JustBeforeEach. I can just set it right before calling the method under test. I am actually going to update this.

certification/internal/engine/engine_test.go
bcrochet: That is correct. It tests the failure path of "not in registry". Probably not a likely scenario, but at least provides coverage of that if err != nil case.

certification/internal/engine/engine_test.go
```diff
@@ -1,10 +1,111 @@
...
+			})
+			It("should throw a crane error on pull", func() {
+				err := engine.ExecuteChecks(context.TODO())
+				Expect(err).To(HaveOccurred())
+			})
+		})
+	})
+})
+
 var _ = Describe("Source RPM name function", func() {
```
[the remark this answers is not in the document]
bcrochet: Agreed. I'm still mulling over where to put that code.

komish (commented):
/lgtm

komish (approved):
Hit enter without hitting Approve.

PR #636, opened 2022-05-16
Refactor Openshift engine to increase testability

bcrochet:
* Also removes an unnecessary abstraction for simplification
* Use the controller-runtime client in place of client-go


certification/engine/engine.go
```diff
@@ -174,3 +181,25 @@ func ScratchContainerPolicy() []string {
...
+	client, err := crclient.New(kubeconfig, crclient.Options{
+		Scheme: scheme,
+	})
+	if err != nil {
+		log.Error("could not get operator group client")
+		return nil
+	}
+
+	return client
+}
```
komish, line 205: This introduces as bit of a UX problem. We initialize our checks right out of the gate of Preflight execution. That means that this code is getting called regardless of what subcommand a user is running. The result is this:

```
$ ./preflight --version
ERRO[0000] could not get operator group client          
preflight version 0.0.0 <commit: 6e773adc8716ab20685474f2576f61bfd9732271>
```

That's because I have KUBECONFIG unset. If I have it set, I still see:

```
$ KUBECONFIG="$PWD/KUBECONFIG" ./preflight --version
I0516 13:40:33.943636    4469 request.go:665] Waited for 1.016926668s due to client-side throttling, not priority and fairness, request: GET:https://mycluster:6443/apis/quota.openshift.io/v1?timeout=32s
preflight version 0.0.0 <commit: 6e773adc8716ab20685474f2576f61bfd9732271>
```

I'm not sure if we want to look at lazy loading the client here to resolve that? Or maybe make this work the first thing that happens when the check runs? 

-

Secondly, a tiny nit, and definitely less importantly, the use of `get` here, and that being an "antipattern"[1].

Given that we're not really calling this on a struct, I don't care if you leave it this way or otherwise.

[1]
https://go.dev/doc/effective_go#Getters

komish (changes requested):

certification/engine/engine.go
bcrochet: This is precisely the piece of code that I wanted eyes on the most. Thanks for the feedback. Clearly this needs more thought because of the loading concern.

komish: The solution here might be something like... accept a `*crclient.Client*`, pass it `nil`, and then run an initializer method on `Validate()` that sets up the client at that time. It's dirty, but it should push the initialization later so it doesn't impact cases that it should not be impacting.

bcrochet: Thanks. I will play with that.

certification/internal/policy/operator/deployable_by_olm.go
```diff
@@ -432,112 +471,76 @@ func (p *DeployableByOlmCheck) installedCSV(ctx context.Context, operatorData Op
...
-	obj, err := runtime.DefaultUnstructuredConverter.ToUnstructured(data)
-	if err != nil {
-		log.Error("unable to convert the object to unstructured.Unstructured: ", err)
-		return err
-	}
-
-	u := &unstructured.Unstructured{Object: obj}
-	switch data.(type) {
-	case *operatorv1alpha1.CatalogSource:
-		u.SetGroupVersionKind(schema.GroupVersionKind{
```
acornett21, line 507: I feel like this was added in for some reason, not sure what it is. We should make sure that the files get written properly and are in the expected format as they were before. IE this change doesn't break connect or the pipeline.

bcrochet: To me, it looks like the unstructured stuff was literally there to be able to unmarshal and then use u.Name() and u.Kind(). If that's all it was for, it's a lot of garbage for nothing.

acornett21: It looks like it's updating the `obj` as well, which made me wonder if the `obj` that is passed in does not have any of the GVK information attached to it.

bcrochet: Looks like it's not writing them at all ATM. I'll figure out what's going on and update.

certification/internal/policy/operator/deployable_by_olm.go
```diff
@@ -41,21 +44,57 @@ type OperatorData struct {
 }
 
 type DeployableByOlmCheck struct {
-	OpenshiftEngine   cli.OpenshiftEngine
 	OperatorSdkEngine cli.OperatorSdkEngine
+	openshiftEngine   cli.OpenshiftEngine
```
komish, line 48: We've got a symbol overloading problem here. It's not directly to your change so consider this non-blocking. Just noting for awareness. `cli.openshiftEngine` (interface) vs. engine.openshiftEngine (impl). My gut instinct was to ask why we have a client key in the next line if openshiftEngine has a client, but then I saw the distinction.

komish: Actually, I think that question is still valid, specifically because NewDepoyableByOLM(...) doesn't seem to accept an openshiftEngine?

certification/internal/policy/operator/deployable_by_olm.go
komish: I'm getting files written in my testing. They're JSON. Don't remember if they used to be or not. LMK what you find.

certification/internal/policy/operator/deployable_by_olm_test.go
```diff
@@ -97,16 +102,25 @@ var _ = Describe("DeployableByOLMCheck", func() {
...
+		Expect(operatorv1.AddToScheme(scheme)).To(Succeed())
+		Expect(operatorv1alpha1.AddToScheme(scheme)).To(Succeed())
+		Expect(imagestreamv1.AddToScheme(scheme)).To(Succeed())
+		Expect(rbacv1.AddToScheme(scheme)).To(Succeed())
+		client = fake.NewClientBuilder().
+			WithScheme(scheme).
+			WithObjects(&csv, &csvDefault, &csvMarketplace, &ns, &secret, &sub).
+			WithLists(&pods, &isList).
+			Build()
+		deployableByOLMCheck.client = client
```
komish, line 116: Okay good, so this was the goal of being able to adjust the client in the check.

komish (changes requested):
Couple more things here for discussion. It's a large PR, so my apologies for the back and forth.

As an aside:

I feel like there's an opportunity to establish all of the inputs we need from the submitted operator bundle into a data structure that then renders out the assets that we'd need to create in OpenShift. Similar to how we might render our secondary resources based on a Custom Resource in operator development.

Might increase legibility of the code. Or maybe make it more clear exactly how preflight is deriving the resources. Nothing required on this topic in this PR. Just food for thought (for myself and anyone else).

certification/internal/policy/operator/deployable_by_olm.go
komish: I'm not getting files. I think I did in a previous iteration of this PR.

Rather, I seem to get _some_ files, but the data is incomplete (missing GVK), and the filenames are incomplete (missing kind).

certification/internal/engine/openshift_test.go
```diff
@@ -0,0 +1,294 @@
...
+		Expect(operatorv1.AddToScheme(scheme)).To(Succeed())
+		Expect(operatorv1alpha1.AddToScheme(scheme)).To(Succeed())
+		Expect(imagestreamv1.AddToScheme(scheme)).To(Succeed())
+		Expect(rbacv1.AddToScheme(scheme)).To(Succeed())
+		cl := fake.NewClientBuilder().
+			WithScheme(scheme).
+			WithObjects(&csv).
+			WithLists(&pods, &isList).
+			Build()
+		oe = &openshiftEngine{Client: cl}
```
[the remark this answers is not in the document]
bcrochet: Correct. Not sure it's worth it just yet?

certification/internal/policy/operator/deployable_by_olm.go
bcrochet: Yeah. I'll put it back the way it way.

certification/internal/policy/operator/deployable_by_olm.go
bcrochet: It does not. In fact, OpenshiftEngine is another "it should have never been called that". It's really just a wrapper around k8s client calls. I want to move it (along with openshift_version.go) into its own package. Open to suggestions.

certification/internal/policy/operator/deployable_by_olm.go
```diff
@@ -41,21 +44,57 @@ type OperatorData struct {
...
-func NewDeployableByOlmCheck(openshiftEngine *cli.OpenshiftEngine, operatorSdkEngine *cli.OperatorSdkEngine) *DeployableByOlmCheck {
+func (p *DeployableByOlmCheck) initClient() error {
+	if p.client != nil {
+		return nil
+	}
+	scheme := apiruntime.NewScheme()
+	operatorv1.AddToScheme(scheme)
+	operatorv1alpha1.AddToScheme(scheme)
+	imagestreamv1.AddToScheme(scheme)
+	rbacv1.AddToScheme(scheme)
```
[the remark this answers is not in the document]
bcrochet: I see. Yeah, that should be doable.

certification/internal/policy/operator/deployable_by_olm_test.go
bcrochet: Yep. Bye bye FakeOpenshiftEngine and BadOpenshiftEngine.

certification/internal/engine/openshift_test.go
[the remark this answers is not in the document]
bcrochet: This will be in a follow-up PR.

acornett21 (approved):

komish (approved):
/lgtm

PR #637, opened 2022-05-17
ioutil calls deprecated

sebrandon1:
Similar to #209 

Was just looking through some of my previous commits and decided to look again for ioutil calls.

https://golang.org/doc/go1.16#ioutil


acornett21 (approved):

bcrochet (approved):
Thanks. I had copied these from go-containerregistry and didn't pay much attention to this.

/lgtm

PR #645, opened 2022-05-18
Fix #366, Expose All Optional Configurations in --help

### LINKED ISSUE #366: Expose all Optional Configurations in `--help` cmd
When trying to decide what optional parameters can be passed into a given check, it would be nice to run the `--help` instead of having to look at the documentation in the repo.

Have all of the optional configurations be listed in the `--help` menu so that it's easier to know what can be configured for a given check. It might also make sense to expose the required ones as well.

N/A

## Additional context.
N/A

rocrisp:


cmd/check_operator.go
```diff
@@ -122,5 +122,24 @@ var checkOperatorCmd = &cobra.Command{
...
+	viper.BindPFlag("serviceaccount", checkOperatorCmd.Flags().Lookup("serviceaccount"))
+
+	checkOperatorCmd.Flags().String("scorecard-image", "", "A uri that points to the scorecard image digest, used in disconnected environments.\n"+
+		"It should only be used in a disconnected environment. Use preflight runtime-assets on a connected \n"+
+		"workstation to generate the digest that needs to be mirrored. (env: PFLT_SCORECARD_IMAGE)")
+	viper.BindPFlag("scorecard_image", checkOperatorCmd.Flags().Lookup("scorecard-image"))
+
+	checkOperatorCmd.Flags().String("scorecard_wait_time", "", "A time value that will be passed to scorecard's --wait-time environment variable.\n"+
+		"(env: PFLT_SCORECARD_WAIT_TIME)")
+	viper.BindPFlag("scorecard_wait_time", checkOperatorCmd.Flags().Lookup("scorecard_wait_time"))
```
acornett21, line 138:
```suggestion
	checkOperatorCmd.Flags().String("scorecard-wait-time", "", "A time value that will be passed to scorecard's --wait-time environment variable.\n"+
		"(env: PFLT_SCORECARD_WAIT_TIME)")
	viper.BindPFlag("scorecard_wait_time", checkOperatorCmd.Flags().Lookup("scorecard-wait-time"))
```

acornett21 (changes requested):

cmd/check_operator.go
rocrisp: ah! Good catch.
Thanks!

cmd/check_operator.go
```diff
@@ -122,5 +122,24 @@ var checkOperatorCmd = &cobra.Command{
...
+	viper.BindPFlag("serviceaccount", checkOperatorCmd.Flags().Lookup("serviceaccount"))
+
+	checkOperatorCmd.Flags().String("scorecard-image", "", "A uri that points to the scorecard image digest, used in disconnected environments.\n"+
+		"It should only be used in a disconnected environment. Use preflight runtime-assets on a connected \n"+
+		"workstation to generate the digest that needs to be mirrored. (env: PFLT_SCORECARD_IMAGE)")
+	viper.BindPFlag("scorecard_image", checkOperatorCmd.Flags().Lookup("scorecard-image"))
+
+	checkOperatorCmd.Flags().String("scorecard-wait-time", "", "A time value that will be passed to scorecard's --wait-time environment variable.\n"+
+		"(env: PFLT_SCORECARD_WAIT_TIME)")
+	viper.BindPFlag("scorecard-wait-time", checkOperatorCmd.Flags().Lookup("scorecard-wait-time"))
```
bcrochet, line 138: The BindPFlag needs to be scorecard_wait_time, otherwise the envvar will be PFLT_SCORECARD-WAIT-TIME.

bcrochet (changes requested):
Quick change on a var name.

acornett21:
@bcrochet For the rest of our configs, log_level, artifacts, etc; do you think it makes sense for those to go in `root.cmd`? I wasn't sure where those should reside.

cmd/check_operator.go
rocrisp: Gotcha.

rocrisp:
@acornett21 @bcrochet 
What do you think about this,
put PFLT_LOGLEVEL and PFLT_LOGFILE in root.go
and put PFLT_ARTIFACTS IN check.go since it's initiated in check.go

bcrochet:
> @acornett21 @bcrochet What do you think about this, put PFLT_LOGLEVEL and PFLT_LOGFILE in root.go and put PFLT_ARTIFACTS IN check.go since it's initiated in check.go

I agree with that.

cmd/check.go
```diff
@@ -27,6 +27,9 @@ func init() {
 		"due to the rate limit imposed for unauthenticated requests. (env: PFLT_DOCKERCONFIG)")
 	viper.BindPFlag("dockerConfig", checkCmd.PersistentFlags().Lookup("docker-config"))
 
+	checkCmd.Flags().String("artifacts", "", "Where check-specific artifacts will be written. (env: PFLT_ARTIFACTS)")
+	viper.BindPFlag("artifacts", checkCmd.Flags().Lookup("artifacts"))
```
bcrochet, line 31: This needs to be PersistentFlags(), like docker-config above it. Otherwise, it doesn't show in help for check container or check operator, only check --help.

cmd/root.go
```diff
@@ -17,6 +18,16 @@ var rootCmd = &cobra.Command{
 	Args:    cobra.MinimumNArgs(1),
 }
 
+func init() {
+	cobra.OnInitialize(initConfig)
+
+	rootCmd.Flags().String("logfile", "", "Where the execution logfile will be written. (env: PFLT_LOGFILE)")
```
bcrochet, line 24: These too need to be PersistentFlags.

bcrochet (changes requested):
PersistentFlags()

cmd/root.go
rocrisp: I added loglevel as well.

cmd/root.go
```diff
@@ -17,6 +18,16 @@ var rootCmd = &cobra.Command{
 	Args:    cobra.MinimumNArgs(1),
 }
 
+func init() {
+	cobra.OnInitialize(initConfig)
```
bcrochet, line 22: Do we know if calling this multiple times is bad? Or necessary? This is being called in check.go as well. It seems that we would only want this once (especially since this is in an init() func).

bcrochet (commented):
Question about init

cmd/root.go
rocrisp: It didn't hurt but it's unnecessary so I removed the call from check.go leaving the root.go intact.

acornett21: I think there are two patters to call `OnInitialize` 
- in the `init()` in the `root.cmd`
- in the `main`
and in both instances it's only called once. So we probably only want to call this once where it already is.

rocrisp: @acornett21 you mean to keep the one in root.go or in check.go

rocrisp: Cool. That's where I left it.

bcrochet (approved):
/lgtm

acornett21 (approved):

PR #652, opened 2022-05-19
Remove errors package

bcrochet:
* Remove the errors package
* Move the errors to be co-located with the package that is using
  the error
* Start the process of only logging errors at upper levels
* Start doc on errors and logging
* certification/pyxis package has been changed to follow practices
  outlined in the ERRORS_AND_LOGGING doc.
* In order to keep this PR from being gigantic, further PRs will be
  opened on a package by package basis, until all packages in the
  codebase have been visited.


certification/internal/engine/errors.go
```diff
@@ -0,0 +1,20 @@
+package engine
```
acornett21, line 1: Are the comments in this package still valid? ie didn't the term `Engine` go away for everything but our own engine?

certification/internal/policy/container/has_license.go
```diff
@@ -2,13 +2,12 @@ package container
 
 import (
 	"context"
-	stdliberrors "errors"
+	"errors"
```
acornett21, line 5: This makes the code below way more readable 👍🏻

certification/pyxis/pyxis.go
```diff
@@ -47,44 +46,41 @@ func NewPyxisClient(pyxisHost string, apiToken string, projectId string, httpCli
 func (p *pyxisClient) createImage(ctx context.Context, certImage *CertImage) (*CertImage, error) {
 	b, err := json.Marshal(certImage)
 	if err != nil {
-		log.Error(err)
-		return nil, err
+		return nil, fmt.Errorf("could not marshal certImage: %w", err)
```
acornett21, line 49: For things that are common across all methods
- `could not marshal`
- `could not read body`
- `could not create request`
- etc

Does it make sense to have `errors` for these? Or at least constants for the strings to reduce duplication?

acornett21 (commented):
This is a great start. Just a few clarifications needed.

certification/pyxis/pyxis.go
bcrochet: If we're going to bother to add context to the error, we might as well have good info.

sebrandon1 (approved):

docs/dev/ERRORS_AND_LOGGING.md
```diff
@@ -0,0 +1,31 @@
...
+
+# Logging
+
+Logging of errors should be limited to only the upper-most layers. Otherwise, wrapped errors could be logged multiple times, causing clutter and limiting the usefulness of the logs.
+
+If an error is logged, and then returned, one should either wrap the current error
+with more context, and return that entire error, or log the wrapped error and then
+return only the wrapper, without the current error.
+
+One notable exception to this is the Builder in the pyxis package. Errors are logged but not returned. Since the errors are not wrapped or propagated up, it makes sense to log them.
```
komish, line 31: Bit of a far fetched question - do we want to enforce this somehow in GH Actions?

cmd/check_container.go
```diff
@@ -163,7 +162,7 @@ var checkContainerCmd = &cobra.Command{
 			// get the project info from pyxis
 			certProject, err := pyxisClient.GetProject(ctx)
 			if err != nil {
-				log.Error(fmt.Errorf("%w: %s", errors.ErrRetrievingProject, err))
+				log.Error(fmt.Errorf("%w: %s", ErrRetrievingProject, err))
```
komish, line 165: Is the wrap ordering incorrect here?

cmd/check_container.go
```diff
@@ -83,7 +82,7 @@ var checkContainerCmd = &cobra.Command{
 			pyxisClient := pyxis.NewPyxisClient(pyxisHost, apiToken, projectId, &http.Client{Timeout: 60 * time.Second})
 			certProject, err := pyxisClient.GetProject(ctx)
 			if err != nil {
-				log.Error(fmt.Errorf("%w: %s", errors.ErrRetrievingProject, err))
+				log.Error(fmt.Errorf("%w: %s", ErrRetrievingProject, err))
```
komish, line 85: Same here, re: wrap ordering?

cmd/check_container.go
```diff
@@ -40,7 +39,7 @@ var checkContainerCmd = &cobra.Command{
 		}
 
 		if len(args) != 1 {
-			return fmt.Errorf("%w: A container image positional argument is required", errors.ErrInsufficientPosArguments)
+			return fmt.Errorf("%w: A container image positional argument is required", ErrInsufficientPosArguments)
```
komish, line 42: Same here, re: wrap ordering?

cmd/support.go
```diff
@@ -38,7 +37,7 @@ var supportCmd = &cobra.Command{
 
 		_, certProjectTypeValue, err := certProjectTypeLabel.Run()
 		if err != nil {
-			return errors.ErrSupportCmdPromptFailed
+			return ErrSupportCmdPromptFailed
```
komish, line 40: According to the doc included here, we should be writing these errors in-line since they don't control execution, correct?

certification/pyxis/submit.go
```diff
@@ -98,8 +92,7 @@ func (p *pyxisClient) SubmitResults(ctx context.Context, certInput *certificatio
 	for _, artifact := range artifacts {
 		artifact.ImageID = certImage.ID
 		if _, err := p.createArtifact(ctx, &artifact); err != nil {
-			log.Error(fmt.Errorf("%w: could not create artifact: %s", err, artifact.Filename))
-			return nil, err
+			return nil, fmt.Errorf("could not create artifact: %s: %v", artifact.Filename, err)
```
komish, line 95: For ever "green" line in this diff, there are two "red" lines. 10/10! =)

certification/pyxis/pyxis.go
```diff
@@ -401,5 +379,5 @@ func (p *pyxisClient) newRequest(ctx context.Context, method string, url string,
 
 // checkStatus is used to check for a 2xx status code
 func checkStatus(statusCode int) bool {
-	return statusCode >= 200 && statusCode < 300
+	return statusCode >= http.StatusOK && statusCode < http.StatusMultipleChoices
```
komish, line 382: This strikes me as a super odd way to check for success. The function signature also doesn't make sense. Sorry, I know this isn't in scope for this PR, it just drew my attention to it. Might be a future quick patch.

certification/pyxis/pyxis.go
```diff
@@ -218,34 +207,32 @@ func (p *pyxisClient) getRPMManifest(ctx context.Context, imageID string) (*RPMM
...
+			"status code: %d: body: %s",
+			resp.StatusCode,
+			string(body))
 	}
 
 	var certProject CertProject
 	if err := json.Unmarshal(body, &certProject); err != nil {
-		log.Error(err)
-		return nil, err
+		return nil, fmt.Errorf("could not unmarshal body: %s: %v", string(body), err)
```
komish, line 235: Should `%v` be `%w`?

certification/pyxis/pyxis.go
```diff
@@ -218,34 +207,32 @@ func (p *pyxisClient) getRPMManifest(ctx context.Context, imageID string) (*RPMM
...
+		return nil, fmt.Errorf("could not get project from pyxis: %v", err)
 	}
 
 	defer resp.Body.Close()
 
 	body, err := io.ReadAll(resp.Body)
 	if err != nil {
-		log.Error(err, "readall failed")
-		return nil, err
+		return nil, fmt.Errorf("could not read body: %v", err)
```
komish, line 223: Should `%v` be `%w`?

certification/pyxis/layers.go
```diff
@@ -46,14 +44,13 @@ func (p *pyxisClient) CertifiedImagesContainingLayers(ctx context.Context, uncom
...
-		return nil, errors.ErrInvalidHttpClient
+		return nil, fmt.Errorf("client could not be used as http.Client")
 	}
 	client := graphql.NewClient(p.getPyxisGraphqlUrl(), httpClient).WithDebug(true)
 
 	err := client.Query(ctx, &query, variables)
 	if err != nil {
-		log.Error(fmt.Errorf("%w: %s", errors.ErrInvalidGraphqlQuery, err))
-		return nil, err
+		return nil, fmt.Errorf("could not marshal GraphQL query: %v", err)
```
komish, line 53: This `could not marshal` string could probably use some massaging. I know @acornett21 asked about this in another thread, but I'm thinking that this particular message isn't super useful to the user because it's internal to us (making a query to a remote, expecting to be able to receive an identifiable response).

komish: I don't know if there's a clear action I can recommend here.

certification/pyxis/pyxis.go
```diff
@@ -218,34 +207,32 @@ func (p *pyxisClient) getRPMManifest(ctx context.Context, imageID string) (*RPMM
...
-		return nil, err
+		return nil, fmt.Errorf("could not create new request: %v", err)
 	}
 	log.Debugf("URL is: %s", req.URL)
 
 	resp, err := p.Client.Do(req)
 	if err != nil {
-		log.Error(err, "client.Do failed")
-		return nil, err
+		return nil, fmt.Errorf("could not get project from pyxis: %v", err)
```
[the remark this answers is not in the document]
komish: So I've gone through and deleted about 5-6 more of these that I saw in this PR. I'm thinking your intention was not to wrap, but just pass through the raw value of the error into the message. If that's not correct, then there are another few `%v`s that need to be updated. If you just want to pass the value, I'm wondering why not use %s, given that I believe all errors can be `.String()`ed.

komish (changes requested):
Couple of comments here for discussion.

certification/pyxis/pyxis.go
bcrochet: And I meant "could not marshal certImage". "could not marshal foo". "could not marshal bar". I'm ok with it as is. If you feel strongly we can discuss further.

docs/dev/ERRORS_AND_LOGGING.md
bcrochet: Maybe? Seems like a bit of a stretch though. Automated that is. I think to start we as human reviewers have to be more diligent and pay attention to the errors being bandied about.

cmd/check_container.go
bcrochet: Probably not. I hadn't gotten to the "fix all the wrapping" phase. Otherwise, this PR would likely touch literally every file. I'm not looking to solve all the things in this particular PR.

cmd/check_container.go
bcrochet: See previous comment. :)

cmd/support.go
bcrochet: Again, this package has not gotten that treatment yet. The main focus of this particular PR is to be rid of the global "errors" package. #654 begins to address this. There will be others. Trying to get a big PR knocked out, that we can then have subsequent PRs that go package or two at a time, and don't have to rely on each other.

komish (approved):
/lgtm

PR #654, opened 2022-05-24
Normalize error handling

bcrochet:
Fix the error handling in three packages:
* artifacts
* external engine
* formatters
    
Just going alphabetical through the packages and normalizing error
handling, and removing errors.go from packages when and where
appropriate.
    
### DIFF
--- a/certification/artifacts/artifacts.go
+++ b/certification/artifacts/artifacts.go
@@ -18,7 +18,7 @@ func WriteFile(filename, contents string) (string, error) {
 
 	err := os.WriteFile(fullFilePath, []byte(contents), 0o644)
 	if err != nil {
-		return fullFilePath, err
+		return fullFilePath, fmt.Errorf("could not write file to artifacts diretory: %v", err)
 	}
 	return fullFilePath, nil
 }
@@ -27,8 +27,7 @@ func createArtifactsDir(artifactsDir string) (string, error) {
 	if !strings.HasPrefix(artifactsDir, "/") {
 		currentDir, err := os.Getwd()
 		if err != nil {
-			log.Error(fmt.Errorf("unable to get current directory: %w", err))
-			return "", err
+			return "", fmt.Errorf("unable to get current directory: %w", err)
 		}
 
 		artifactsDir = filepath.Join(currentDir, artifactsDir)
@@ -36,8 +35,7 @@ func createArtifactsDir(artifactsDir string) (string, error) {
 
 	err := os.MkdirAll(artifactsDir, 0o777)
 	if err != nil {
-		log.Error(fmt.Errorf("unable to create artifactsDir: %w", err))
-		return "", err
+		return "", fmt.Errorf("unable to create artifactsDir: %w", err)
 	}
 	return artifactsDir, nil
 }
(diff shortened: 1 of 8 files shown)


bcrochet:
Hold until #652 lands

certification/artifacts/artifacts.go
```diff
@@ -17,7 +18,7 @@ func WriteFile(filename, contents string) (string, error) {
 
 	err := os.WriteFile(fullFilePath, []byte(contents), 0o644)
 	if err != nil {
-		return fullFilePath, err
+		return fullFilePath, fmt.Errorf("could not write file to artifacts diretory: %v", err)
```
acornett21, line 21: Should this be `%w` for consistency? `%v and %w` are synonyms, but maybe we should pick one and stick with it?

certification/formatters/generic.go
```diff
@@ -39,8 +37,7 @@ func genericXMLFormatter(ctx context.Context, r runtime.Results) ([]byte, error)
 
 	responseJSON, err := xmlMarshalIndent(response, "", "    ")
 	if err != nil {
-		e := fmt.Errorf("%w with formatter %s: %s",
-			errors.ErrFormattingResults,
+		e := fmt.Errorf("error formatting results with formatter %s: %w",
 			"json",
```
acornett21, line 41: Should this be `xml`?

komish: also `responseXML?`

certification/internal/engine/engine.go
```diff
@@ -327,27 +326,27 @@ func untar(dst string, r io.Reader) error {
...
 	manifest, err := imageRef.ImageInfo.Manifest()
 	if err != nil {
-		return fmt.Errorf("%w: %s", errors.ErrImageInspectFailed, err)
+		return fmt.Errorf("%w: %s", ErrImageInspectFailed, err)
 	}
 
 	digest, err := imageRef.ImageInfo.Digest()
 	if err != nil {
-		return fmt.Errorf("%w: %s", errors.ErrImageInspectFailed, err)
+		return fmt.Errorf("%w: %s", ErrImageInspectFailed, err)
```
komish, line 339: Does this need to be `("%s: %w", ...)`? Or was that done in another PR?

komish (commented):
Small things.

certification/artifacts/artifacts.go
acornett21: `%v` means `default` so I assume in the [docs](https://pkg.go.dev/fmt#Errorf) just mean `%w` is the same as `default`, at least that's how i read it.

bcrochet: They aren't exactly synonyms. %w allows for unwrapping the error, whereas %v does not. Not sure what you mean by default.

certification/formatters/generic.go
bcrochet: That looks like that's been there a while. I'll fix though.

certification/artifacts/artifacts.go
bcrochet: And %v was chosen precisely so that the error type doesn't leak out of the package. If %w is returned from a package export, it will end up being a part of the API, and we do not want to do that unless we are explicit about it.

certification/formatters/generic.go
acornett21: Thanks for cleaning this up too!

acornett21 (approved):

PR #680, opened 2022-05-27
Enforce a single point of configuration

### LINKED ISSUE #641: Enforce a single-point-of-configuration
### Problems

- We rely on Viper’s global configuration which allows us to use a magic, internal configuration Singleton to get values.

- We (can) also set values in certain cases, which technically means that our runtime configuration can change from anywhere in our code.
Ideally, we should not be changing our runtime configuration from within the code.

- The most common location where we use this is in our Check execution because checks do NOT have access to configuration (the interface signature does not allow for it).

### The solution should:

- avoid modifying the configuration in its entirety. Accessing values seems fine, but I do think Preflight would benefit from restructuring how it interacts with configuration such that nothing is needed from the configuration at execution time. That is, all of the information necessary to execute, from top to bottom, is passed into Preflight’s core runtime logic as either functional parameters or struct properties so that the global configuration does not need to be accessed.

---
Challenges already exist with this today, in part, because the Check interface definition is designed to be entirely standalone. That is, the [Check](https://github.com/redhat-openshift-ecosystem/openshift-preflight/blob/main/certification/certification.go#L5-L17) interface has a Validate method which does not accept configuration information. This means that we must initialize our Check implementations with all of the information necessary to execute when we instantiate them. There are some examples where we do this today, but doing this for other checks may pose some logistical problems.

komish:
Fixes #641 

This PR reconfigures Preflight execution to utilize a single point of configuration, which is then passed through to Engines in a read-only state.

Fill disclosure, I'm still testing behavior here, but given the size, figured it wise to get eyes on things to make sure we're happy with the approaches/changes.

Notable high-level benefits include:

- Lazy initialization of checks at execution time (e.g. in preparation of the engine.)
- Configuration pass through from user all the way down to checks.
- Reduced footprint of the spf13/viper as an import. We heavily leverage it as an integration to our flag parsing, environment variable parsing, etc., but it exists in much less of our code outside of the `cmd` package.
- Inability to change configuration at any point behind the calling of `ExecuteChecks`.
- Removed reliance on environment manipulation in testing (e.g. to establish artifacts dir, set logfiles directories, etc).

This is a large PR. Viper was so deeply rooted in most of our code due to the need to access the user configuration that pulling it out often meant passing in additional data to functions/methods/structs that were previously using viper.

A lot of things changed to make this work. The spirit of most tests was kept intact (if possible) in order to reduce the likelihood that I changed a test enough to not catch an error. Still possible, but that was the goal.

Scrutinize heavily. Here are some details on things that have changed and what motivated those changes:

---

**New: `certification.Config` interface:**

This interface definition defines all the methods necessary to execute Preflight. Previously, we relied on `runtime.Configuration{}` for a similar purpose. `runtime.Config{}` now will fulfill the `certification.Config` interface when calling its `runtime.Config{}.ReadOnly()` method.

In theory, I could have just bound these methods to `runtime.Config{}`, because accepting the `certification.Config{}` interface would make access read-only in a sense, but I ran into issues with Method name and struct field collision, so it just seemed easier to roll it into its own type. 

Organizationally, this is placed in the `certification` package because I wasn't sure if there was a better place. I think, ideally, that we would have defined `runtime.Config` to be this interface, and had our implementation maybe live in the `cmd` package, but I'm not sure we want to move that around currently. 

Many functions that previously accepted `runtime.Config{}` now accept the `certification.Config` interface instead.

**Derived `runtime.Config` from viper**
To avoid fully removing Viper, which wires up our environment variables, binds with our flags, etc., I've derived our `runtime.Config` from the viper configuration using by function `runtime.NewConfigFrom(...)`.

**Removal of global check variables in favor of lazy initialization**

We previously had our checks defined globally within the `engine` package as variables. This limited our ability to configure them at runtime, leading to the use of `viper.Get<Type>(...)` to configure them within their own executions. This has been removed. Checks are now initialized within an `initializeChecks` function that accepts a `certification.Config`. This config can then inform how the checks are configured and returned to the engine for execution.

**Removal of `EnabledChecks` from runtime.Config**
In a prior use case, we had the ability to dynamically change the checks that were executed based on user configuration. We disabled part of this feature in https://github.com/redhat-openshift-ecosystem/openshift-preflight/pull/26 in favor of ensuring that users get the appropriate checks for what they're trying to certify. We had some leftover configuration from these days that are now removed, such as `runtime.Config.EnabledChecks`.

We now have a Policy type that maps to our existing policies, which we provide to our check initialization logic in our configuration.

Logic associated with asserting that the EnabledChecks in the configuration _matches_ what we know to have has been removed as unnecessary.

**Refactor the `artifacts` package into a Singleton**

The `artifacts` package was previously used to facilitate writing artifacts by not requiring the caller to know about the artifacts configuration. It did this by effectively calling viper to get the value for the `"artifacts"` key, which generally pointed to the artifacts director (either by default, or by user configuration). This package has been reconfigured to allow the same functionality by using a package-level variable.

Note that it is the responsibility of the caller to configure the new artifacts directory if the user has requested such by way of a new function `SetDir`. A `Reset` function has also been provided to return this back to default configuration (in my mind, to facilitate testing).

**Extraction of Operator and Container check logic into functions**

New functions `checkOperator` and `checkContainer` have been created, now containing the logic that previously existed within each respective cobra command's `RunE` field.

Note that these still work with concrete `runtime.Config` because they _must_ set the appropriate policy (and potentially other fields) before they execute. They pass along `certification.Config`s to the appropriate downstream methods.

This should pave the way to extracting this logic for library use in a future PR. I didn't include it here because... well, this is already massive. And also, we may want to evalute breaking them into separate, logical components as well. Story for another day.

**Authn Keychain refactor into a singleton**
The `authn` package which contains our implementation of the `craneauthn.Keychain` and related types has been refactored into a singleton as well. This is because it previously made a call out to viper to grab `dockerconfig` in its `Resolve()` chain. It's design to be initialized once, and then it doesn't have to be initialized again further downstream. That said, it totally can be.

At a high level, we were previously able to just call authn.PreflightKeychain which was a variable containing our keychain, and the `Resolve` logic read the viper configuration for a docker config.

This must now be initialized (as is being done in engine calls that set it up) with the `WithDockerConfig` option containing the dockerconfig provided by the user. The option pattern was used here because it allows downstream calls (e.g. calls after it has been initially configured) to call `PreflightKeychain()` completely without parameters. With that said, if downstream calls are passed the same parameter for dockerconfig as is in the certification.Config, subsequent calls shouldn't change anything.

**Internal crane engine has access to certification.Config**
Which means that we can technically _remove_ some fields of the internal crane engine, such as `Image`. That said, I left that for a future PR, as I want to make sure this works before we start ripping things out. That felt like a small self-encompassed change that could be whipped up and added later.

**OperatorSDK engine(s) no longer require viper**
We _heavily_ used viper in OperatorSDK based checks (think scorecard, deployablyByOLM). This is no longer the case. OperatorSDK initializes (e.g. NewOperatorSdkEngine) now accept the parameters they need when initialized. These have a lot of changes, and most of them are related to viper removal and replacement with config values.

**New `Policy` package**
Simply contains policy definitions in a common place that can be imported by many different packages without introducing cyclical import errors.

### DIFF
--- a/certification/artifacts/artifacts_test.go
+++ b/certification/artifacts/artifacts_test.go
@@ -5,7 +5,6 @@ import (
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
-	"github.com/spf13/viper"
 )
 
 var _ = Describe("Artifacts package utility functions", func() {
@@ -14,8 +13,8 @@ var _ = Describe("Artifacts package utility functions", func() {
 		// in each of these tests. This removes only the artifacts
 		// directory value, not the temporary dir established in
 		// BeforeSuite.
-		err := os.RemoveAll(viper.GetString("artifacts"))
-		Expect(err).ToNot(HaveOccurred())
+
+		Expect(os.RemoveAll(ads)).To(Succeed())
 	})
 
 	Context("With an artifacts directory provided via configuration", func() {
--- a/certification/config.go
+++ b/certification/config.go
@@ -1,0 +1,45 @@
+package certification
+
+import "github.com/redhat-openshift-ecosystem/openshift-preflight/certification/policy"
+
+// Config is a read-only preflight configuration.
+type Config interface {
+	commonConfig
+	containerConfig
+	operatorConfig
+}
+
+// commonConfig contains configurables common
+// to all certification.
+type commonConfig interface {
+	Image() string
+	Policy() policy.Policy
+	ResponseFormat() string
+	LogFile() string
+	Artifacts() string
+	WriteJUnit() bool
+	DockerConfig() string
+}
+
+// containerConfig are configurables relevant to
+// container certification.
+type containerConfig interface {
+	IsScratch() bool
+	CertificationProjectID() string
+	PyxisHost() string
+	PyxisAPIToken() string
+	Submit() bool
+}
+
+// operatorConfig are configurables relevant to
+// operator certification.
+type operatorConfig interface {
+	IsBundle() bool
+	Namespace() string
+	ServiceAccount() string
+	ScorecardImage() string
+	ScorecardWaitTime() string
+	Channel() string
+	Kubeconfig() string
+	IndexImage() string
+}
--- a/certification/engine/engine.go
+++ b/certification/engine/engine.go
@@ -11,9 +11,9 @@ import (
 	internal "github.com/redhat-openshift-ecosystem/openshift-preflight/certification/internal/engine"
 	containerpol "github.com/redhat-openshift-ecosystem/openshift-preflight/certification/internal/policy/container"
 	operatorpol "github.com/redhat-openshift-ecosystem/openshift-preflight/certification/internal/policy/operator"
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/policy"
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/pyxis"
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/runtime"
-	"github.com/spf13/viper"
 )
 
 // CheckEngine defines the functionality necessary to run all checks for a policy,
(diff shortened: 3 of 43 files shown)


tkrishtop:
hi @komish I see two errors in the logs of [DCI job](https://www.distributed-ci.io/jobs/f21344eb-41aa-4d04-a23b-8224171a7056/jobStates) 

1/ JUnit files with the results are not generated anymore
2/ It seems like the way to pass registry credentials was changed for `check operator` (but not for `check container`)

Here is the command I run:
```
podman run --rm 
--pull=always 
--privileged 
-e PFLT_JUNIT=true 
-e PFLT_ARTIFACTS=/artifacts 
-e PFLT_LOGFILE=/artifacts/preflight.log 
-e PFLT_LOGLEVEL=trace 
-e DOCKER_CONFIG=/opt 
-e PFLT_DOCKERCONFIG=/opt/config.json 
-e KUBECONFIG=/kubeconfig 
-e PFLT_INDEXIMAGE=registry.dfwt5g.lab:4443/preflight/disconnected-catalog@sha256:4d40749b713a5cc4295812eedb69920a08af9975b88ce9bab59bb7cc0c6eef60 
-e PFLT_NAMESPACE=preflight-testing 
-e PFLT_SERVICEACCOUNT=default 
-e PFLT_SCORECARD_IMAGE=quay.io/operator-framework/scorecard-test@sha256:c20b539ac6887f93175511da559f29a6b7601ed11826db32f04cb466e275918d 
-v /tmp/preflight_tmp_dir.15w44fzo/kubeconfig:/kubeconfig 
-v /tmp/preflight_operator_artifacts.t1vod2c8:/artifacts 
-v /tmp/preflight_tmp_dir.15w44fzo/config.json:/opt/config.json 
registry.dfwt5g.lab:4443/preflight/preflight:pr680 
check operator 
quay.io/telcoci/simple-demo-operator-bundle@sha256:6cfbca9b14a51143cfc5d0d56494e7f26ad1cd3e662eedd2bcbebf207af59c86
```

and it results in the following error:
```
failed to pull remote container: GET https://quay.io/v2/telcoci/simple-demo-operator-bundle/manifests/sha256:6cfbca9b14a51143cfc5d0d56494e7f26ad1cd3e662eedd2bcbebf207af59c86: UNAUTHORIZED: access to the requested resource is not authorized
```

Do you have any input please, maybe DCI integration should be changed?

itroyano:
One potential use case - loading a configuration file (yaml or something else).
e.g. the pipeline task running preflight to be able to mount a config map (instead of passing params to Viper's parsing).

komish:
@tkrishtop Thanks for checking, reviewing the DCI failures was on my list. I'll take a look at what you've noted.

komish:
@itroyano in theory that's possible today because we enable Viper's config parsing. I don't believe we accept a flag for it, so it just has to be the right file name in the working directory. Haven't tested this myself.

komish:
@tkrishtop I'm not sure I'm following re: docker config. I don't _think_ I've changed how it's read. It's also bound as a persistent flag on the `check` command so it should be utilized by both operator and container.

I do see that you're exporting `DOCKER_CONFIG`, and I think we stopped using that in a recent PR. @bcrochet or @acornett21 to confirm.

acornett21:
> I do see that you're exporting `DOCKER_CONFIG`, and I think we stopped using that in a recent PR. @bcrochet or @acornett21 to confirm.

This is correct we use the same `PFLT_DOCKERCONFIG` env for both `check container` and `check operator` since this is code is common in the engine for pulling images.

komish:
@tkrishtop Can you link me the error you see? I poked around the results but I couldn't quite find an error other than the junit.

cmd/check_container.go
```diff
@@ -62,218 +63,214 @@ var checkContainerCmd = &cobra.Command{
...
-		results := engine.Results(ctx)
+	// also write to stdout
+	resultsOutputTarget := io.MultiWriter(os.Stdout, resultsFile)
 
-		// return results to the user and then close output files
-		formattedResults, err := formatter.Format(ctx, results)
-		if err != nil {
-			return err
-		}
+	// execute the checks
```
bcrochet, line 139: Here is where the cmd.SilenceUsage was previously. If it is not here, it doesn't have the same functionality.

cmd/check_operator.go
```diff
@@ -57,67 +58,80 @@ var checkOperatorCmd = &cobra.Command{
...
-			return err
-		}
-
-		// also write to stdout
-		resultsOutputTarget := io.MultiWriter(os.Stdout, resultsFile)
-
-		// At this point, we would no longer want usage information printed out
-		// on error, so it doesn't contaminate the output.
+		// Run the operator check
 		cmd.SilenceUsage = true
```
bcrochet, line 72: See the commends in check_container.go about this.

bcrochet (commented):
A few comments to start.

certification/artifacts/artifacts_test.go
```diff
@@ -14,7 +13,7 @@ var _ = Describe("Artifacts package utility functions", func() {
 		// in each of these tests. This removes only the artifacts
 		// directory value, not the temporary dir established in
 		// BeforeSuite.
-		err := os.RemoveAll(viper.GetString("artifacts"))
+		err := os.RemoveAll(ads)
```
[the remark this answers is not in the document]
komish: +1

cmd/check_container.go
```diff
@@ -62,218 +63,214 @@ var checkContainerCmd = &cobra.Command{
...
-				log.Error(fmt.Errorf("%w: %s", ErrRetrievingProject, err))
-				return err
-			}
-			log.Debugf("Certification project name is: %s", certProject.Name)
-			if certProject.Container.OsContentType == "scratch" {
-				cfg.EnabledChecks = engine.ScratchContainerPolicy()
-				cfg.Scratch = true
-			}
+		// Run the  container check.
+		cmd.SilenceUsage = true
```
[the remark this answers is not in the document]
komish: Yep. I understand that, but the only way to implement that with the new model is either to (a) break the code that happened before this previously into its own function and run it in RunE, or accept the `cmd`'s type in check<Target> funcs. This seemed compromise-ish.

certification/config.go
```diff
@@ -0,0 +1,45 @@
...
+// commonConfig contains configurables common
+// to all certification.
+type commonConfig interface {
+	Image() string
+	Policy() policy.Policy
+	ResponseFormat() string
+	LogFile() string
+	Artifacts() string
+	WriteJUnit() bool
+	DockerConfig() string
```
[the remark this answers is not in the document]
komish: Hmm, I'm not sure. This is storing configuration values, right? It's not storing the implementation details, just what the user preference was.

certification/config.go
```diff
@@ -0,0 +1,45 @@
...
+// operatorConfig are configurables relevant to
+// operator certification.
+type operatorConfig interface {
+	IsBundle() bool
+	Namespace() string
+	ServiceAccount() string
+	ScorecardImage() string
+	ScorecardWaitTime() string
+	Channel() string
+	Kubeconfig() string
```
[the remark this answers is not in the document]
komish: Discussing this in the previous thread

certification/internal/policy/operator/operator_pkg_name_uniqueness.go
```diff
@@ -34,10 +33,18 @@ type packageData struct {
 	Source         string `json:"source"`
 }
 
-// OperatorPkgNameIsUniqueMountedCheck finds the package name as defined in the operator bundle's annotations
+// OperatorPkgNameIsUniqueCheck finds the package name as defined in the operator bundle's annotations
```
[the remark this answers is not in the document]
komish: I'll remove changes from this file.

cmd/check_container.go
komish: Continuing this discussion in a previous thread.

cmd/check_operator.go
komish: Continuing this discussion in a previous thread.

certification/internal/policy/operator/operator_pkg_name_uniqueness.go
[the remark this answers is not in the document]
komish: I can remove changes from this file as long as it builds.

certification/internal/policy/container/has_unique_tag.go
```diff
@@ -30,7 +34,7 @@ func (p *hasUniqueTagCheck) Validate(ctx context.Context, imgRef certification.I
 func (p *hasUniqueTagCheck) getDataToValidate(ctx context.Context, image string) ([]string, error) {
 	options := []crane.Option{
 		crane.WithContext(ctx),
-		crane.WithAuthFromKeychain(authn.PreflightKeychain),
+		crane.WithAuthFromKeychain(authn.PreflightKeychain()),
```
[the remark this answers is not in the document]
komish: Fixed.

certification/artifacts/artifacts_test.go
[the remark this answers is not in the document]
komish: Fixed

certification/engine/engine_test.go
```diff
@@ -3,82 +3,24 @@ package engine
...
-	})
-})
-
 var _ = Describe("Engine Creation", func() {
 	Describe("When getting a new engine for a configuration", func() {
 		Context("with a valid configuration", func() {
 			cfg := runtime.Config{
 				Image:          "dummy/image",
-				EnabledChecks:  ContainerPolicy(),
+				Policy:         policy.PolicyContainer,
```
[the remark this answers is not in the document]
komish: Fixed!

cmd/check_container.go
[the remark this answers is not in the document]
bcrochet: The more I think about it, the more I think that the outcome of how you implemented it is actually a better UX. We probably only want usage printed if there is an Args problem anyway.

tkrishtop:
Hi @komish the full logs for `check operator` run are [here](https://www.distributed-ci.io/jobs/3890cc7f-333d-4ffe-94d1-51ffb2be6c57/jobStates#3145c7e3-e4d7-449b-ad7a-1c51bea43797:file90).

They're ignored to give a possibility to load files like preflight.log in /files tab. Here is preflight.log: preflight_operator_simple-demo-operator_0.0.6_preflight.log 

```
time="2022-05-31T21:12:37Z" level=debug msg="config file not found, proceeding without it"
time="2022-05-31T21:12:37Z" level=info msg="certification library version pr680 <commit: 869492776d0da30d0ed6e41a92e3bcb95f064859>"
time="2022-05-31T21:12:37Z" level=debug msg="target image: quay.io/telcoci/simple-demo-operator-bundle@sha256:6cfbca9b14a51143cfc5d0d56494e7f26ad1cd3e662eedd2bcbebf207af59c86"
time="2022-05-31T21:12:37Z" level=debug msg="pulling image from target registry"
time="2022-05-31T21:12:37Z" level=trace msg="entering preflight keychain Resolve"
time="2022-05-31T21:12:38Z" level=fatal msg="failed to pull remote container: GET https://quay.io/v2/telcoci/simple-demo-operator-bundle/manifests/sha256:6cfbca9b14a51143cfc5d0d56494e7f26ad1cd3e662eedd2bcbebf207af59c86: UNAUTHORIZED: access to the requested resource is not authorized; map[]"
```

certification/engine/engine.go
```diff
@@ -28,146 +28,111 @@ type CheckEngine interface {
...
+			&containerpol.HasNoProhibitedPackagesCheck{},
+			&containerpol.HasRequiredLabelsCheck{},
+			&containerpol.RunAsNonRootCheck{},
+			&containerpol.HasModifiedFilesCheck{},
+			containerpol.NewBasedOnUbiCheck(pyxis.NewPyxisClient(
+				certification.DefaultPyxisHost,
+				cfg.PyxisAPIToken(),
+				cfg.CertificationProjectID(),
+				&http.Client{Timeout: 60 * time.Second})),
+			// containerpol.NewRunnableContainerCheck(internal.NewPodmanEngine())
```
acornett21, line 72: We can probably remove these, I think @bcrochet is removing them in another PR.

komish: Better to remove them in one place, and rebasing I think?

komish:
Currently resolving an issue with `--list-checks`. Push incoming.

komish:
Latest change 24857f2 is a one-line fix resolving an issue in `makeCheckList(...)` that was preventing us from listing checks when passed `--list-checks`.

komish:
Latest push d549453  rebases on recently merged changes.

certification/internal/policy/operator/operator_pkg_name_uniqueness.go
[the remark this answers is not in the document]
bcrochet: You can rebase now.

acornett21 (approved):

certification/engine/engine.go
```diff
@@ -28,137 +28,107 @@ type CheckEngine interface {
...
-	for _, checkString := range config.EnabledChecks {
-		check := queryChecks(checkString)
-		if check == nil {
-			err := fmt.Errorf("requested check not found: %s", checkString)
-			return nil, err
-		}
-
-		checks = append(checks, check)
+func NewForConfig(cfg certification.Config) (CheckEngine, error) {
+	checks, err := initializeChecks(context.Background(), cfg.Policy(), cfg)
```
bcrochet, line 32: Context should be passed in here. We are now creating the context in cmd/root.go. Anything that needs a context now should be getting it from there.

bcrochet (changes requested):

certification/engine/engine.go
komish: Done. I passed `context.TODO()` in tests, and `cmd.Context()` in other cases.

komish:
@bcrochet Rebase complete, tests are running. Please have a look to make sure I didn't trample any changes from your merged PRs. Particularly in check_container.go because its generated diff was a mess for some reason.

certification/internal/policy/operator/operator_pkg_name_uniqueness.go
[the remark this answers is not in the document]
komish: Done

certification/engine/engine.go
```diff
@@ -28,137 +28,107 @@ type CheckEngine interface {
...
+			&containerpol.MaxLayersCheck{},
+			&containerpol.HasNoProhibitedPackagesCheck{},
+			&containerpol.HasRequiredLabelsCheck{},
+			&containerpol.RunAsNonRootCheck{},
+			&containerpol.HasModifiedFilesCheck{},
+			containerpol.NewBasedOnUbiCheck(pyxis.NewPyxisClient(
+				certification.DefaultPyxisHost,
+				cfg.PyxisAPIToken(),
+				cfg.CertificationProjectID(),
+				&http.Client{Timeout: 60 * time.Second})),
```
bcrochet, line 71: Since we have a context, it should probably be used here, but that can be a later PR.

cmd/check_operator.go
```diff
@@ -31,19 +33,19 @@ var checkOperatorCmd = &cobra.Command{
...
 
 		if len(args) != 1 {
-			return fmt.Errorf("An operator image positional argument is required")
+			return fmt.Errorf("an operator image positional argument is required")
 		}
 		return nil
 	},
 	// this fmt.Sprintf is in place to keep spacing consistent with cobras two spaces that's used in: Usage, Flags, etc
 	Example: fmt.Sprintf("  %s", "preflight check operator quay.io/repo-name/operator-bundle:version"),
+	PreRun:  preRunConfig,
```
bcrochet, line 42: This shouldn't be here.

cmd/check_test.go
```diff
@@ -117,24 +117,17 @@ var _ = Describe("cmd package check command", func() {
...
-				viper.Set("junit", false)
-				Expect(writeJUnit(context.TODO(), *results)).To(Succeed())
-				_, err := os.Stat(junitfile)
-				Expect(errors.Is(err, os.ErrNotExist)).To(BeTrue())
-			})
-		})
-		When("junit is enabled", func() {
-			It("should return no error and file should be created", func() {
-				viper.Set("junit", true)
+		When("The additional JUnitXML results file is erquested", func() {
```
bcrochet, line 129: "requested"

cmd/root.go
```diff
@@ -39,54 +36,3 @@ func Execute() {
 		os.Exit(-1)
 	}
 }
```
bcrochet, line 38: This should have stayed here? This was just moved here.

cmd/utils.go
```diff
@@ -0,0 +1,96 @@
+package cmd
+
+import (
+	"fmt"
```
bcrochet, line 4: This file shouldn't be here.

bcrochet (changes requested):
A few things based on the rebase.

cmd/check_test.go
komish: Fixed

cmd/check_operator.go
komish: Fixed. It wasn't flagged as removed in my diff, and was just hanging out there in the wrong place.

cmd/root.go
komish: Fixed.

cmd/utils.go
komish: Fixed.

bcrochet (approved):
/lgtm

PR #689, opened 2022-06-08
Add FindImagesByDigest to the pyxis client

bcrochet:
* Add a method to the Pyxis client that accepts a slice of digests,
  and returns a slice of CertImages.
* So far, it only fills in the ID and Certified fields of the image.

This will be used in the upcoming CsvImagesCertified check.

### DIFF
--- a/certification/pyxis/pyxis.go
+++ b/certification/pyxis/pyxis.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	"net/http"
 
+	"github.com/shurcooL/graphql"
 	log "github.com/sirupsen/logrus"
 )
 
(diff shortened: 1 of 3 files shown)


certification/pyxis/pyxis.go
```diff
@@ -126,6 +127,59 @@ func (p *pyxisClient) getImage(ctx context.Context, dockerImageDigest string) (*
...
+	err := client.Query(ctx, &query, variables)
+	if err != nil {
+		return nil, fmt.Errorf("error while executing find_images query: %v", err)
+	}
+
+	images := make([]CertImage, 0, len(query.FindImages.ContainerImage))
+	for _, image := range query.FindImages.ContainerImage {
+		images = append(images, CertImage{
+			ID:        string(image.ID),
+			Certified: bool(image.Certified),
```
acornett21, line 176: I think it might be usefully to also add the `DockerImageDigest` since in our code the `ID` has no meaning. I'm thinking logging in the check, and things of that nature that it would be helpful for.

certification/pyxis/pyxis.go
```diff
@@ -126,6 +127,59 @@ func (p *pyxisClient) getImage(ctx context.Context, dockerImageDigest string) (*
 	return &data.Data[0], nil
 }
 
+// FindImagesByDigest uses an unauthenticated call to find_images() graphql function, and will
+// return a slice of CertImages. It accepts a slice of image digests. Additional fields
+// for return should be specified in ContainerImage
```
komish, line 132: Two things:

1) Should `s/ContainerImage/CertImage/?` in this comment? 

2) Could this be reword to include how we go from `[]digests` to `[]CertImage`? Right now it reads to me as an explanation of the method signature more than what the method accomplishes.

certification/pyxis/pyxis_suite_test.go
```diff
@@ -41,11 +41,13 @@ func mustWrite(w io.Writer, s string) {
...
-	pyxisRPMManifestHandler  struct{}
-	pyxisTestResultsHandler  struct{}
-	pyxisGraphqlLayerHandler struct{}
+	pyxisProjectHandler           struct{}
+	pyxisImageHandler             struct{}
+	pyxisRPMManifestHandler       struct{}
+	pyxisTestResultsHandler       struct{}
+	pyxisGraphqlLayerHandler      struct{}
+	pyxisGraphqlFindImagesHandler struct{}
+	errorHandler                  struct{}
```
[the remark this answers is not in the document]
acornett21: That got me too, I was like `wait what all is getting removed` but it's really just two additions.

certification/pyxis/pyxis.go
```diff
@@ -126,6 +127,59 @@ func (p *pyxisClient) getImage(ctx context.Context, dockerImageDigest string) (*
...
+func (p *pyxisClient) FindImagesByDigest(ctx context.Context, digests []string) ([]CertImage, error) {
+	// our graphQL query
```
komish, line 134: Should we bail early if `digests` is empty?

komish (commented):
Couple of things

certification/pyxis/pyxis.go
bcrochet: That's a great suggestion.

certification/pyxis/pyxis.go
```diff
@@ -126,6 +127,59 @@ func (p *pyxisClient) getImage(ctx context.Context, dockerImageDigest string) (*
...
+			Total graphql.Int
+			Page  graphql.Int
+			// filter to make sure we get exact results
+		} `graphql:"find_images(filter: {docker_image_digest:in:$digests})"`
+	}
+
+	graphqlDigests := make([]graphql.String, 0, len(digests))
+	for _, digest := range digests {
+		graphqlDigests = append(graphqlDigests, graphql.String(digest))
+	}
```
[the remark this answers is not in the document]
bcrochet: I would need to find something to back this up, but it feels more idiomatic this way.

certification/pyxis/pyxis.go
bcrochet: Sure. I can add that.

certification/pyxis/pyxis.go
bcrochet: CertImages is correct. ContainerImage is the internal name given for the graphql query.

And yes, I'll take a stab at a reword.

certification/pyxis/pyxis.go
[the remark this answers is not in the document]
acornett21: I think doing `append` is the idiomatic way, `go`'s own code is littered with this pattern, I don't really see them keeping track of indexes much, if at all.

komish (approved):
/lgtm

certification/pyxis/pyxis.go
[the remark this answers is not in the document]
bcrochet: In this case, I changed it, since there is a 1:1 correlation between the source and the destination. If there was a possible difference in size, I would argue for the other for sure. But here, I'm good either way.

jomkz (approved):

PR #690, opened 2022-06-08
Add capability to extract images from a bundle

bcrochet:
* ExtractImagesFromBundle will pull all of the images out of a
  bundle dir in the ClusterServiceVersion. It reads both the
  container section and the relatedImages section.
* It dedupes the slice along the way.

### DIFF
--- a/certification/internal/bundle/bundle.go
+++ b/certification/internal/bundle/bundle.go
@@ -10,9 +10,10 @@ import (
 	"strings"
 
 	"github.com/blang/semver"
+	operatorv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/internal/cli"
 	log "github.com/sirupsen/logrus"
-	"gopkg.in/yaml.v2"
+	"sigs.k8s.io/yaml"
 )
 
 const ocpVerV1beta1Unsupported = "4.9"
@@ -179,7 +180,7 @@ func GetCsvFilePathFromBundle(mountedDir string) (string, error) {
 }
 
 func GetSupportedInstallModes(ctx context.Context, csvReader io.Reader) (map[string]bool, error) {
-	var csv ClusterServiceVersion
+	var csv operatorv1alpha1.ClusterServiceVersion
 	bts, err := io.ReadAll(csvReader)
 	if err != nil {
 		return nil, fmt.Errorf("could not get CSV from reader: %v", err)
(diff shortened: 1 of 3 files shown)


certification/internal/bundle/bundle.go
```diff
@@ -198,16 +198,66 @@ func GetSupportedInstallModes(ctx context.Context, csvReader io.Reader) (map[str
...
+	if err != nil {
+		return nil, fmt.Errorf("could not read CSV from path: %v", err)
+	}
+
+	var csv ClusterServiceVersion
+	if err := yaml.Unmarshal(csvBytes, &csv); err != nil {
+		return nil, fmt.Errorf("could not unmarshal CSV: %v", err)
+	}
+
+	imageMap := make(map[string]struct{}, len(csv.Spec.RelatedImages))
```
acornett21, line 212: Do we know if the pipeline updates the `relatedImages` section in the bundle that we get? Or do they just update it for their own use, and we get the raw csv that the partner put in their PR that might not have `relatedImages`?

certification/internal/bundle/bundle.go
```diff
@@ -198,16 +198,66 @@ func GetSupportedInstallModes(ctx context.Context, csvReader io.Reader) (map[str
...
+					Spec struct {
+						Containers []struct {
+							Image string `yaml:"image,omitempty"`
+						} `yaml:"containers,omitempty"`
+					} `yaml:"spec,omitempty"`
+				} `yaml:"template,omitempty"`
+			} `yaml:"spec,omitempty"`
+		} `yaml:"deployments,omitempty"`
+	} `yaml:"spec,omitempty"`
+}
```
komish, line 263: At this point, why don't we just start using the upstream API definitions? I understand not using it when we only care about very specific fields, and fewer of the structs, but we're adding more and more here. 

https://github.com/operator-framework/api/blob/master/pkg/operators/v1alpha1/clusterserviceversion_types.go

bcrochet: We're already pulling in that dep, so probably not a bad idea.

komish:
Also worth mentioning as food for thought. As we start to move to confirm images within bundles are certified, we'll have to figure out a new "happy path" operator that's certified. Simple Demo Operator is not (for obvious reasons).

certification/internal/bundle/bundle.go
acornett21: Yeah I have no idea either, I was just wondering, since this could be the length of RelatedImages could be zero. Not really a blocker for this PR or a code change, but we should probably know that when we go to add this check in and fail certification.

acornett21 (approved):

komish (approved):
/lgtm

PR #691, opened 2022-06-08
Add coverage for certification/internal/rpm

### LINKED ISSUE #618: Test coverage improvement: github.com/redhat-openshift-ecosystem/openshift-preflight/certification/internal/rpm
This package has a single function that is not tested. It may benefit from being relocated (up to the implementer).

komish:
Fixes #618
This PR adds coverage for the **certification/internal/rpm** package by adding fixtures covering as many cases as possible in both the sqlite3 rpmdb format and the berkeleydb rpmdb format.

There's an included readme with the fixtures indicating how they were created. The TL;DR there is that they were created as brand new rpmdbs, and then a single, minimal package was installed so that we could assert we were reading the package list correctly.

There is commented code for the moment, but it can be uncommented once I go and resolve what I believe to be a bug for the go-rpmdb package (sqlite3), or whenever I/we figure out how to modify the "schema" of a berkeleydb "table". Not very familiar with it yet.

Either way, this is 90%+ coverage for the package.


bcrochet (changes requested):
This seems a bit much. The library that we use already has tests like this. Is there no way to mock the library out to test our code and not the library?

certification/internal/rpm/rpm_test.go
```diff
@@ -0,0 +1,93 @@
...
+		basePath := path.Join(wd, "..", "..", "..", "test", "rpmdbs", "bdb")
+
+		It("should succeed", func() {
+			packages, err := GetPackageList(context.TODO(), basePath)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(len(packages)).To(Equal(1))
+		})
+	})
+
+	Context("With no rpmdb of any kind found at know good locations", func() {
```
bcrochet, line 40: "known"

test/rpmdbs/bdb-corrupt/var/lib/rpm/Packages
```diff
@@ -0,0 +1 @@
+CORRUPT
```
bcrochet, line 1: I don't think any of the other files are necessary besides this one.

komish: Spot on. I thought I tried it and was unsuccessful, but I tried it again and it's working so I've removed the extras.

certification/internal/rpm/rpm_test.go
komish: Fixed

acornett21 (approved):

certification/internal/rpm/rpm_test.go
```diff
@@ -0,0 +1,93 @@
...
+		// rpmdb.sqlite was replaced with just the string "CORRUPT"
+		basePath := path.Join(wd, "..", "..", "..", "test", "rpmdbs", "sqlite-corrupt")
+		It("should throw an error", func() {
+			packages, err := GetPackageList(context.TODO(), basePath)
+			Expect(err).To(HaveOccurred())
+			Expect(len(packages)).To(Equal(0)) // basically a nil check
+		})
+	})
+
+	/* Commented until we can identify how to properly fail a read in bdb
```
bcrochet, line 70: I would say instead of commenting the whole thing, we'd be better off putting in a skip at the beginning of the spec and have it be documented that way.

acornett21: I think the RPM changes have been merged, so I think this PR could reference the new RPM package and uncomment these changes. At least I think that was the intent.

bcrochet (changes requested):
Something about this "smells". I feel like we are testing the external lib, instead of relying on testing in that lib that's already been done. We should instead be looking at a way to abstract this to make our code more testable. I don't have a direct suggestion at the moment.

komish:
> Something about this "smells". I feel like we are testing the external lib, instead of relying on testing in that lib that's already been done. We should instead be looking at a way to abstract this to make our code more testable. I don't have a direct suggestion at the moment.

Okay. Given there are more pressing libraries that need coverage, I'll just table this for the moment until I free up to return to it.

bcrochet:
> > Something about this "smells". I feel like we are testing the external lib, instead of relying on testing in that lib that's already been done. We should instead be looking at a way to abstract this to make our code more testable. I don't have a direct suggestion at the moment.
> 
> Okay. Given there are more pressing libraries that need coverage, I'll just table this for the moment until I free up to return to it.

Sorry about the basically duplicate comment. I'm fine with putting this in, and looking at a refactor later. Coverage is coverage.

komish:
I'm actually gonna close this for now. We're not getting that much coverage here.

Issue #685, opened 2022-06-01
[Container cert] Require --docker-config option if --submit is specified; display which file is not found.


jfrancin:
## Is your feature request related to a problem? Please describe.
At present in Preflight 1.2.1, if a user runs a container cert with the `--submit` option but leaves out the `--docker-config` option for a container whose repo is readable without authentication, the scan runs, but then fails with an inscrutable error message:

```
./preflight check container quay.io/purestorage/openstack-cinder-containers:16.1 --submit --pyxis-api-token=<HIDDEN> --certification-project-id=<PID>
time="2022-05-31T11:37:11-04:00" level=info msg="certification library version 1.2.1 <commit: 16e00a287ab183eb06e08839d0b87c5c7ab7c548>"
[...] 
time="2022-05-31T11:39:45-04:00" level=info msg="preparing results that will be submitted to Red Hat"
Error: open : no such file or directory
time="2022-05-31T11:39:47-04:00" level=fatal msg="open : no such file or directory"
```
This error is from the missing docker-config file.

So there are two problems here: 
a) The `--submit` option requires a `--docker-config` file option, but it's not enforced in command parsing.

b) File open error gives no clue about which file is missing.

## Describe the solution you'd like.

a) Enforce the requirement: if `--submit` is given but `--docker-config` is not, fail the command line parse. If you really want to be pedantic, check for the existence of the docker-config file during the parse rather than waiting until it tries to do the submit.

b) Beef up file open error message to indicate which file is missing.

acornett21:
> a) The `--submit` option requires a `--docker-config` file option, but it's not enforced in command parsing.

This is used to be true that both were required, but now the user can give `PFLT_DOCKERCONFIG` which is why the `--docker-config` is not required with the `--submit` flag. Maybe we just need to add back that check but include an `|| os.Getenv("PFLT_DOCKERCONFIG").

jfrancin:
That would work very well. One way or the other, the option should be satisfied.

acornett21:
@bcrochet @komish  Two options:

- Add back in the logic to check for the `--docker-config` or `PFLT_DOCKERCONFIG` presence upfront and error out early
- Since `docker_config_json` is `nullable` in pyxis I'm thinking maybe we only add the dockerconfig to the project struct if `viper.GetString("dockerConfig")` is `""`. If we got to this point in the code, we were able to pull the image, which means it's a public repo.

[...]

PR #695, opened 2022-06-14
updating check container to only read the docker file from disk if present

acornett21:
- Fixes: #685 
- updating to only read the docker file from dish if the user passes in a value
- updating to not set pyxis docker config json if a user did not pass the value, for the case where a docker config is already on the project
- adding in a warning if a user does not provide a docker config file for a github repo 
   - I'm on the fence about if we want to do this or not, it almost seems that Clair should create tokens for our backend systems for dockerhub

### DIFF
--- a/cmd/check_container_test.go
+++ b/cmd/check_container_test.go
@@ -156,6 +156,20 @@ var _ = Describe("Check Container Command", func() {
 			})
 		})
 
+		Context("and no docker config command argument was provided", func() {
+			BeforeEach(func() {
+				fakePC.setSRFuncSubmitSuccessfully("", "")
+			})
+			It("should not throw an error", func() {
+				sbmt.dockerConfig = ""
+				err := os.Remove(dockerConfigPath)
+				Expect(err).ToNot(HaveOccurred())
+
+				err = sbmt.Submit(context.TODO())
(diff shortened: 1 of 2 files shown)


cmd/check_container.go
```diff
@@ -182,13 +183,39 @@ func checkContainer(ctx context.Context, cfg *runtime.Config) error {
...
 		if err != nil {
-			return err
+			return fmt.Errorf("image uri could not be parsed: %v", err)
 		}
 
-		certProject.Container.DockerConfigJSON = string(dockerConfigJsonBytes)
+		// outputting a warning if a user does not give us a DockerConfig and is using a Docker Hub registry
+		if cfg.DockerConfig == "" && reference.Context().RegistryStr() == name.DefaultRegistry {
+			log.Warn("DockerConfig not provided for Docker Hub image, this is strongly encouraged due to the " +
+				"rate limit imported for unauthenticated requests.")
```
bcrochet, line 217: I think "imported" should be "imposed"?

acornett21: Nice catch!

skattoju (approved):
lgtm

bcrochet:
Needs a rebase, and unit tests.

bcrochet (approved):
/lgtm

Issue #620, opened 2022-05-09
Test coverage improvement: github.com/redhat-openshift-ecosystem/openshift-preflight/cmd


komish:
This package has low test coverage at 10.2%. Given that this is our CMD package, we may need to consider refactoring how we establish our runnable functions (e.g. values for a cobra command's `RunE` or `Run` function) in order to improve coverage.

PR #698, opened 2022-06-15
Add runtime assets tests

komish:
Part of #620 

NOTE: This builds on #697 so I'll put do-not-merge on this until I can rebase. That said, it's ready for review. The only files that changed are **cmd/runtime_assets.go** and **cmd/runtime_assets_test.go**. 

This PR adds runtime-assets testing by:
- Extracting write operations to an io.Writer
- Extracting formatting to its own function
- Passing through the cmd output to the io.Writer

Doesn't quite reach 100% because of two error paths that I'm not hitting. Seemed like too much fragmenting of the code just to hit the two error paths, but we hit the same error path in at least one test.

I'm seeing 85.7% on this file now. Okay for the moment.

### DIFF
--- a/cmd/runtime_assets.go
+++ b/cmd/runtime_assets.go
@@ -1,8 +1,10 @@
 package cmd
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
+	"io"
 
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/runtime"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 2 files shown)


acornett21 (approved):

acornett21:
I just noticed this needs a `make fmt` but other then that, this should get GTG.

komish:
> I just noticed this needs a `make fmt` but other then that, this should get GTG.

Fixed. I don't know why I haven't added a pre-commit hook for this x_x already.

PR #702, opened 2022-06-21
Fallback to cli tag in container HasUniqueTag test

### LINKED ISSUE #701: Some private registries always fail HasUniqueTag container check
## Bug Description
This issue is a variant of #454 which relates to the has_unique_tag policy check and private registries.

As for the specific details, the issue stems from at least one partner's registry implementation, where the `/tags/list` API endpoint is not accessible, even with authorized client credentials. So this isn't so much a "bug" per se in preflight, but it does necessitate a workaround for partners who wish to certify but hide the tag-list endpoint.

```
preflight version 1.2.1 <commit: 16e00a287ab183eb06e08839d0b87c5c7ab7c548>
```

## Steps to Reproduce:
This isn't directly reproducible without access to a registry which has the `/tags/list` API endpoint for an image repository hidden/inaccessible.

The current symptom is failure of the HasUniqueTag test, despite the image/repository otherwise complying to the certification requirements:

- image must have at least one tag
- image may have more than one tag
- if image has only one tag, then that tag must not be `latest`

## Expected Result
For all accessible image registries (private or otherwise) that do not throw an error (eg: 404 not found) when attempting to access the `/tags/list` API endpoint, we should ignore the empty list that gets returned from crane and inspect the tag provided with the image pull spec instead (provided it's a tag, and not a digest).

## Actual Result
Crane returns an empty list of tags, which doesn't meet the policy requirements above.

## Additional Context
PR / fix pending.

jsm84:
Fixes #701 

Summary:
This introduces a change to the HasUniqueTag container check which
falls back to using the tag provided from the cli if:
* tags list comes back empty from crane
* crane returns no error from fetching tags list
* cli img tag ref is not a digest

Signed-off-by: Josh Manning <19478595+jsm84@users.noreply.github.com>

### DIFF
--- a/certification/internal/policy/container/has_unique_tag.go
+++ b/certification/internal/policy/container/has_unique_tag.go
@@ -24,10 +24,20 @@ type hasUniqueTagCheck struct {
 }
 
 func (p *hasUniqueTagCheck) Validate(ctx context.Context, imgRef certification.ImageReference) (bool, error) {
-	tags, err := p.getDataToValidate(ctx, fmt.Sprintf("%s/%s", imgRef.ImageRegistry, imgRef.ImageRepository))
+	imgRepo := fmt.Sprintf("%s/%s", imgRef.ImageRegistry, imgRef.ImageRepository)
+	tags, err := p.getDataToValidate(ctx, imgRepo)
 	if err != nil {
-		return false, fmt.Errorf("could not get image tags: %v", err)
+		return false, fmt.Errorf("Error: failed to get tags list for %s: %v", imgRepo, err)
 	}
+
+	if len(tags) < 1 {
+		if !strings.HasPrefix(imgRef.ImageTagOrSha, "sha256:") {
+			tags = append(tags, imgRef.ImageTagOrSha)
+		} else {
+			return false, fmt.Errorf("Error: no tags found for %s: cannot assert tag from digest", imgRepo)
+		}
+	}
+
 	return p.validate(tags)
 }
 
--- a/certification/internal/policy/container/has_unique_tag_test.go
+++ b/certification/internal/policy/container/has_unique_tag_test.go
@@ -1,10 +1,13 @@
 package container
 
 import (
+	"bytes"
 	"context"
+	"encoding/json"
 	"fmt"
 	"io"
 	"log"
+	"net/http"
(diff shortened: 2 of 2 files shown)


bcrochet:
Please add a unit test that covers this.

certification/internal/policy/container/has_unique_tag.go
```diff
@@ -24,10 +24,22 @@ type hasUniqueTagCheck struct {
 }
 
 func (p *hasUniqueTagCheck) Validate(ctx context.Context, imgRef certification.ImageReference) (bool, error) {
-	tags, err := p.getDataToValidate(ctx, fmt.Sprintf("%s/%s", imgRef.ImageRegistry, imgRef.ImageRepository))
+	imgRepo := fmt.Sprintf("%s/%s", imgRef.ImageRegistry, imgRef.ImageRepository)
+	tags, err := p.getDataToValidate(ctx, imgRepo)
 	if err != nil {
-		return false, fmt.Errorf("could not get image tags: %v", err)
+		err = fmt.Errorf("Error: failed to get tags list for %s: %w", imgRepo, err)
+		return false, err
```
bcrochet, line 31: Two things (and this goes for line 38 as well)

1. Don't bother reassigning err. Just use fmt.Errorf in the return statement directly.
2. Use %v instead of %w. We don't want to leak the actual error from here.

certification/internal/policy/container/has_unique_tag_test.go
```diff
@@ -61,6 +68,13 @@ var _ = Describe("UniqueTag", func() {
 				Expect(ok).To(BeFalse())
 			})
 		})
+		Context("When registry returns an empty tag list", func() {
+			It("should pass Validate", func() {
```
acornett21, line 72: To me, it would make sense for this test to have its own `BeforeEach` with the `mockRegistry` in it, then to add to the existing before each. But not sure what others think.

bcrochet: I think I would wrap the existing tests in a Context()/Describe(), and add the new tests to their own, with their own BeforeEach.

acornett21 (approved):

bcrochet (approved):
/lgtm

Issue #694, opened 2022-06-14
[Container cert] Preflight blows up on MongoDB RPMs with "invalid tag summary"


jfrancin:
## Bug Description
A container whose RPM data shows that both of these RPMs are installed:

https://repo.mongodb.org/yum/redhat/8Server/mongodb-org/5.0/x86_64/RPMS/mongodb-org-shell-5.0.3-1.el8.x86_64.rpm
https://repo.mongodb.org/yum/redhat/8Server/mongodb-org/5.0/x86_64/RPMS/mongodb-cli-1.20.3.x86_64.rpm
  
will cause Preflight to blow up: 

`time="2022-06-13T16:39:54-04:00" level=info msg="check completed: HasNoProhibitedPackages" ERROR="could not get rpm list: invalid package info: invalid tag summary" result="could not get rpm list: invalid package info: invalid tag summary"`

## Version and Command Invocation
Preflight version 1.2.1

Command: 
`preflight check container quay.io/redhat_emp1/jfrancin-quaytest-1:9.30 --docker-config /run/user/1000/containers/auth.json`

## Steps to Reproduce:
(How can we reproduce this?)
1) Create a simple container that has the two MongoDB packages mentioned above installed using Yum in the Dockerfile (see Additonal Context)
2) Push the container to a registry.
3) Run the preflight command above

## Expected Result
The "could not get rpm list" error would not occur

## Actual Result
The error in the Description occurred.

## Additional Context
Minimal Dockerfile to reproduce this:

---cut here----
#
# Test dockerfile to reproduce Preflight error
#

FROM registry.access.redhat.com/ubi8/ubi-init

LABEL name="abot/stacks"\
      summary="Stacks image for Abot application" \
      description="ABot-Stacks(Red Hat UBI)" \
      maintainer="Rebaca Technologies contactus@rebaca.com"\
      vendor="Rebaca Technologies Pvt. Ltd"\
      version="9.22"\
      io.openshift.tags="abot-stacks 9.22" \
      io.k8s.display-name="abot-stacks 9.22"

RUN yum -y install https://repo.mongodb.org/yum/redhat/8Server/mongodb-org/5.0/x86_64/RPMS/mongodb-org-shell-5.0.3-1.el8.x86_64.rpm && \
    yum -y install https://repo.mongodb.org/yum/redhat/8Server/mongodb-org/5.0/x86_64/RPMS/mongodb-cli-1.20.3.x86_64.rpm && \
[...]

PR #705, opened 2022-06-23
updating to latest version of go-rpmdb

acornett21:
- Fixes: #694 
- Update to latest version of `go-rpmdb` in opdev fork from branch `preflight-next`
- Includes a fix for a change in upstream that renamed a method that we were using 
- Includes changes to support `mongo-cli` and other `rpm`s that have a `RPMTAG_SUMMARY` header tag as a value of `6` (STRING) instead of the expect value of `9` (I18NSTRING) 

Testing:
I tested the same image before and after this change and got the same list of files before/after. And the new method and old method look the same to me in the `go-rpmdb` codebase. But I would be interested in better/other ways to make sure this doesn't introduce unintended consequences, thoughts?


bcrochet (approved):
/lgtm

komish (approved):
/lgtm

PR #706, opened 2022-06-23
updating gh actions to login to a given registry

acornett21:
## Motivation
There is sometimes a need to cut a hotfix for a partner and the easiest way to do this is to create a release from a fork, if we have changes in `main` that are not read to be released. In-order to accomplish this, the login steps need to be moved around/added in various steps to guard against failures.

## Testing
This was used in my fork in the `ghactions` [branch](https://github.com/acornett21/openshift-preflight/tree/ghactions) to cut release `1.2.1-rpm-hotfix` found [here](https://github.com/acornett21/openshift-preflight/releases/tag/1.2.1-rpm-hotfix)

### DIFF
--- a/.github/workflows/build-multiarch.yml
+++ b/.github/workflows/build-multiarch.yml
@@ -33,6 +33,14 @@ jobs:
     name: create and push a multiarch manifest to the repo
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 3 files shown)


bcrochet (approved):
/lgtm

komish (approved):
/lgtm

PR #707, opened 2022-06-23
Add cmd package interfaces with implementations

komish:
This PR adds type/interface definitions to help abstract out the cmd package's executions, with the end goal of improving testability. 

This also adds a few implementations that will be used.

This PR _may_ reduce test coverage, but this is temporary. The bulk of the testing consuming this code is done in check_container_test.go to keep things organized logically. It's written, but is not included here to avoid redundant setup code.


cmd/types.go
```diff
@@ -0,0 +1,176 @@
...
+type pyxisClient interface {
+	FindImagesByDigest(ctx context.Context, digests []string) ([]pyxis.CertImage, error)
+	GetProject(context.Context) (*pyxis.CertProject, error)
+	SubmitResults(context.Context, *pyxis.CertificationInput) (*pyxis.CertificationResults, error)
+}
+
+// newPyxisClient initializes a pyxisClient with relevant information from cfg.
+// If the the CertificationProjectID, PyxisAPIToken, or PyxisHost are empty, then nil is returned.
+// Callers should treat a nil pyxis client as an indicator that pyxis calls should not be made.
+func newPyxisClient(ctx context.Context, cfg certification.Config) pyxisClient {
```
bcrochet, line 40: Perhaps this should be 'newAuthenticatedPyxisClient'? There is a valid use case for getting the pyxis client without authentication.

cmd/types.go
```diff
@@ -0,0 +1,176 @@
...
+// containerCertificationSubmitter submits container results to Pyxis, and implements
+// a resultSubmitter.
+type containerCertificationSubmitter struct {
+	certificationProjectID string
+	pyxis                  pyxisClient
+	dockerConfig           string
+	preflightLogFile       string
+}
+
+func (s *containerCertificationSubmitter) Submit(ctx context.Context) error {
```
acornett21, line 62: There seems to be at least [this](https://github.com/redhat-openshift-ecosystem/openshift-preflight/blob/main/cmd/check_container.go#L252) missing in this function. I'm not sure if other things are missed as well.

acornett21 (changes requested):

cmd/types.go
komish: That's not part of the submit function, which is why it's not included here.

acornett21: Yeah I noticed that in the other PR. Hard to follow all of these, just want to make sure we don't miss something.

cmd/types.go
bcrochet: Since it is package local, should be fine. Do as you wish.

bcrochet (approved):
/lgtm

acornett21 (approved):

PR #703, opened 2022-06-23
Make CertificationInput a public type

komish:
In restructuring the `cmd` package to increase test coverage, and specifically writing out the Pyxis Submission abstraction, I found myself needing the CertificationInput to test various scenarios. Typical cases should use the builder, but testing shouldn't have to use the builder.

Related #620 

Signed-off-by: Jose R. Gonzalez <jose@flutes.dev>

PR #708, opened 2022-06-23
Add interface-driven preflightCheck function

komish:
This PR depends on #704, #703, #707. I'll rebase as those merge, if necessary.

Related #620 

This PR adds a new (unused in this PR) preflightCheck function that is testable, by being primarily interface driven. This reuses old interfaces (engine, formatters) and some new definitions that encapsulate our behaviors.

A future PR will reconfigure check_container.go to use this. In addition, I believe the way this is currently written means it will also be usable by check_operator.go, but that code isn't written.


komish:
This PR reduces coverage, but I believe that's due to the dependencies which reduce coverage in the same way.

komish:
Rebased after the merging of 704 and 703.

acornett21 (approved):

bcrochet (approved):
/lgtm

PR #710, opened 2022-06-24
Add test coverage for support cmd

komish:
This PR adds test coverage to the support package. This is an interact prompt that builds out text for users seeking support, including a link to Red Hat's support dahsboard with prepopulated data.

I was unable to successfully test the execution through the cobra command here, as I couldn't figure out how to properly wire up stdin from the cobra command context to the promptui stdin. I tried a bunch of permutations, including passing cmd's stdin value all the way through the promptUI's stdin values, and I still came up unsuccessful. I included functions for such tests in case someone wants to take a stab at it in the future.

Either way, this command was partitioned into logic and validations, and each of those (including prompts) are tested. 

I also included a test function to run these prompts with a timeout. This is intended to prevent tests from hanging in the event the inputs passed to prompts no longer align with what a completed prompt looks like.

### DIFF
--- a/cmd/root_test.go
+++ b/cmd/root_test.go
@@ -5,6 +5,8 @@ import (
 	"context"
 	"os"
 	"path/filepath"
+	"sync"
+	"time"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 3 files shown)


komish:
Data races aren't promising! I'll resolve next week

komish:
I can't see a way to remove the data race because it's in promptui. I'm gonna close this, rewrite support, and then submit it for PR.

PR #718, opened 2022-06-28
remove extraneous defined error

komish:
I had originally added this error because I was leveraging a function that would evaluate if this error was encountered, and return the default formatter. That function never made it to the code base, so we don't need this function (in keeping in line with our new approach to standalone error definitions).


bcrochet (approved):
/lgtm

acornett21 (approved):

PR #719, opened 2022-06-28
adding in PFLT_JUNIT to the configs md for clarity

acornett21:
- `PFLT_JUNIT` environment variable was never documented, so adding it to CONFIG.md, so we do not forget that it exists.


bcrochet (approved):
/lgtm

PR #720, opened 2022-06-28
Increase coverage of the engine package

bcrochet:
This gets the engine package to 100%.

Also added the context param to all functions, as these could be
called by other places of the code in the future, and the context
should come from the top-most levels.


acornett21 (approved):

komish (approved):
/lgtm

PR #721, opened 2022-06-28
fix bug in derived runtime config where values were not populated from parameter

komish:
In writing tests, I discovered a bug in the runtime.Config derivation where the configuration values were not being populated from the passed in parameter, but rather from the upstream viper instance. 

It probably would have looked the same at execution time because we still fully leverage the upstream viper early on in the process, but the tests called just the derivation functions and caught it. 

Should add 100% coverage for this source file.

### DIFF
--- a/certification/runtime/config_test.go
+++ b/certification/runtime/config_test.go
@@ -1,0 +1,76 @@
+package runtime
+
+import (
+	"reflect"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/spf13/viper"
+)
+
+var _ = Describe("Viper to Runtime Config", func() {
+	var baseViperCfg *viper.Viper
+	var expectedRuntimeCfg *Config
+	BeforeEach(func() {
+		baseViperCfg = viper.New()
+		expectedRuntimeCfg = &Config{}
+
+		baseViperCfg.Set("logfile", "logfile")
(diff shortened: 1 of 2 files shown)


certification/runtime/config_test.go
```diff
@@ -0,0 +1,65 @@
...
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/spf13/viper"
+)
+
+var _ = Describe("Viper to Runtime Config", func() {
+	var baseViperCfg *viper.Viper
+	var expectedRuntimeCfg *Config
+	BeforeEach(func() {
+		baseViperCfg = viper.New()
```
bcrochet, line 13: Doesn't have to be... But it feels like you could get away with a DescribeTable here? It's no longer an "extension" in ginkgo. It would get rid of the second context even. Feel free to ignore.

bcrochet (commented):
Suggestion, but not hard requirement.

certification/runtime/config_test.go
komish: I think the DescribeTable approach may require more edge case handling than it's wortyh because expectedRuntimeConfig is a struct without getters/setters to simplify it. With that said, I've added a struct key count check to make sure that Config changing causes a test failure, resulting in a change in tests with the included PR.

skattoju (approved):

PR #724, opened 2022-06-29
Refactor operatorsdk and add some unit tests

bcrochet:
This was completely untested before. Now, the new operatorsdk
package has 90+% coverage. For the most part, this is a somewhat
transparent change.

The change also removes any reference to 'OperatorSdkEngine'. It
should never have been named that.

### DIFF
--- a/certification/internal/bundle/bundle.go
+++ b/certification/internal/bundle/bundle.go
@@ -11,7 +11,7 @@ import (
 
 	"github.com/blang/semver"
 	operatorv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
-	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/internal/cli"
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/internal/operatorsdk"
 	log "github.com/sirupsen/logrus"
 	"sigs.k8s.io/yaml"
 )
@@ -21,9 +21,13 @@ const ocpVerV1beta1Unsupported = "4.9"
 // versionsKey is the OpenShift versions in annotations.yaml that lists the versions allowed for an operator
 const versionsKey = "com.redhat.openshift.versions"
 
-func Validate(ctx context.Context, engine cli.OperatorSdkEngine, imagePath string) (*cli.OperatorSdkBundleValidateReport, error) {
+type operatorSdk interface {
+	BundleValidate(context.Context, string, operatorsdk.OperatorSdkBundleValidateOptions) (*operatorsdk.OperatorSdkBundleValidateReport, error)
+}
+
+func Validate(ctx context.Context, operatorSdk operatorSdk, imagePath string) (*operatorsdk.OperatorSdkBundleValidateReport, error) {
 	selector := []string{"community", "operatorhub"}
-	opts := cli.OperatorSdkBundleValidateOptions{
+	opts := operatorsdk.OperatorSdkBundleValidateOptions{
 		Selector:        selector,
 		Verbose:         true,
 		ContainerEngine: "none",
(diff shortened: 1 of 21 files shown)


certification/internal/bundle/bundle.go
```diff
@@ -21,9 +21,13 @@ const ocpVerV1beta1Unsupported = "4.9"
 // versionsKey is the OpenShift versions in annotations.yaml that lists the versions allowed for an operator
 const versionsKey = "com.redhat.openshift.versions"
 
-func Validate(ctx context.Context, engine cli.OperatorSdkEngine, imagePath string) (*cli.OperatorSdkBundleValidateReport, error) {
+type operatorSdk interface {
+	BundleValidate(context.Context, string, operatorsdk.OperatorSdkBundleValidateOptions) (*operatorsdk.OperatorSdkBundleValidateReport, error)
+}
+
+func Validate(ctx context.Context, engine operatorSdk, imagePath string) (*operatorsdk.OperatorSdkBundleValidateReport, error) {
```
komish, line 28: Probably should rename the parameter `engine`

certification/internal/operatorsdk/operatorsdk.go
```diff
@@ -1,29 +1,32 @@
...
 
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/artifacts"
-	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/internal/cli"
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/runtime"
 	log "github.com/sirupsen/logrus"
 )
 
-func NewOperatorSdkEngine(userProvidedScorecardImage string) *cli.OperatorSdkEngine {
-	var engine cli.OperatorSdkEngine = operatorSdkEngine{scorecardImage: userProvidedScorecardImage}
+func NewOperatorSdk(userProvidedScorecardImage string, cmdContext execContext) *operatorSdk {
```
komish, line 17: SUPERNIT: Should this just be...
```suggestion
func New(userProvidedScorecardImage string, cmdContext execContext) *operatorSdk {
```

certification/internal/operatorsdk/operatorsdk_test.go
```diff
@@ -0,0 +1,167 @@
...
+				Selector:       []string{"selector1", "selector2"},
+				Kubeconfig:     "/my/kubeconfig",
+				Namespace:      "awesome-namespace",
+				ServiceAccount: "this-service-account",
+				Verbose:        true,
+				WaitTime:       "120m",
+			})
+			Expect(err).ToNot(HaveOccurred())
+		})
+	})
```
komish, line 44: What are we testing here? This reads to me like we're evaluating the results of our scorecard execution to see if the asset passed. I know we're mocking here, but do we need to get an *operatorSDK to test that scorecard results have passed?

certification/internal/operatorsdk/operatorsdk.go
bcrochet: Totally agree!

certification/internal/bundle/bundle.go
bcrochet: Done.

certification/internal/operatorsdk/operatorsdk_test.go
bcrochet: It's basically the happy path, with all options set. Should probably figure out how to check that the actual options made it into the exec.

certification/internal/operatorsdk/operatorsdk.go
```diff
@@ -1,29 +1,32 @@
...
 }
 
-type operatorSdkEngine struct {
+type operatorSdk struct {
 	scorecardImage string
+	cmdContext     execContext
 }
 
-func (o operatorSdkEngine) Scorecard(image string, opts cli.OperatorSdkScorecardOptions) (*cli.OperatorSdkScorecardReport, error) {
+type execContext = func(name string, arg ...string) *exec.Cmd
```
komish, line 27: I know it's not public, but let's add a comment here so we understand what's happening without having to find uses of this type.

bcrochet: Done.

komish (approved):
/lgtm

PR #725, opened 2022-06-30
Increase coverage in the openshift package

bcrochet:
Adds a couple of small cases for GetOpenshiftClusterVersion. Will need
refactoring to get further coverage.

Adds cases to the Openshift client tests, where the item already
exists.


komish (approved):
/lgtm

PR #726, opened 2022-06-30
Bump go-rpmdb and go-containerregistry deps

bcrochet:
* go-containerregistry => v0.10.0
* go-rpmdb => opdev/preflight-next


komish (approved):
/lgtm

PR #733, opened 2022-07-19
Add fallback paths to credential resolution when using docker.io as your registry 

### LINKED ISSUE #731: Authentication to docker.io appears to be broken
## Bug Description
Trying to pull private images from docker.io appears to be broken, even with valid credentials.

## Version and Command Invocation, steps to reproduce
(Reproduce) Create a valid docker credential file and pass it to preflight to pull a private image.

```
$ PFLT_LOGLEVEL=trace ./preflight check container docker.io/xkomish/sleeper-test:latest --docker-config ./auth.json 
time="2022-07-18T09:55:06-05:00" level=debug msg="config file not found, proceeding without it"
time="2022-07-18T09:55:06-05:00" level=info msg="certification library version 0.0.0 <commit: 4ae2b40b01d3376d443d7a351c77f08a0fdf4cf0>"
time="2022-07-18T09:55:06-05:00" level=debug msg="target image: docker.io/xkomish/sleeper-test:latest"
time="2022-07-18T09:55:06-05:00" level=debug msg="pulling image from target registry"
time="2022-07-18T09:55:06-05:00" level=trace msg="entering preflight keychain Resolve"
Error: failed to pull remote container: GET https://index.docker.io/v2/xkomish/sleeper-test/manifests/latest: UNAUTHORIZED: authentication required; [map[Action:pull Class: Name:xkomish/sleeper-test Type:repository]]
time="2022-07-18T09:55:07-05:00" level=fatal msg="failed to pull remote container: GET https://index.docker.io/v2/xkomish/sleeper-test/manifests/latest: UNAUTHORIZED: authentication required; [map[Action:pull Class: Name:xkomish/sleeper-test Type:repository]]"
```

The same creds with podman appear to work

```
$ podman pull --authfile ./auth.json docker.io/xkomish/sleeper-test:latest
Trying to pull docker.io/xkomish/sleeper-test:latest...
Getting image source signatures
Copying blob sha256:bdbbaa22dec6b7fe23106d2c1b1f43d9598cd8fc33706cc27c1d938ecd5bffc7
Copying blob sha256:bdbbaa22dec6b7fe23106d2c1b1f43d9598cd8fc33706cc27c1d938ecd5bffc7
Copying config sha256:4d24b97186c7a265c78aebe5d356cb108aefad9648eb36608d11530b39ac64a3
Writing manifest to image destination
Storing signatures
4d24b97186c7a265c78aebe5d356cb108aefad9648eb36608d11530b39ac64a3
```

## Expected Result
Preflight should be able to authenticate to docker.io using the provided credentials.

## Actual Result
It's unauthorized.

komish:
Fixes #731 

This patch resolves an issue in credential resolution when attempting to work with images that reside in Docker hub. I'm not certain, but I believe this came to be after bumping go-containerregistry's version which is why this would have worked previously but not now. I added a test to cover the case with the new resolution logic.

# The problem

Internally, Crane (the library used by Preflight) rewrites image targets that have the `docker.io` hostname to `index.docker.io`. The Preflight Keychain would query a credential file (e.g. `auth.json`) for a normalized value, `https://index.docker.io/v1/`. 

If a user logs into Docker hub using `podman`, it seems whatever value the user passed in as the registry is what's written to the configuration file. Contrast that with `docker` cli or `crane`, which write the normalized value. So for example:

```
<container-tool> login docker.io --username example --password password
```

In my testing:

Podman (4.0.z) would produce...

```
{
	"auths": {
		"docker.io": {
			"auth": "encodedvalues"
		}
	}
}
```

DockerCLI (tested against fedora 36 I believe, but I didn't write it down) and Crane would produce:

```
{
	"auths": {
		"https://index.docker.io/v1/": {
			"auth": "encodedvalues"
		}
	}
}
```

The former would not be read by preflight, but the latter would.

# The solution 

This patch resolves this issue by adding `docker.io` as an additional target when searching for the right credentials.

So for an image `docker.io/example/test`, the following in your credential file _should_ work **from Preflight's perspective**.

- https://index.docker.io/v1/
- docker.io/example/test
- docker.io

We still recommend the credential file is built using podman or related tools for the moment, as there are upstream services (on submission) that I believe do not expect the first option, and should accept any of the others.

### DIFF
--- a/certification/internal/authn/keychain.go
+++ b/certification/internal/authn/keychain.go
@@ -3,6 +3,7 @@ package authn
 import (
 	"fmt"
 	"os"
+	"strings"
 
 	"github.com/docker/cli/cli/config"
 	"github.com/docker/cli/cli/config/types"
@@ -69,11 +70,25 @@ func (k *preflightKeychain) Resolve(target craneauthn.Resource) (craneauthn.Auth
 		return nil, fmt.Errorf("could not load authfile from reader: %v", err)
 	}
 
-	var cfg, empty types.AuthConfig
-	for _, key := range []string{
+	// We'll check the authconfig for creds associated with these endpoints.
+	authFileTargets := []string{
 		target.String(),
 		target.RegistryStr(),
-	} {
+	}
+
+	// If the user logged into docker.io using podman, the auth.json would
+	// contain docker.io. Crane rewrites this to index.docker.io internally,
+	// but the credential file does not have an entry for this, so we also
+	// check for docker.io/* entries that match.
+	if strings.Contains(name.DefaultRegistry, target.RegistryStr()) {
+		authFileTargets = append(authFileTargets,
+			strings.Replace(target.String(), name.DefaultRegistry, "docker.io", 1),
+			strings.Replace(target.RegistryStr(), name.DefaultRegistry, "docker.io", 1),
+		)
+	}
+
+	var cfg, empty types.AuthConfig
+	for _, key := range authFileTargets {
 		if key == name.DefaultRegistry {
 			key = craneauthn.DefaultAuthKey
 		}
--- a/certification/internal/authn/keychain_test.go
+++ b/certification/internal/authn/keychain_test.go
@@ -123,6 +123,14 @@ func TestVariousPaths(t *testing.T) {
 			Password: "bar",
 		},
 	}, {
+		desc:    "valid config file as written by podman; default registry",
+		target:  defaultRegistry,
+		content: fmt.Sprintf(`{"auths": {"docker.io": {"auth": %q}}}`, encode("foo", "bar")),
+		cfg: &craneauthn.AuthConfig{
+			Username: "foo",
+			Password: "bar",
+		},
+	}, {
 		desc:   "valid config file; matches registry w/ v1",
 		target: testRegistry,
 		content: fmt.Sprintf(`{


acornett21:
> I'm not certain, but I believe this came to be after bumping go-containerregistry's version which is why this would have worked previously but not now. I added a test to cover the case with the new resolution logic.

FWIW, this has always been broken when we implemented our own auth logic. I checked out each version/commit to test this going back to v1.2.0 I believe, and applying go-containerregistry's `latest` ontop of each version.

acornett21 (approved):

bcrochet:
> So for an image `docker.io/example/test`, the following in your credential file _should_ work **from Preflight's perspective**.
> 
>     * https://index.docker.io/v1/
> 
>     * docker.io/example/test
> 
>     * docker.io
> 

Shouldn't the more explicit docker.io/example/test come first? If someone is bothering to put an explicit 'example/test' then that means they have specific credentials for that repo/image.

komish:
> Shouldn't the more explicit docker.io/example/test come first? If someone is bothering to put an explicit 'example/test' then that means they have specific credentials for that repo/image.

In theory, yep. What I left out is that technically the full order is:

- `index.docker.io/example/test` (This is what the user provided with docker.io replaced with index.docker.io)
- `https://index.docker.io/v1/` (This is the base domain of what the user provided, which we're rewriting)
- `docker.io/example/test` (this is our fallback)
- `docker.io` (this is our fallback, but just the base domain)

I left out the first item because, while its existence is checked in auth.json, the only way we hit that is if a user handcrafts their auth.json, or they `podman login index.docker.io/example/test`.

Docker, at least on Fedora 36[1], trims the path. 

```
[vagrant@fedora36 ~]$ docker login index.docker.io/namespace/image --username xkomish
Password: 
Login Succeeded
[vagrant@fedora36 ~]$ cat ~/.docker/config.json 
{
	"auths": {
		"index.docker.io": {
			"auth": "encodedvalue"
		}
	}
}
```

Note that it apparently doesn't write `https://index.docker.io/v1/` ... for whatever reason.

To your point, we're effectively specifying precedence, I guess? So we could change the order to be:

- MostSpecific
- FallbackMostSpecific
- OriginalRegistryOnly
- FallbackRegistryOnly

WDYT

[1]
```
[vagrant@fedora36 ~]$ docker -v
Docker version 20.10.17, build aa7e414
```

bcrochet:
> To your point, we're effectively specifying precedence, I guess? So we could change the order to be:
> 
>     * MostSpecific
> 
>     * FallbackMostSpecific
> 
>     * OriginalRegistryOnly
> 
>     * FallbackRegistryOnly
> 
> 
> WDYT

I'm fine with it as written. Just wanted to make sure we were thinking about it.

/lgtm

PR #744, opened 2022-07-26
Add related images check

### LINKED ISSUE #740: Add a Check to Verify that the RelatedImages Field Contains all Images for the Operator

We need to add a check that will scan the CSV for an operator and output any images that are referenced in the CSV but not listed in the `spec.relatedImages` field. This check will simply output a warning with the list of images not referenced in the `relatedImages` field for the OpenShift 4.11 release. With OpenShift 4.12, the check should fail if any images are not listed in the `relatedImages` field.

Preflight should output the list of images as a warning that are not found in the `relatedImages` field in the CSV.

N/A

## Additional context.

N/A

bcrochet:
This check verifies that all images referred to in the CSV are actually
in the relatedImages section of the CSV.

Currently, it just logs a warning, but still passes. It will only fail
if there is an error.

Fixes #740


acornett21 (approved):

skattoju (approved):

certification/internal/policy/operator/related_images.go
```diff
@@ -0,0 +1,89 @@
...
+		}
+	}
+
+	return imageNames, relatedImages, nil
+}
+
+func (p *RelatedImagesCheck) validate(ctx context.Context, images []string, relatedImages map[string]struct{}) (bool, error) {
+	for _, image := range images {
+		if _, ok := relatedImages[image]; !ok {
+			log.Warningf("Image %s is not in relatedImages. This will eventually cause this check to fail", image)
```
komish, line 65: So since this is temporarily always passing and currently gets written to results, should we consider modifying the Help() or Metadata() text in a way that informs the user that this check is auto passing? So we don't get users who passed one day, and then fail when we flip the bit?

komish (approved):
Just a non-blocking comment, up to you on how to proceed with it. LGTM and Approved.

PR #747, opened 2022-07-27
Only add results that are not marked as "optional"

bcrochet:
The "optional" level will treat the check as normal, but will not
add it's outcome to the actual results. A user will still see the
log message with the status, but it will not cause a failure, nor
will it be reported in the output.

### DIFF
--- a/certification/internal/engine/engine.go
+++ b/certification/internal/engine/engine.go
@@ -174,6 +174,9 @@ func (c *CraneEngine) ExecuteChecks(ctx context.Context) error {
 		c.results.TestedImage = c.Image
 
 		log.Debug("running check: ", check.Name())
+		if check.Metadata().Level == "optional" {
+			log.Infof("Check %s is not currently being enforced.", check.Name())
+		}
 
 		// run the validation
 		checkStartTime := time.Now()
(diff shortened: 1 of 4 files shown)


certification/internal/engine/engine.go
```diff
@@ -224,6 +227,13 @@ func (c *CraneEngine) ExecuteChecks(ctx context.Context) error {
 	return nil
 }
 
+func appendUnlessOptional(results []runtime.Result, result runtime.Result) []runtime.Result {
+	if result.Check.Metadata().Level == "optional" {
+		return results
+	}
+	return append(results, result)
+}
+
```
komish, line 236: Needs a test

komish (changes requested):
LGTM, just needs a test for the new func

certification/internal/engine/engine.go
bcrochet: Done.

komish (approved):
/lgtm

certification/internal/engine/engine_test.go
```diff
@@ -81,6 +99,8 @@ var _ = Describe("Execute Checks tests", func() {
 				goodCheck,
 				errorCheck,
 				failedCheck,
+				optionalCheckPassing,
+				optionalCheckFailing,
```
skattoju, line 103: Is it worth it to add an 'InformationalCheck' result for things that will always pass but we want the user to be informed ex: the SCC check where we can show a help text if a custom scc is in use ?

bcrochet: Let's discuss on that issue.

certification/internal/engine/engine.go
```diff
@@ -224,6 +227,13 @@ func (c *CraneEngine) ExecuteChecks(ctx context.Context) error {
 	return nil
 }
 
+func appendUnlessOptional(results []runtime.Result, result runtime.Result) []runtime.Result {
+	if result.Check.Metadata().Level == "optional" {
```
komish, line 231: I know I already gave you the LGTM, but:
1) Should this be an exported constant, usable by checks?
2) Should we add dev docs?

acornett21: Now that we have more then one check implementing `optional`, I agree on both points above

bcrochet: Neither are bad ideas. (I thought we already had dev docs? Or, since this would be an exported constant, it will end up in API docs)

acornett21 (approved):

certification/internal/engine/engine.go
komish: That works for me.

PR #748, opened 2022-07-28
add check that logs a message if a scc is specified in the csv

### LINKED ISSUE #738: Add a Check to Output the Security Context Constraint (SCC) in Use

Add a check to determine and output the Security Context Constraint the operator has requested. If nothing is specified in the CSV, preflight should indicate that the default will be used.

Display a message to the user indicating the identified Security Context Constraint.

N/A

## Additional context.

N/A

skattoju:
fixes #738

### DIFF
--- a/certification/internal/bundle/bundle.go
+++ b/certification/internal/bundle/bundle.go
@@ -9,6 +9,8 @@ import (
 	"path/filepath"
 	"strings"
 
+	rbacv1 "k8s.io/api/rbac/v1"
+
 	"github.com/blang/semver"
 	operatorv1alpha1 "github.com/operator-framework/api/pkg/operators/v1alpha1"
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/internal/operatorsdk"
@@ -210,4 +212,47 @@ func GetSupportedInstallModes(ctx context.Context, csvReader io.Reader) (map[str
 		}
 	}
 	return installedModes, nil
+}
+
+// GetSecurityContextConstraints returns an string array of SCC resource names requested by the operator as specified
+// in the csv
+func GetSecurityContextConstraints(ctx context.Context, csvReader io.Reader) ([]string, error) {
+	var csv operatorv1alpha1.ClusterServiceVersion
+	bts, err := io.ReadAll(csvReader)
+	if err != nil {
+		return nil, fmt.Errorf("could not get CSV from reader: %v", err)
+	}
+	err = yaml.Unmarshal(bts, &csv)
+	if err != nil {
+		return nil, fmt.Errorf("malformed CSV detected: %v", err)
+	}
+	for _, cp := range csv.Spec.InstallStrategy.StrategySpec.ClusterPermissions {
+		for _, rule := range cp.Rules {
+			if hasSCCApiGroup(rule) && hasSCCResource(rule) {
+				return rule.ResourceNames, nil
+			}
+		}
+	}
+	return nil, nil
+}
+
+// hasSCCApiGroup returns a bool indicating if security.openshift.io is in the list of apigroups referenced in a policy
+// rule
+func hasSCCApiGroup(rule rbacv1.PolicyRule) bool {
+	for _, apiGroup := range rule.APIGroups {
+		if apiGroup == "security.openshift.io" {
+			return true
+		}
+	}
+	return false
+}
+
+// hasSCCResource returns a bool indicating if any securitycontextconstraints resources are referenced in a policy rule
+func hasSCCResource(rule rbacv1.PolicyRule) bool {
+	for _, resource := range rule.Resources {
+		if resource == "securitycontextconstraints" {
+			return true
+		}
+	}
+	return false
 }
--- a/certification/internal/policy/operator/scc_info.go
+++ b/certification/internal/policy/operator/scc_info.go
@@ -1,0 +1,102 @@
+package operator
+
+import (
+	"context"
+	"fmt"
+	"os"
+
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/internal/bundle"
+	log "github.com/sirupsen/logrus"
+
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification"
+)
+
+// securityContextConstraintsInCSV evaluates the csv and logs a message if a non default security context constraint is
+// needed by the operator
+type securityContextConstraintsInCSV struct {
+	customSCCSpecified bool
+}
+
+func NewSecurityContextConstraintsCheck() *securityContextConstraintsInCSV {
+	return &securityContextConstraintsInCSV{
+		false,
+	}
+}
+
+func (p *securityContextConstraintsInCSV) Validate(ctx context.Context, bundleRef certification.ImageReference) (bool, error) {
+	requestedSCCList, err := p.dataToValidate(ctx, bundleRef.ImageFSPath)
+	if err != nil {
+		return false, err
+	}
+
+	return p.validate(ctx, requestedSCCList)
+}
+
+func (p *securityContextConstraintsInCSV) dataToValidate(ctx context.Context, imagePath string) ([]string, error) {
+	csvFilepath, err := bundle.GetCsvFilePathFromBundle(imagePath)
+	if err != nil {
+		return nil, err
+	}
+
+	csvFileReader, err := os.Open(csvFilepath)
+	if err != nil {
+		return nil, err
+	}
+
+	requestedSccList, err := bundle.GetSecurityContextConstraints(ctx, csvFileReader)
+	if err != nil {
+		return nil, fmt.Errorf("unable to extract security context constraints from ClusterServiceVersion: %w", err)
+	}
+
+	return requestedSccList, nil
+}
+
+func (p *securityContextConstraintsInCSV) validate(ctx context.Context, requestedSccList []string) (bool, error) {
+	if len(requestedSccList) == 0 {
+		log.Infof("No custom security context constraint was detected in the CSV. The default restricted SCC will be used.")
+		return true, nil
+	}
+
+	if len(requestedSccList) != 1 {
+		return false, fmt.Errorf("only one scc should be requested at a time")
+	}
+
+	log.Infof("A custom scc was specified: %s , action may be needed by a cluster admin, please check the "+
+		"operator documentation for more information", requestedSccList[0])
+	// set bool to change help text
+	p.customSCCSpecified = true
+	return true, nil
+}
+
+func (p *securityContextConstraintsInCSV) Name() string {
+	return "SecurityContextConstraintsInCSV"
+}
+
+func (p *securityContextConstraintsInCSV) Metadata() certification.Metadata {
+	return certification.Metadata{
+		Description:      "Evaluates the csv and logs a message if a non default security context constraint is needed by the operator",
+		Level:            "optional",
+		KnowledgeBaseURL: "https://redhat-connect.gitbook.io/certified-operator-guide/troubleshooting-and-resources/sccs", // Placeholder
+		CheckURL:         "https://redhat-connect.gitbook.io/certified-operator-guide/troubleshooting-and-resources/sccs", // Placeholder
+	}
+}
+
+func (p *securityContextConstraintsInCSV) Help() certification.HelpText {
+	// TODO implement a way to display help text for informational checks
+	// Currently this bool triggered customized help is not shown anywhere since this check is purely informational
+	// and always passes. Since the ask is to indicate this information to the user it should eventually be shown in the
+	// output rather than just logs
+	if p.customSCCSpecified {
+		return certification.HelpText{
+			Message: "A custom security context constraint was detected in the csv. " +
+				"Please see the operators documentation for details.",
+			Suggestion: "A custom security context constraint may need to be applied by a cluster admin.",
+		}
+	}
+	// This is the default help text returned in case the check fails
+	return certification.HelpText{
+		Message: "The securityContextConstraintsInCSV logs a message if the operator requests a " +
+			"security context constraint. Please review the operators documentation to see if this is needed.",
+		Suggestion: "If no scc is detected the default restricted scc will be used.",
+	}
+}
--- a/certification/engine/engine.go
+++ b/certification/engine/engine.go
@@ -62,6 +62,7 @@ func initializeChecks(ctx context.Context, p policy.Policy, cfg certification.Co
 				"",
 				&http.Client{Timeout: 60 * time.Second}),
 			),
+			operatorpol.NewSecurityContextConstraintsCheck(),
 		}, nil
 	case policy.PolicyContainer:
 		return []certification.Check{
--- a/certification/internal/policy/operator/scc_info_test.go
+++ b/certification/internal/policy/operator/scc_info_test.go
@@ -1,0 +1,106 @@
+package operator
+
+import (
+	"context"
+	"os"
+	"path/filepath"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification"
+)
+
+var _ = Describe("securityContextConstraintsInCSV", func() {
+	const (
+		manifestsDir                  = "manifests"
+		clusterServiceVersionFilename = "myoperator.clusterserviceversion.yaml"
+		customSCC                     = "\n      clusterPermissions:\n      - rules:\n        - apiGroups:\n          - security.openshift.io\n          resources:\n          - securitycontextconstraints\n          resourceNames:\n          - custom\n          verbs:\n          - use\n        serviceAccountName: my-operand\n"
+		multipleSCC                   = "\n      clusterPermissions:\n      - rules:\n        - apiGroups:\n          - security.openshift.io\n          resources:\n          - securitycontextconstraints\n          resourceNames:\n          - anyuid\n          - nonroot\n          verbs:\n          - use\n        serviceAccountName: my-operand\n"
+		noSCC                         = "\n"
+	)
+	var (
+		securityContextConstraintsCheck securityContextConstraintsInCSV
+		imageRef                        certification.ImageReference
+		csvContents                     = buildCsvContent(noSCC)
+	)
+	AssertMetaData(&securityContextConstraintsCheck)
+	BeforeEach(func() {
+		securityContextConstraintsCheck = *NewSecurityContextConstraintsCheck()
+		tmpDir, err := os.MkdirTemp("", "scc-check-bundle-*")
+		Expect(err).ToNot(HaveOccurred())
+		imageRef.ImageFSPath = tmpDir
+		DeferCleanup(os.RemoveAll, tmpDir)
+		err = os.Mkdir(filepath.Join(tmpDir, manifestsDir), 0o755)
+		Expect(err).ToNot(HaveOccurred())
+		err = os.WriteFile(filepath.Join(tmpDir, manifestsDir, clusterServiceVersionFilename), []byte(csvContents), 0o644)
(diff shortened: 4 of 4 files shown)


certification/internal/bundle/bundle.go
```diff
@@ -211,3 +213,70 @@ func GetSupportedInstallModes(ctx context.Context, csvReader io.Reader) (map[str
 	}
 	return installedModes, nil
 }
+
+func ExtractImagesFromBundle(ctx context.Context, csvReader io.Reader) ([]string, error) {
```
acornett21, line 217: Wasn't this method removed previously?

skattoju: hmm weird it didnt go away with rebase will remove..

certification/internal/policy/operator/scc_check.go
```diff
@@ -0,0 +1,121 @@
...
+// SecurityContextConstraintsCheck evaluates the csv and logs a message if a non default security context constraint is
+// needed by the operator
+
+type SecurityContextConstraintsCheck struct {
+	customSCCSpecified  bool
+	builtInSCCSpecified bool
+}
+
+func getDefaultSCCs() []string {
+	return []string{"restricted", "privileged", "nonroot", "hostnetwork", "hostmount-anyuid", "hostaccess", "anyuid"}
```
acornett21, line 23: Is there some other place we can get these from? Say the OpenShift API? As new SCCs come to be, I don't want us to have to remember these values.

skattoju: hmm we could do `oc get scc` but not sure if there is a way to filter built-in ones 🤔

acornett21: I wouldn't want to shell out to get the values, but I'd hope they are structs, constants in the API somewhere

certification/internal/policy/operator/scc_check.go
```diff
@@ -0,0 +1,121 @@
...
+
+func (p *SecurityContextConstraintsCheck) Name() string {
+	return "SecurityContextConstraintsCheck"
+}
+
+func (p *SecurityContextConstraintsCheck) Metadata() certification.Metadata {
+	return certification.Metadata{
+		Description:      "Evaluates the csv and logs a message if a non default security context constraint is needed by the operator",
+		Level:            "optional",
+		KnowledgeBaseURL: "https://redhat-connect.gitbook.io/certified-operator-guide/troubleshooting-and-resources/sccs", // Placeholder
```
acornett21, line 94: I don't think we want to link to gitbook anymore, only the cert docs, however I don't think that has anything on SCC's yet.

acornett21 (changes requested):

certification/internal/policy/operator/scc_check.go
skattoju: and too your point that list was _already_ out of date 😑 😅

certification/internal/bundle/bundle.go
acornett21: I might be wrong, But I thought it was removed.

skattoju: you are right its not in main 👍

certification/internal/policy/operator/scc_check.go
skattoju: apparently not though : https://github.com/openshift/apiserver-library-go/blob/aa2bfcc66a43fc37769e3112642d7e089cce7913/pkg/securitycontextconstraints/sccadmission/admission.go#L140

certification/internal/policy/operator/scc_check.go
acornett21: Usually it would be [this](https://access.redhat.com/documentation/en-us/red_hat_software_certification/8.45/html/red_hat_openshift_software_certification_policy_guide/assembly-products-managed-by-an-operator_openshift-sw-cert-policy-container-images#doc-wrapper), but that doesn't talk about SCC. Asking to see if this is getting updated. This can be a non-blocker till we find out.

certification/internal/bundle/bundle.go
```diff
@@ -211,3 +213,41 @@ func GetSupportedInstallModes(ctx context.Context, csvReader io.Reader) (map[str
 	}
 	return installedModes, nil
 }
+
+func GetSecurityContextConstraints(ctx context.Context, csvReader io.Reader) ([]string, error) {
```
komish, line 217: Needs Doc

certification/internal/bundle/bundle.go
```diff
@@ -211,3 +213,41 @@ func GetSupportedInstallModes(ctx context.Context, csvReader io.Reader) (map[str
...
+		for _, rule := range cp.Rules {
+			if hasSCCApiGroup(rule) && hasSCCResource(rule) {
+				return rule.ResourceNames, nil
+			}
+		}
+	}
+	return nil, nil
+}
+
+func hasSCCApiGroup(rule v1.PolicyRule) bool {
```
komish, line 237: Needs Doc

certification/internal/bundle/bundle.go
```diff
@@ -211,3 +213,41 @@ func GetSupportedInstallModes(ctx context.Context, csvReader io.Reader) (map[str
...
+	for _, apiGroup := range rule.APIGroups {
+		if apiGroup == "security.openshift.io" {
+			return true
+		}
+	}
+	return false
+}
+
+func hasSCCResource(rule v1.PolicyRule) bool {
```
komish, line 246: Needs Doc

certification/internal/policy/operator/scc_check.go
```diff
@@ -0,0 +1,121 @@
...
+
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/internal/bundle"
+	log "github.com/sirupsen/logrus"
+
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification"
+)
+
+// SecurityContextConstraintsCheck evaluates the csv and logs a message if a non default security context constraint is
+// needed by the operator
+
```
komish, line 16: Extra space I think?

certification/internal/policy/operator/scc_check.go
```diff
@@ -0,0 +1,121 @@
...
+type SecurityContextConstraintsCheck struct {
+	customSCCSpecified  bool
+	builtInSCCSpecified bool
+}
```
komish, line 20: If the expected way to access this is via the Initializer `New...` func, then this can probably be unexported.

certification/internal/policy/operator/scc_check.go
```diff
@@ -0,0 +1,121 @@
...
+		return nil, fmt.Errorf("unable to extract security context constraints from ClusterServiceVersion: %w", err)
+	}
+
+	return requestedSccList, nil
+}
+
+func (p *SecurityContextConstraintsCheck) validate(ctx context.Context, requestedSccList []string) (bool, error) {
+	defaultSccList := getDefaultSCCs()
+
+	if requestedSccList == nil || len(requestedSccList) == 0 {
```
komish, line 64:
```suggestion
	if len(requestedSccList) == 0 {
```

Just check the length. It should be 0 if `requestedSccList` is nil. I'd rather we returned `[]Type{}` from the generating function, but either way, this should be sufficient.

https://go.dev/play/p/_WAM2hd1YDL

certification/internal/policy/operator/scc_check.go
```diff
@@ -0,0 +1,121 @@
...
+		log.Infof("No security context constraint was requested so using the default %s scc", defaultSccList[0])
```
komish, line 65: I would argue that we don't really need to log this, right? We're just concerned if the developer plans on using _custom_ SCCs?

certification/internal/policy/operator/scc_check.go
```diff
@@ -0,0 +1,121 @@
...
+
+	for _, defaultScc := range defaultSccList {
+		if requestedSccList[0] == defaultScc {
+			log.Infof("A built in default security context constraint %s was requested", defaultScc)
+			p.builtInSCCSpecified = true
+			return true, nil
+		}
+	}
+
+	log.Infof("A custom scc was specified: %s", requestedSccList[0])
```
komish, line 81: This is a log-only check, right? So if we are not planning on including it in the results.json, we need the log line to contain all of the useful information for the user.

I see you setting the p.customSCCSpecified bool and using that to drive Metadata/Help, but I don't think the user will ever see those if I'm understanding how this check is supposed to play out.

komish (changes requested):
Couple of changes.

Also, I'm unsure about it as I've not given it much though, but should we be parsing other files in the manifests directory that include RBAC for the potential inclusion of SCCs there? I vaguely remember we discussed it, but I can't remember where we landed.

certification/internal/policy/operator/scc_check.go
acornett21: +1

certification/internal/policy/operator/scc_check.go
acornett21: I created the below issue in their repo
- Relates: openshift/apiserver-library-go#88

certification/internal/policy/operator/scc_check.go
skattoju: part of the ask 🤷 "If nothing is specified in the CSV, preflight should indicate that the default will be used."

certification/internal/policy/operator/scc_check.go
skattoju: Yes currently these are not shown anywhere.. but was hoping we could have a level called "informational" or something along those lines that would show the appropriate help text at some point ? don't think logs are sufficient to indicate things to the end user.. I imagine end users should be indicated to with actual output ?

certification/internal/policy/operator/scc_check.go
komish: Gotcha. I'm not sure what it adds, but if it's a part of the ask I suppose it can stay. That said, can we reword?

```suggestion
		log.Infof("No custom security context constraint was detected in this bundle so using the default %s scc", defaultSccList[0])
```

Or something to that extent?

certification/internal/policy/operator/scc_check.go
komish: > but was hoping we could have a level called "informational" or something along those lines that would show the appropriate help text at some point

+1 That's fine, but until that happens, we need to log all of the context. I'd also leave a `// TODO` comment in with a note indicating what you mentioned to be resolved at a later date.

certification/internal/bundle/bundle.go
skattoju: will add

certification/internal/bundle/bundle.go
skattoju: will add

certification/internal/bundle/bundle.go
skattoju: will add

certification/internal/policy/operator/scc_check.go
skattoju: will fix

certification/internal/policy/operator/scc_check.go
skattoju: will fix

certification/internal/policy/operator/scc_check.go
skattoju: will fix

certification/internal/policy/operator/scc_check.go
skattoju: will say in the csv because thats what we check.. I have been assuming that if an scc is specified it should be in the csv so far ? I think thats the case with RBACs in general ?

certification/internal/policy/operator/scc_check.go
```diff
@@ -0,0 +1,120 @@
+package operator
```
bcrochet, line 1: nit, but none of the other files in this package have '_check' in them. I'd rather that not become a thing.

certification/internal/bundle/bundle.go
```diff
@@ -9,6 +9,8 @@ import (
 	"path/filepath"
 	"strings"
 
+	v1 "k8s.io/api/rbac/v1"
```
bcrochet, line 12: If you're going to have an alias, something more descriptive like 'rbacv1' would be appropriate here.

certification/internal/policy/operator/scc_check.go
skattoju: how about scc_info.go and scc_info_test.go ?

certification/internal/policy/operator/scc_check.go
skattoju: added and updated log messaged 👍

certification/internal/bundle/bundle.go
skattoju: fixed 👍

certification/internal/policy/operator/scc_check.go
komish: I _think_ you can include arbitrary RBAC in the manifests... I can never find the link to operator-sdk where they list the valid Kinds, though. All that said, We're only checking the CSV, so that's all fine. Just indicating custom vs. default?

acornett21: Since OLM doesn't apply custom SCC's and a user would need to, I think us just checking the CSV is fine, since the developer of the operator would have to mention their SCC in the CSV for the operator to know about it anyways. Us scanning the entire `/bundle` is overkill IMO.

skattoju: also added that if custom check operator docs as it may need to be defined..

certification/internal/policy/operator/scc_check.go
bcrochet: That's fine for now. Just don't want to introduce _check.

certification/internal/policy/operator/scc_info.go
```diff
@@ -0,0 +1,124 @@
...
+		if requestedSccList[0] == defaultScc {
+			log.Infof("A built in default security context constraint %s was requested", defaultScc)
+			p.builtInSCCSpecified = true
+			return true, nil
+		}
+	}
+
+	log.Infof("A custom scc was specified: %s , action may be needed by a cluster admin, please check the "+
+		"operator documentation for more information", requestedSccList[0])
+	p.customSCCSpecified = true
```
komish, line 82: One last thing, add a comment here that just says we've set this so that Metadata values change appropriately. This one is just to help future me 😆

certification/internal/policy/operator/scc_info.go
```diff
@@ -0,0 +1,124 @@
...
+	return true, nil
+}
+
+func (p *securityContextConstraintsCheck) Name() string {
+	return "securityContextConstraintsCheck"
```
komish, line 87: 1) Start with a capital letter
2) Remove "Check" at the end
3) Maybe needs to be more descriptive? `SecurityContextConstraintsInCSV`

komish (changes requested):
I'm mostly good to go. Just a few more things already mentioned ^, and I'll review again as other reviewer comments are resolved 👍 Thanks @skattoju!

bcrochet:
The commit message itself needs more than just the summary.

certification/internal/policy/operator/scc_info.go
```diff
@@ -0,0 +1,124 @@
...
+)
+
+// securityContextConstraintsCheck evaluates the csv and logs a message if a non default security context constraint is
+// needed by the operator
+type securityContextConstraintsCheck struct {
+	customSCCSpecified  bool
+	builtInSCCSpecified bool
+}
+
+func getDefaultSCCs() []string {
```
bcrochet, line 21: Nit: Can probably lose the 'get' here.

certification/internal/policy/operator/scc_info.go
```diff
@@ -0,0 +1,124 @@
...
+	if len(requestedSccList) == 0 {
+		log.Infof("No custom security context constraint was detected in the csv the default the restricted scc will be used")
+		return true, nil
+	}
+
+	if len(requestedSccList) != 1 {
+		return false, fmt.Errorf("only one scc should be requested at a time")
+	}
+
+	for _, defaultScc := range defaultSccList {
```
bcrochet, line 72: You could avoid this loop if you just made the default SCCs a map instead of a list. map[string]struct{} would do the trick. Then you could just check if the requested SCC is in the map as a key. i probably wouldn't even bother with the 'getDefaultsSccs' function, and just have the map unexported as a var.

bcrochet (changes requested):

certification/internal/policy/operator/scc_info.go
bcrochet: Or remove this entirely, based on my further comment.

acornett21:
Based on our prelfight meeting, this check can be simplified...we do not need to match to any of the known SCC's. We just need to see if they mention one in the csv, and if not log that the `default` will be used.

certification/internal/policy/operator/scc_info.go
skattoju: didnt we change it to small letter because its only accessed via NewSecurityContextConstraintsInCSV ?

certification/internal/policy/operator/scc_info.go
skattoju: removed 👍

certification/internal/policy/operator/scc_info.go
skattoju: removed entirely since we only need to print out the scc and we dont care if its custom or default..

certification/internal/policy/operator/scc_info.go
```diff
@@ -0,0 +1,102 @@
...
+
+	log.Infof("A custom scc was specified: %s , action may be needed by a cluster admin, please check the "+
+		"operator documentation for more information", requestedSccList[0])
+	// set bool to change help text
+	p.customSCCSpecified = true
+	return true, nil
+}
+
+func (p *securityContextConstraintsInCSV) Name() string {
+	return "securityContextConstraintsInCSV"
```
bcrochet, line 72: The string itself should be a Capital.

certification/internal/policy/operator/scc_info.go
```diff
@@ -0,0 +1,102 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("unable to extract security context constraints from ClusterServiceVersion: %w", err)
+	}
+
+	return requestedSccList, nil
+}
+
+func (p *securityContextConstraintsInCSV) validate(ctx context.Context, requestedSccList []string) (bool, error) {
+	if len(requestedSccList) == 0 {
+		log.Infof("No custom security context constraint was detected in the csv the default the restricted scc will be used")
```
bcrochet, line 56:
```suggestion
		log.Infof("No custom security context constraint was detected in the csv. The default restricted scc will be used")
```

bcrochet (changes requested):
2 more changes, then I think I'm good.

certification/internal/policy/operator/scc_info.go
skattoju: ah thats what jose meant 😅 fixed 👍

acornett21 (approved):

certification/internal/policy/operator/scc_info.go
```diff
@@ -0,0 +1,102 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("unable to extract security context constraints from ClusterServiceVersion: %w", err)
+	}
+
+	return requestedSccList, nil
+}
+
+func (p *securityContextConstraintsInCSV) validate(ctx context.Context, requestedSccList []string) (bool, error) {
+	if len(requestedSccList) == 0 {
+		log.Infof("No custom security context constraint was detected in the CSV. The default the restricted SCC will be used.")
```
bcrochet, line 56: There is still an extra 'the' in the second sentence.

skattoju: oops sorry typo blindness

bcrochet (approved):
/lgtm

PR #749, opened 2022-07-29
Do not write artifacts to current dir in tests

bcrochet:
A lot of the tests would create artifacts folders in the directory
under test. They would also create preflight.log in those same
locations. This makes sure that that does not occur.

It also removes the side effect from the artifacts.Path() call.
That call would create the artifacts dir, and that's just very
unexpected. Instead, this moves the dir creation side effect into
the artifacts.WriteFile() call, which is much more an expected
behaviour.


cmd/check_container_test.go
```diff
@@ -21,27 +22,20 @@ import (
...
 		Context("With all of the required parameters", func() {
+			BeforeEach(func() {
+				tmpDir, err := os.MkdirTemp("", "cmd-execute-*")
+				Expect(err).ToNot(HaveOccurred())
+				artifacts.SetDir(filepath.Join(tmpDir, "artifacts"))
+				DeferCleanup(os.RemoveAll, tmpDir)
+				os.Setenv("PFLT_ARTIFACTS", artifacts.Path())
+				os.Setenv("PFLT_LOGFILE", filepath.Join(tmpDir, "preflight.log"))
+				DeferCleanup(os.Unsetenv, "PFLT_ARTIFACTS")
+				DeferCleanup(os.Unsetenv, "PFLT_LOGFILE")
```
komish, line 37: 1) Could we have used flags here?
2) Can we organize the Defers together?

cmd/check_operator_test.go
```diff
@@ -15,6 +15,7 @@ import (
 
 var _ = Describe("Check Operator", func() {
 	Context("when running the check operator subcommand", func() {
+		BeforeEach(artifactsAndLogTemp)
```
komish, line 18: SUPER NIT: Can this function be renamed to declare exactly what it's doing?

cmd/suite_test.go
```diff
@@ -1,13 +1,27 @@
...
+
+var artifactsAndLogTemp = func() {
+	tmpDir, err := os.MkdirTemp("", "cmd-execute-*")
+	Expect(err).ToNot(HaveOccurred())
+	artifacts.SetDir(filepath.Join(tmpDir, "artifacts"))
+	DeferCleanup(os.RemoveAll, tmpDir)
+	os.Setenv("PFLT_ARTIFACTS", artifacts.Path())
+	os.Setenv("PFLT_LOGFILE", filepath.Join(tmpDir, "preflight.log"))
+	DeferCleanup(os.Unsetenv, "PFLT_ARTIFACTS")
+	DeferCleanup(os.Unsetenv, "PFLT_LOGFILE")
```
komish, line 26: Same as previous instance of similar code, reorganize defers near each other and do we need to use envs?

cmd/preflight_check.go
```diff
@@ -31,7 +30,11 @@ func preflightCheck(
 
 	// create the results file early to catch cases where we are not
 	// able to write to the filesystem before we attempt to execute checks.
-	resultsFile, err := rw.OpenFile(filepath.Join(artifacts.Path(), resultsFilenameWithExtension(formatter.FileExtension())))
+	resultsFilePath, err := artifacts.WriteFile(resultsFilenameWithExtension(formatter.FileExtension()), "")
```
komish, line 33: Aha, nice. I came looking straight for this when I saw this PR. Nice work.

komish (approved):
Approving because most of my comments are nitty.

/lgtm

cmd/check_container_test.go
bcrochet: Possibly... Let me look into that. If not, I'll put the defers together.

cmd/check_operator_test.go
bcrochet: Sure. It was a quick name. It's not the most descriptive.

cmd/check_container_test.go
bcrochet: On the one hand, yes, flags could be used. But then that makes L40 ugly... I think the env route is the correct one. I'll clean up the defers.

acornett21 (approved):

PR #750, opened 2022-08-02
Add recipe for check container with the config file

### LINKED ISSUE #692: [Security fix] Add pyxis-api-token-path for those who don't want to disclose the token during the Preflight run
During the `preflight check container` run with the `--submit` option, the `pyxis-api-token` is displayed in the console. That exposes the token in the history search, and in the logs of different CI tools.

It would be nice to deal with the Pyxis token the same way as we deal with the docker config: instead of providing the token itself, provide a path to the file that contains the token.  

To keep backward compatibility, that could be implemented as an additional option `--pyxis-api-token-path`.

## Additional context.
If you're still accepting the contributors, I would be happy to work on the implementation.

tkrishtop:
Adding a recipe to fix #692

### DIFF
--- a/docs/RECIPES.md
+++ b/docs/RECIPES.md
@@ -178,6 +178,29 @@ preflight check container registry.example.org/your-namespace/your-image:sometag
 --docker-config=/path/to/your/dockerconfig 
 ```
 
+### Testing Container and Passing Parameters in the Config File
+To avoid displaying the Pyxis token in the console, you may pass it in the config file. First, add config.yaml in the directory with the Preflight binary
+
+```bash
+$ cat config.yaml
+dockerConfig: path/to/config.json
+loglevel: trace
+logfile: artifacts/preflight.log
+artifacts: artifacts
+junit: true
+certification_project_id: my_nice_project_id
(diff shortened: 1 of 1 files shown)


acornett21 (approved):

skattoju (approved):

komish:
Weird, why do we _lose_ coverage here?

tkrishtop:
> Weird, why do we lose coverage here?

@komish it seems to be a bug. Here is the [coverage report for the PR](https://coveralls.io/builds/51347932), it complains that two lines were added into cmd/root.go file. Since I didn't change that file, does it look like some automated formatting during the code coverage? A new line somewhere?

tkrishtop:
Hello team, the proposed config is tested, [here is the invocation and the results](https://www.distributed-ci.io/jobs/b8d1a446-cdc7-42cb-8d43-12850c559d25/jobStates#c32b1199-c4c8-42f1-8f93-eece4ba5fc15:file111). It seems like I don't have permissions to merge a PR, so please feel free to merge when it seems to be convenient for you.

PR #751, opened 2022-08-02
Refactor cmd/ to have less coupling

bcrochet:
No more inits are being used in the individual command files.
The one that remains is to make sure that initConfig is called 
when cobra is initialized.

This should increase coverage slightly, and make the commands less
interdependent.

### DIFF
--- a/cmd/check.go
+++ b/cmd/check.go
@@ -13,13 +13,13 @@ import (
 	"github.com/spf13/viper"
 )
 
-var checkCmd = &cobra.Command{
-	Use:   "check",
-	Short: "Run checks for an operator or container",
-	Long:  "This command will allow you to execute the Red Hat Certification tests for an operator or a container.",
-}
+func checkCmd() *cobra.Command {
+	checkCmd := &cobra.Command{
+		Use:   "check",
+		Short: "Run checks for an operator or container",
+		Long:  "This command will allow you to execute the Red Hat Certification tests for an operator or a container.",
+	}
 
-func init() {
 	checkCmd.PersistentFlags().StringP("docker-config", "d", "", "Path to docker config.json file. This value is optional for publicly accessible images.\n"+
 		"However, it is strongly encouraged for public Docker Hub images,\n"+
 		"due to the rate limit imposed for unauthenticated requests. (env: PFLT_DOCKERCONFIG)")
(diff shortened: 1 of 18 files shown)


cmd/support_container_test.go
```diff
@@ -0,0 +1,33 @@
+// go:test !race
```
acornett21, line 1: Why does the race detection need to be ignored?

cmd/support_operator_test.go
```diff
@@ -0,0 +1,41 @@
+// go:test !race
```
acornett21, line 1: Same question as above.

cmd/support_container_test.go
bcrochet: This was added by @komish to the support.go. I don't know why though.

cmd/support_operator_test.go
bcrochet: Same answer as above. :)

cmd/support_container_test.go
acornett21: @komish Is this some leftover from `promptui`? Or is it still needed?

komish: Yep, this can be removed. PromptUI leftover.

cmd/support_operator_test.go
komish: Same answer as above!

cmd/check.go
```diff
@@ -13,13 +13,13 @@ import (
 	"github.com/spf13/viper"
 )
 
-var checkCmd = &cobra.Command{
-	Use:   "check",
-	Short: "Run checks for an operator or container",
-	Long:  "This command will allow you to execute the Red Hat Certification tests for an operator or a container.",
-}
+func check() *cobra.Command {
```
komish, line 16: Can we use the naming `<action>Cmd` for all functions that return cobra.Command? I won't go through and mark them all.

komish (changes requested):
Just one requested change, otherwise looks good. 

ALSO I totally see at a support command refactor here that could have been a separate PR :P

We already discussed offline, but I generally don't like our use of functions to return cobra commands here. I know this pattern can be seen in various other projects, but _typically_ I see it in cases where the function's parameters might change the resulting cobra command. I'm not particularly a fan of _that_ approach either, but I understand what drives it. 

All that said: if it works, it works!

cmd/check.go
komish: Can avoid using `new<action>Cmd` as I commonly see if you want, or use that instead if you prefer

bcrochet: Used the 'Cmd' suffix. Good suggestion.

acornett21 (approved):

PR #755, opened 2022-08-05
Fix bugs in new checks

bcrochet:
This PR contains two fixes.

1. Adding the RelatedImages check to the operator policy.
2. Fix the logic in the Certified Images check, so that if an image is not found in Pyxis, it's marked as uncertified.


acornett21 (approved):

jomkz (approved):

komish (approved):
/lgtm

PR #756, opened 2022-08-08
Reject submit if api token or cert id are empty

### LINKED ISSUE #754: Panic in stable version
## Bug Description
When running preflight stable image the panic can be seen during a submit

https://github.com/jrsmroz/gateway-operator/runs/7689619774?check_suite_focus=true
```shell
$ docker run -v /var/run/docker.sock:/var/run/docker.sock -v /home/runner/.docker/:/docker quay.io/opdev/preflight:stable check container scan.connect.redhat.com/***/gateway-operator:0.0.0-test3-redhat --docker-config=/docker/config.json --submit --certification-project-id=*** --pyxis-api-token=
Unable to find image 'quay.io/opdev/preflight:stable' locally
stable: Pulling from opdev/preflight
9849471674a8: Pulling fs layer
21ca3853a314: Pulling fs layer
d3dd878ef08b: Pulling fs layer
21ca3853a314: Download complete
9849471674a8: Verifying Checksum
9849471674a8: Download complete
d3dd878ef08b: Verifying Checksum
d3dd878ef08b: Download complete
9849471674a8: Pull complete
21ca3853a314: Pull complete
d3dd878ef08b: Pull complete
Digest: sha256:6670809cbe8de9495325cbae12d565388b772c6773d6537fad74e7202c2b2403
Status: Downloaded newer image for quay.io/opdev/preflight:stable
time="2022-08-05T10:44:59Z" level=info msg="certification library version 1.3.3 <commit: 6c30f7caae731ecc2faeaebd78c6d3981c708fe6>"
time="2022-08-05T10:45:09Z" level=info msg="check completed: HasLicense" result=PASSED
time="2022-08-05T10:45:09Z" level=info msg="check completed: HasUniqueTag" result=PASSED
time="2022-08-05T10:45:09Z" level=info msg="check completed: LayerCountAcceptable" result=PASSED
time="2022-08-05T10:45:09Z" level=info msg="check completed: HasNoProhibitedPackages" result=PASSED
time="2022-08-05T10:45:09Z" level=info msg="check completed: HasRequiredLabel" result=PASSED
time="2022-08-05T10:45:09Z" level=info msg="USER 1000 specified that is non-root"
time="2022-08-05T10:45:09Z" level=info msg="check completed: RunAsNonRoot" result=PASSED
time="2022-08-05T10:45:14Z" level=info msg="check completed: HasModifiedFiles" result=PASSED
time="2022-08-05T10:45:14Z" level=info msg="check completed: BasedOnUbi" result=PASSED
{
    "image": "scan.connect.redhat.com/***/gateway-operator:0.0.0-test3-redhat",
    "passed": true,
    "test_library": {
        "version": "1.3.3",
        "commit": "6c30f7caae731ecc2faeaebd78c6d3981c708fe6"
    },
    "results": {
        "passed": [
            {
                "name": "HasLicense",
                "elapsed_time": 0,
                "description": "Checking if terms and conditions applicable to the software including open source licensing information are present. The license must be at /licenses"
            },
            {
                "name": "HasUniqueTag",
                "elapsed_time": 275,
                "description": "Checking if container has a tag other than 'latest', so that the image can be uniquely identified."
            },
            {
                "name": "LayerCountAcceptable",
                "elapsed_time": 0,
                "description": "Checking if container has less than 40 layers.  Too many layers within the container images can degrade container performance."
            },
            {
                "name": "HasNoProhibitedPackages",
                "elapsed_time": 83,
                "description": "Checks to ensure that the image in use does not include prohibited packages, such as Red Hat Enterprise Linux (RHEL) kernel packages."
            },
            {
                "name": "HasRequiredLabel",
                "elapsed_time": 0,
                "description": "Checking if the required labels (name, vendor, version, release, summary, description) are present in the container metadata."
            },
            {
                "name": "RunAsNonRoot",
                "elapsed_time": 0,
                "description": "Checking if container runs as the root user because a container that does not specify a non-root user will fail the automatic certification, and will be subject to a manual review before the container can be approved for publication"
            },
            {
                "name": "HasModifiedFiles",
                "elapsed_time": 4549,
                "description": "Checks that no files installed via RPM in the base Red Hat layer have been modified"
            },
            {
                "name": "BasedOnUbi",
                "elapsed_time": 567,
                "description": "Checking if the container's base image is based upon the Red Hat Universal Base Image (UBI)"
            }
        ],
        "failed": [],
        "errors": []
    }
}
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x2c pc=0x533e7d]

goroutine 1 [running]:
github.com/sirupsen/logrus.(*Logger).Log(0x19046fb, 0x2a, {0xc000c097b8, 0x4ee66e, 0xc0001e7520})
	/go/pkg/mod/github.com/sirupsen/logrus@v1.8.1/logger.go:196 +0x1d
github.com/sirupsen/logrus.(*Logger).Info(...)
	/go/pkg/mod/github.com/sirupsen/logrus@v1.8.1/logger.go:220
github.com/redhat-openshift-ecosystem/openshift-preflight/cmd.(*noopSubmitter).Submit(0x1b0d0e0, {0xc00054c0c0, 0xc000c098e8})
	/go/src/preflight/cmd/types.go:187 +0x105
github.com/redhat-openshift-ecosystem/openshift-preflight/cmd.preflightCheck({0x1b37078, 0xc000044030}, 0xc000515440, {0x0, 0x0}, {0x1b1fed8, 0xc0000ae140}, {0x1b25010, 0xc0001aede0}, {0x1b25070, ...}, ...)
	/go/src/preflight/cmd/preflight_check.go:63 +0x448
github.com/redhat-openshift-ecosystem/openshift-preflight/cmd.checkContainerRunE(0x2774700, {0xc00010eaa0, 0x1, 0x5})
	/go/src/preflight/cmd/check_container.go:99 +0x32a
github.com/spf13/cobra.(*Command).execute(0x2774700, {0xc00010ea50, 0x5, 0x5})
	/go/pkg/mod/github.com/spf13/cobra@v1.4.0/command.go:856 +0x60e
github.com/spf13/cobra.(*Command).ExecuteC(0x2774e80)
	/go/pkg/mod/github.com/spf13/cobra@v1.4.0/command.go:974 +0x3bc
github.com/spf13/cobra.(*Command).Execute(...)
	/go/pkg/mod/github.com/spf13/cobra@v1.4.0/command.go:902
github.com/spf13/cobra.(*Command).ExecuteContext(...)
	/go/pkg/mod/github.com/spf13/cobra@v1.4.0/command.go:[89](https://github.com/jrsmroz/gateway-operator/runs/7689619774?check_suite_focus=true#step:17:91)5
github.com/redhat-openshift-ecosystem/openshift-preflight/cmd.Execute()
	/go/src/preflight/cmd/root.go:38 +0x58
main.main()
	/go/src/preflight/main.go:8 +0x17
```

Extracted from the logs:
`time="2022-08-05T10:44:59Z" level=info msg="certification library version 1.3.3 <commit: 6c30f7caae731ecc2faeaebd78c6d3981c708fe6>"`

## Steps to Reproduce:
(How can we reproduce this?)
1) We've run the preflight against our operator in the github CI pipeline. Seems to be random, as most of the time the panic does not occur

## Expected Result
I would expect the preflight not to panic.

## Actual Result
Preflight have panicked

## Additional Context
https://github.com/jrsmroz/gateway-operator/runs/7689619774?check_suite_focus=true

bcrochet:
Just setting the flags to required on submit doesn't test for whether
the flags are actually provided. It only checks for their existence.
This code checks that the data is actually provided, and that another
flag isn't accidentally accepted as the value.

Fixes #754


acornett21 (approved):

komish (approved):
/lgtm

skattoju (approved):

cmd/check_container.go
```diff
@@ -31,8 +36,9 @@ func checkContainerCmd() *cobra.Command {
 	checkContainerCmd.Flags().BoolVarP(&submit, "submit", "s", false, "submit check container results to red hat")
 	viper.BindPFlag("submit", checkContainerCmd.Flags().Lookup("submit"))
 
-	checkContainerCmd.Flags().String("pyxis-api-token", "", "API token for Pyxis authentication (env: PFLT_PYXIS_API_TOKEN)")
+	checkContainerCmd.Flags().StringVar(&pyxisApiToken, "pyxis-api-token", "", "API token for Pyxis authentication (env: PFLT_PYXIS_API_TOKEN)")
 	viper.BindPFlag("pyxis_api_token", checkContainerCmd.Flags().Lookup("pyxis-api-token"))
+	// checkContainerCmd.Flags().Lookup("pyxis-api-token").NoOptDefVal = "NOTSET"
```
acornett21, line 41: I think this line and line 53 might need to be removed or uncommented? Not sure which.

bcrochet: Removed. Thanks for catching that. Added a test for going against a config file.

cmd/check_container.go
```diff
@@ -168,17 +170,40 @@ func getContainerPolicyExceptions(ctx context.Context, pc pyxisClient) (policy.P
...
-		return fmt.Errorf("a container image positional argument is required")
+		return fmt.Errorf("a container image positional argument is required or a flag is not specified as --flag=value")
 	}
 
+	cmd.Flags().VisitAll(func(f *pflag.Flag) {
+		if f.Changed && f.Value.String() == "--submit" {
+			// We have --submit in one of the flags. That's a problem.
+			// We will set the submit flag to true so that the next block functions properly
+			submit = true
+		}
```
komish, line 181: I don't think this matches a case where a user did:

```
--pyxis-api-token --submit=true`
```

And I think that `--submit` is valid otherwise.

cmd/check_container.go
```diff
@@ -168,17 +170,40 @@ func getContainerPolicyExceptions(ctx context.Context, pc pyxisClient) (policy.P
...
+	})
+
+	// --submit was specified
 	if submit {
```
komish, line 185: I know we use the global for other reasons, but for this case, does it make sense to call the value directly from flags?

It probably doesn't make a difference, but I'm wondering if it would be more clear here?

komish: Also I know you didn't change this

komish (commented):
We already discussed this offline, but for the record (happy to be in the minority of opinion, of course, and this is not blocking): I just don't feel like this adds a huge amount of value to us. 

We effectively special-case a single flag's behavior in a bit of a unique/opaque way, in an effort to ensure that the submit flag wasn't "eaten" by another. This feels a bit weird to do.

It's a small change, and so for that reason, I'm not going to block on that opinion that may not be shared. Few comments, otherwise looks okay.

cmd/check_container.go
bcrochet: If they did --submit=true, that should be caught in the regular flow. But I'll add a test for that just in case.

cmd/check_container.go
bcrochet: I could localise submit with `submit := submit` and not affect the package global. But I'm actually thinking that I do want to change the package global.

cmd/check_container.go
bcrochet: You were right. I've added a test to cover this case. And updated the code to reflect that.

cmd/check_container.go
```diff
@@ -168,17 +170,40 @@ func getContainerPolicyExceptions(ctx context.Context, pc pyxisClient) (policy.P
 
 func checkContainerPositionalArgs(cmd *cobra.Command, args []string) error {
 	if len(args) != 1 {
-		return fmt.Errorf("a container image positional argument is required")
+		return fmt.Errorf("a container image positional argument is required or a flag is not specified as --flag=value")
```
komish, line 173: This is a strange user-facing message to send. Why would a flag not being specified as `--flag=value` cause us to throw an error if `len(args) != 1`?

cmd/check_container.go
```diff
@@ -168,17 +170,40 @@ func getContainerPolicyExceptions(ctx context.Context, pc pyxisClient) (policy.P
...
 	}
 
+	cmd.Flags().VisitAll(func(f *pflag.Flag) {
+		if f.Changed && strings.HasPrefix(f.Value.String(), "--submit") {
```
komish, line 177: Any reason not to use `strings.Contains`?

komish (commented):
Just a few final questions. Otherwise, this looks ok.

cmd/check_container.go
bcrochet: I think I changed that when I tried on of the other settings. This should be removed.

cmd/check_container.go
bcrochet: I suppose that would work... Any reason not to use strings.HasPrefix? It seems the scenario of --submit showing up anywhere other than the beginning seems slim to none.

bcrochet:
> Just a few final questions. Otherwise, this looks ok.

Addressed.

komish (approved):
/lgtm

Issue #757, opened 2022-08-09
HasUniqueTag Does Not Fall Back to Tag User Provided


acornett21:
## Bug Description
When a partner is using a registry where `/tags/list` is not avilable `preflight` should use the tag that came in on the command line. This does not seem to be happening.

## Version and Command Invocation
`v1.3.4`

## Steps to Reproduce:
1) `preflight check container public.ecr.aws/sumologic/sumologic-otel-collector:0.54.0-sumo-0-ubi` (or any other image in ECR)

## Expected Result
preflight should try to call `tags/list` and if there is a failure use the tag that was passed in on the command line.

## Actual Result
preflight returned the error, and marked the check as a failure.

## Additional Context
Logs from partner below:
```
time="2022-08-08T09:41:28Z" level=info msg="check completed: HasUniqueTag" err="Error: failed to get tags list for public.ecr.aws/sumologic/sumologic-otel-collector: GET https://public.ecr.aws/v2/sumologic/sumologic-otel-collector/tags/list?n=1000: NOT_FOUND: 404 page not found" result=ERROR
```

acornett21:
Removed blocker, since the parter is using RHCC registry so they were able to push to that registry to certify.

PR #759, opened 2022-08-10
update HasUniqueTag logic to fall back to the value the user provided

acornett21:
Updating the HashUniqueTag logic to follow the below rules:

1. if sha is given on cli, you MUST have a reachable /tags/list
2. If tag given on cli is "latest", you MUST have a reachable /tags/list.
3. If tag given on cli is not "latest", and is not a sha, use value user passed in.

The order above is so that the failover logic can be preserved for partners where we call the partners registry, and `/tags/list` is available but they also return `tags` as an empty list.

- Fixes: #757 

### DIFF
--- a/certification/internal/policy/container/has_unique_tag.go
+++ b/certification/internal/policy/container/has_unique_tag.go
@@ -25,17 +25,26 @@ type hasUniqueTagCheck struct {
 
 func (p *hasUniqueTagCheck) Validate(ctx context.Context, imgRef certification.ImageReference) (bool, error) {
 	imgRepo := fmt.Sprintf("%s/%s", imgRef.ImageRegistry, imgRef.ImageRepository)
-	tags, err := p.getDataToValidate(ctx, imgRepo)
-	if err != nil {
-		return false, fmt.Errorf("Error: failed to get tags list for %s: %v", imgRepo, err)
+
+	tags := make([]string, 0)
+	var err error
+	// if sha or latest tag is passed in `/tags/list` must be exposed and available to validate that the image is being tagged properly
+	if strings.HasPrefix(imgRef.ImageTagOrSha, "sha256:") || imgRef.ImageTagOrSha == "latest" {
+		tags, err = p.getDataToValidate(ctx, imgRepo)
+		if err != nil {
+			return false, fmt.Errorf("failed to get tags list for %s: %v", imgRepo, err)
+		}
 	}
 
-	if len(tags) < 1 {
-		if !strings.HasPrefix(imgRef.ImageTagOrSha, "sha256:") {
-			tags = append(tags, imgRef.ImageTagOrSha)
-		} else {
-			return false, fmt.Errorf("Error: no tags found for %s: cannot assert tag from digest", imgRepo)
+	// if tags is of length zero we know that either
+	// the partners registry returned an empty list so fall back
+	// or the value imgRef.ImageTagOrSha did not meet the previous conditions so falling back to use the value passed in
+	if len(tags) == 0 {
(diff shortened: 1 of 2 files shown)


acornett21:
There might be a better way to throw errors in the tests, so if there is please let me know! Thanks!

certification/internal/policy/container/has_unique_tag_test.go
```diff
@@ -52,18 +52,57 @@ var _ = Describe("UniqueTag", func() {
...
 				Expect(err).ToNot(HaveOccurred())
 				Expect(ok).To(BeFalse())
 			})
 		})
+		Context("When it has only latest tag and the registry throws an error", func() {
+			BeforeEach(func() {
+				s := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+					w.WriteHeader(http.StatusInternalServerError)
+				}))
+				DeferCleanup(func() {
```
bcrochet, line 92:
```suggestion
				DeferCleanup(s.Close)
```

certification/internal/policy/container/has_unique_tag_test.go
```diff
@@ -52,18 +52,57 @@ var _ = Describe("UniqueTag", func() {
...
 				Expect(err).ToNot(HaveOccurred())
 				Expect(ok).To(BeTrue())
 			})
 		})
+		Context("When it has tags other than latest and registry throws an error", func() {
+			BeforeEach(func() {
+				s := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+					w.WriteHeader(http.StatusInternalServerError)
+				}))
+				DeferCleanup(func() {
```
bcrochet, line 65:
```suggestion
				DeferCleanup(s.Close)
```

certification/internal/policy/container/has_unique_tag.go
```diff
@@ -25,17 +25,34 @@ type hasUniqueTagCheck struct {
...
-	if err != nil {
-		return false, fmt.Errorf("Error: failed to get tags list for %s: %v", imgRepo, err)
+
+	tags := make([]string, 0)
+	var err error
+	// if sha is passed in `/tags/list` must be exposed and available to validate that the image is being tagged properly
+	if strings.HasPrefix(imgRef.ImageTagOrSha, "sha256:") {
+		tags, err = p.getDataToValidate(ctx, imgRepo)
+		if err != nil {
+			return false, fmt.Errorf("failed to get tags list for %s when sha256 provided: %v", imgRepo, err)
```
bcrochet, line 35: It looks like the only difference between these is just the log message. I don't think that necessarily adds any value. This can be reduced to a single `if this || that`.

bcrochet (changes requested):
Couple of changes, and please fix the DeferCleanup on L109 to be consistent.

certification/internal/policy/container/has_unique_tag.go
komish: Along this line, to improve readability, I'd say that these two cases should be methods or functions. Then it literally becomes `if this || that`.

komish:
Barring what Brad already said (I'm in agreement), this LGTM.

/lgtm

komish (approved):

certification/internal/policy/container/has_unique_tag_test.go
acornett21: Fixed this everywhere.

certification/internal/policy/container/has_unique_tag.go
acornett21: This has been simplified, into one `if` statement.

PR #761, opened 2022-08-11
Add static check for Check interface

bcrochet:
Add a check in each of the checks to check for the check interface.

### DIFF
--- a/certification/internal/policy/container/base_on_ubi.go
+++ b/certification/internal/policy/container/base_on_ubi.go
@@ -9,6 +9,8 @@ import (
 	pyxis "github.com/redhat-openshift-ecosystem/openshift-preflight/certification/pyxis"
 )
 
+var _ certification.Check = &BasedOnUBICheck{}
+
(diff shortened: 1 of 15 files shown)


komish:
@bcrochet I'm not sure we need this.

We bind our checks to the CheckEngine which only accepts the Check interface, right? You get a compiler error without these assertions.

bcrochet:
> @bcrochet I'm not sure we need this.
> 
> We bind our checks to the CheckEngine which only accepts the Check interface, right? You get a compiler error without these assertions.

That is true. I was thinking though that until you actually do hook it up to the CheckEngine, it would be helpful to start with this assertion while developing a new check. I'm not super behind doing this, just thought it might be a nice to have. But if you think it's extra, I'm happy to just close this and move on.

komish (approved):
Good either way. We can be a little extra. 🎉 

/lgtm

acornett21 (approved):

PR #762, opened 2022-08-15
Update the bundle version check with latest releases

bcrochet:
There are new deprecated APIs in 1.25 and 1.26 of upstream K8s.
Operator-sdk Bundle Validate now supports these deprecations. This
changes the check to encompass more versions without having to change
dramatically as more are added.


acornett21 (approved):

skattoju (approved):

komish (approved):
/lgtm

certification/internal/bundle/bundle.go
```diff
@@ -18,17 +18,27 @@ import (
...
 )
 
-const ocpVerV1beta1Unsupported = "4.9"
-
 // versionsKey is the OpenShift versions in annotations.yaml that lists the versions allowed for an operator
 const versionsKey = "com.redhat.openshift.versions"
 
+// This table signifies what the NEXT release of OpenShift will
+// deprecate, not what it matches up to.
+var ocpVerV1beta1Unsupported = map[string]string{
```
komish, line 26: Should we rename this variable? It doesn't mean anything to me right now. Maybe I'm missing something

certification/internal/bundle/bundle.go
```diff
@@ -51,75 +61,69 @@ func Validate(ctx context.Context, operatorSdk operatorSdk, imagePath string) (*
...
-			log.Error("unable to parse the version")
-			return false
+			return fmt.Sprintf("%d.%d", verParsed.Major, verParsed.Minor), nil
 		}
 
-		if semVerOCPV1beta1Unsupported.GE(verParsed) {
-			return true
-		}
+		// This is an open-ended range: v1-. This is not valid.
+		// So, we just fall through to the default return.
```
komish, line 124: It would seem that we _know_ the reason why this hitting this part of the code is invalid, but falling through to the default return implies we don't know why. Should we just return with a useful error message, even if it's an extra return?

komish (commented):
Few comments

certification/internal/bundle/bundle.go
bcrochet: Valid.

certification/internal/bundle/bundle.go
bcrochet: Valid.

komish (approved):
/lgtm

PR #768, opened 2022-08-18
Remove the project type when submitting updates to pyxis

### LINKED ISSUE #766: UpdateProject Fails for Projects of `scratch` Type
## Bug Description
Since preflight is sending back all the data to pyxis that it received, pyxis throws a validation error 
```
" received non 200 status code in updateProject: {\"detail\": {\"container\": {\"type\": [\"Partners can set a container type only to one of ['container', 'operator bundle image']\"]}}, \"status\": 400, \"title\": \"Bad Request\", \"type\": \"about:blank\", \"trace_id\": \"0xefcd96533f1e57be9089971990da23ec\"}"
```

main

## Steps to Reproduce:
1) Have someone setup a project as `scratch` (we don't have access to do this)
2) run `preflight check container --submit` for the above project
3)

## Expected Result
I'd expect the partner to be able to submit their image.

## Actual Result
Pyxis throws a validation error, since partners are not allowed to set the project type as `scratch` which is caused by preflight sending back all of the data we recieved.

## Additional Context
preflight should only send the data we have set/updated in our PATCH request to pyxis.

When this issues is close so should this [JIRA](https://issues.redhat.com/browse/ISVISSUE-196)

komish:
Fixes #766

The issue listed above results from our CertProject struct, which contains the project type. Certain values for the project's type cannot be set via the external API, but our submission included them, resulting in a rejection from the API.

This patch adjusts the CertProject.Type key such that it uses omitempty, and on update calls, we zero out the value of CertProject.Type before we submit.

We'll want to cut a beta release after this merges, which can then be tested by the blocked partner.

### DIFF
--- a/certification/pyxis/pyxis.go
+++ b/certification/pyxis/pyxis.go
@@ -299,7 +299,19 @@ func (p *pyxisClient) GetProject(ctx context.Context) (*CertProject, error) {
 }
 
 func (p *pyxisClient) updateProject(ctx context.Context, certProject *CertProject) (*CertProject, error) {
-	b, err := json.Marshal(certProject)
+	// We cannot send the project type to pyxis in a Patch.
+	// Copy the CertProject and strip the Type value to
+	// have omitempty skip the key in the JSON patch.
+	patchCertProject := &CertProject{
+		ID:                  certProject.ID,
+		CertificationStatus: certProject.CertificationStatus,
+		Container:           certProject.Container,
+		Name:                certProject.Name,
+		ProjectStatus:       certProject.ProjectStatus,
+		// Do not copy the Type.
+	}
+
+	b, err := json.Marshal(patchCertProject)
 	if err != nil {
 		return nil, fmt.Errorf("could not marshal certProject: %w", err)
 	}
--- a/certification/pyxis/types.go
+++ b/certification/pyxis/types.go
@@ -99,9 +99,9 @@ type CertProject struct {
 	ID                  string    `json:"_id,omitempty"`
 	CertificationStatus string    `json:"certification_status" default:"In Progress"`
 	Container           Container `json:"container"`
-	Name                string    `json:"name"`                      // required
-	ProjectStatus       string    `json:"project_status"`            // required
-	Type                string    `json:"type" default:"Containers"` // required
+	Name                string    `json:"name"`           // required
+	ProjectStatus       string    `json:"project_status"` // required
+	Type                string    `json:"type,omitempty"` // required
 }
 
 type Container struct {


certification/pyxis/types.go
```diff
@@ -99,9 +99,9 @@ type CertProject struct {
 	ID                  string    `json:"_id,omitempty"`
 	CertificationStatus string    `json:"certification_status" default:"In Progress"`
 	Container           Container `json:"container"`
-	Name                string    `json:"name"`                      // required
-	ProjectStatus       string    `json:"project_status"`            // required
-	Type                string    `json:"type" default:"Containers"` // required
+	Name                string    `json:"name"`           // required
+	ProjectStatus       string    `json:"project_status"` // required
+	Type                string    `json:"type,omitempty"` // required
```
komish, line 104: A note to reviewers: These lines all changed because the comment's alignment changed.

certification/pyxis/pyxis.go
```diff
@@ -299,6 +299,10 @@ func (p *pyxisClient) GetProject(ctx context.Context) (*CertProject, error) {
 }
 
 func (p *pyxisClient) updateProject(ctx context.Context, certProject *CertProject) (*CertProject, error) {
+	// We cannot send the project type to pyxis in a Patch.
+	// Rely on omitempty to strip it from the JSON submission.
+	certProject.Type = ""
```
bcrochet, line 304: This will change the project from the caller, since it's a pointer. Are we sure we want to do this?

skattoju: Don't think this is used anywhere.. except in tests.. but probably shouldn’t modify it either..

komish: Fair - Fixed, latest commit copies all of the data into a new struct.

certification/pyxis/pyxis.go
```diff
@@ -299,7 +299,19 @@ func (p *pyxisClient) GetProject(ctx context.Context) (*CertProject, error) {
 }
 
 func (p *pyxisClient) updateProject(ctx context.Context, certProject *CertProject) (*CertProject, error) {
-	b, err := json.Marshal(certProject)
+	// We cannot send the project type to pyxis in a Patch.
+	// Copy the CertProject and strip the Type value to
+	// have omitempty skip the key in the JSON patch.
+	patchCertProject := &CertProject{
```
bcrochet, line 305: Just a suggestion, unless you want to be as explicit as you are. Feel free to ignore.

```suggestion
	patchCertProject := *certProject
	patchCertProject.Type = ""
```

acornett21 (approved):

PR #770, opened 2022-08-18
Switch the bundle validate to use the API

bcrochet:
This uses the validators directly, but still has the same usage
semantics, for the most part.
    
### DIFF
--- a/certification/internal/policy/operator/deployable_by_olm_test.go
+++ b/certification/internal/policy/operator/deployable_by_olm_test.go
@@ -3,13 +3,11 @@ package operator
 import (
 	"context"
 	"os"
-	"path/filepath"
 	"time"
 
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification"
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/artifacts"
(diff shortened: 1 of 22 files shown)


bcrochet:
Waiting on: https://github.com/operator-framework/api/pull/255

certification/internal/policy/operator/deployable_by_olm_test.go
```diff
@@ -155,7 +92,7 @@ var _ = Describe("DeployableByOLMCheck", func() {
 		})
 		Context("When index image is in a private registry and CSV has been created successfully", func() {
 			BeforeEach(func() {
-				deployableByOLMCheck.dockerConfig = filepath.Join(tmpDockerDir, registryConfigDir, registryConfigFilename)
+				deployableByOLMCheck.dockerConfig = "./testdata/dockerconfig.json"
```
acornett21, line 95: Can we comment that this file is intended to be empty? Seeing that empty file in the test data confused me till I came to this code.

komish (commented):
Looks good for the most part. I think part of the challenge I have with this particular collection of checks is that the logic is feels somewhat tricky to follow. That's neither here nor there for this PR, though.

komish (approved):

acornett21 (approved):

PR #772, opened 2022-08-22
Fix some tests that would fail with existing env vars

bcrochet:
Some of the unit tests would fail if KUBECONFIG or PFLT_INDEXIMAGE
were already present in the environment. This patch cleans those up
a bit by checking for the existence, clearing them when required, and
cleaning up appropriately after the test is run.

This also includes minor renaming of some of the test suites, as they
overlapped, and it was confusing where the error was actually coming
from.

### DIFF
--- a/certification/runtime/config_test.go
+++ b/certification/runtime/config_test.go
@@ -1,6 +1,7 @@
 package runtime
 
 import (
+	"os"
 	"reflect"
 
 	. "github.com/onsi/ginkgo/v2"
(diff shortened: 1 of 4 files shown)


acornett21 (approved):

certification/runtime/config_test.go
```diff
@@ -15,6 +16,10 @@ var _ = Describe("Viper to Runtime Config", func() {
 		baseViperCfg = viper.New()
 		expectedRuntimeCfg = &Config{}
 
+		if val, ok := os.LookupEnv("KUBECONFIG"); ok {
```
komish, line 19: NIT: good to merge without this change.
```suggestion
		if val, isSet := os.LookupEnv("KUBECONFIG"); isSet {
```

komish: Happens in a couple of places, but because it's a nit, I'm only mentioning it here.

komish (approved):
/lgtm

certification/runtime/config_test.go
bcrochet: Updated.

skattoju (approved):

Issue #758, opened 2022-08-09
certification-project-id with multiple dashes causes 404 status when trying to submit


jrsmroz:
## Bug Description

When a certification-project-id has multiple dashes, the preflight fails with 404 error as a result of truncating the project id.

```console
$ docker run                                                              \
  -v <my_home_path>/.docker/:/docker/                                     \
  quay.io/opdev/preflight:stable                                          \
  check container scan.connect.redhat.com/ospid-f8304390-ff4c-4b7f-8c7e-1ed13370b3c2/nightly-ingress-controller:nightly-redhat \
  --docker-config=/docker/config.json                                     \
  --submit                                                                \
  --certification-project-id="ospid-f8304390-ff4c-4b7f-8c7e-1ed13370b3c2" \
  --pyxis-api-token="<my-pyxis-token>"                                    \
  --loglevel=debug

time="2022-08-09T22:31:47Z" level=debug msg="config file not found, proceeding without it"
time="2022-08-09T22:31:47Z" level=info msg="certification library version 1.3.3 <commit: 6c30f7caae731ecc2faeaebd78c6d3981c708fe6>"
time="2022-08-09T22:31:47Z" level=debug msg="URL is: https://catalog.redhat.com/api/containers/v1/projects/certification/id/f8304390"
Error: could not retrieve project: status code: 404: body: {"detail": "The requested URL was not found on the server. If you entered the URL manually please check your spelling and try again.", "status": 404, "title": "Not Found", "type": "about:blank", "trace_id": "0x7ded39401a01e8813f796704de288202"}
Usage:
  preflight check container [flags]

Examples:
  preflight check container quay.io/repo-name/container-name:version

Flags:
      --certification-project-id string   Certification Project ID from connect.redhat.com/projects/{certification-project-id}/overview
                                          URL paramater. This value may differ from the PID on the overview page. (env: PFLT_CERTIFICATION_PROJECT_ID)
  -h, --help                              help for container
      --pyxis-api-token string            API token for Pyxis authentication (env: PFLT_PYXIS_API_TOKEN)
      --pyxis-env string                  Env to use for Pyxis submissions. (default "prod")
      --pyxis-host string                 Host to use for Pyxis submissions. This will override Pyxis Env. Only set this if you know what you are doing.
                                          If you do set it, it should include just the host, and the URI path. (env: PFLT_PYXIS_HOST)
  -s, --submit                            submit check container results to red hat

Global Flags:
[...]

PR #773, opened 2022-08-22
updating ospid logic to exit if legacy project id is provided

acornett21:
- Fixes: #758 
- Added a `PreRunE` to validate the project id, and throw an error if it's in an improper format is used.

Output from test
```
Examples:
  preflight check container quay.io/repo-name/container-name:version

Flags:
      --certification-project-id string   Certification Project ID from connect.redhat.com/projects/{certification-project-id}/overview
                                          URL paramater. This value may differ from the PID on the overview page. (env: PFLT_CERTIFICATION_PROJECT_ID)
  -h, --help                              help for container
      --pyxis-api-token string            API token for Pyxis authentication (env: PFLT_PYXIS_API_TOKEN)
      --pyxis-env string                  Env to use for Pyxis submissions. (default "prod")
      --pyxis-host string                 Host to use for Pyxis submissions. This will override Pyxis Env. Only set this if you know what you are doing.
                                          If you do set it, it should include just the host, and the URI path. (env: PFLT_PYXIS_HOST)
  -s, --submit                            submit check container results to red hat

Global Flags:
      --artifacts string       Where check-specific artifacts will be written. (env: PFLT_ARTIFACTS)
  -d, --docker-config string   Path to docker config.json file. This value is optional for publicly accessible images.
                               However, it is strongly encouraged for public Docker Hub images,
                               due to the rate limit imposed for unauthenticated requests. (env: PFLT_DOCKERCONFIG)
      --logfile string         Where the execution logfile will be written. (env: PFLT_LOGFILE)
      --loglevel string        The verbosity of the preflight tool itself. Ex. warn, debug, trace, info, error. (env: PFLT_LOGLEVEL)

2022/08/24 16:11:12 certification project id: ospid-62423-f26c346-6cc1dc7fae92 is improperly formatted see help command for instructions on obtaining proper value

```

### DIFF
--- a/certification/runtime/config.go
+++ b/certification/runtime/config.go
@@ -2,7 +2,6 @@ package runtime
 
 import (
 	"os"
-	"strings"
 
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/policy"
 	"github.com/spf13/viper"
@@ -62,13 +61,7 @@ func (c *Config) storeContainerPolicyConfiguration(vcfg viper.Viper) {
 	c.PyxisAPIToken = vcfg.GetString("pyxis_api_token")
 	c.Submit = vcfg.GetBool("submit")
 	c.PyxisHost = pyxisHostLookup(vcfg.GetString("pyxis_env"), vcfg.GetString("pyxis_host"))
-
-	// Strip the ospid- prefix from the project ID if provided.
-	certificationProjectID := vcfg.GetString("certification_project_id")
-	if strings.HasPrefix(certificationProjectID, "ospid-") {
-		certificationProjectID = strings.Split(certificationProjectID, "-")[1]
-	}
-	c.CertificationProjectID = certificationProjectID
+	c.CertificationProjectID = vcfg.GetString("certification_project_id")
 }
 
 // storeOperatorPolicyConfiguration reads operator-policy-specific config
--- a/cmd/check_container.go
+++ b/cmd/check_container.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strings"
 
@@ -27,6 +28,7 @@ func checkContainerCmd() *cobra.Command {
 		Args:  checkContainerPositionalArgs,
 		// this fmt.Sprintf is in place to keep spacing consistent with cobras two spaces that's used in: Usage, Flags, etc
 		Example: fmt.Sprintf("  %s", "preflight check container quay.io/repo-name/container-name:version"),
+		PreRunE: validateCertificationProjectId,
 		RunE:    checkContainerRunE,
 	}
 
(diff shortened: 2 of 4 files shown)


certification/runtime/config.go
```diff
@@ -51,24 +52,36 @@ func NewConfigFrom(vcfg viper.Viper) (*Config, error) {
...
 	c.Submit = vcfg.GetBool("submit")
 	c.PyxisHost = pyxisHostLookup(vcfg.GetString("pyxis_env"), vcfg.GetString("pyxis_host"))
 
 	// Strip the ospid- prefix from the project ID if provided.
 	certificationProjectID := vcfg.GetString("certification_project_id")
 	if strings.HasPrefix(certificationProjectID, "ospid-") {
+		// counting the `-` values, if the user provides more than one, we know they inputted a legacy project id
+		// which can not be used to query pyxis, exiting early to show the help menu
+		if strings.Count(certificationProjectID, "-") != 1 {
+			return errors.Errorf("certification prodject id: %s is improperly formatted see help for instructions on obtaining proper value", certificationProjectID)
```
bcrochet, line 78:
```suggestion
			return errors.Errorf("certification prodject id: %s is improperly formatted, see instructions to obtain the proper value", certificationProjectID)
```

bcrochet: But also... Where are these instructions? A link might help.

certification/runtime/config.go
```diff
@@ -51,24 +52,36 @@ func NewConfigFrom(vcfg viper.Viper) (*Config, error) {
...
+		}
 		certificationProjectID = strings.Split(certificationProjectID, "-")[1]
```
bcrochet, line 80: Have you thought about still doing the strings.Split first, then just checking for the length of the returned slice? If it's > 2, then you would know you have more than one '-'. This would remove the extra scan of the string.

bcrochet (commented):
Couple of suggestions..

certification/runtime/config.go
```diff
@@ -51,24 +52,36 @@ func NewConfigFrom(vcfg viper.Viper) (*Config, error) {
...
 // items in viper, normalizes them, and stores them in Config.
-func (c *Config) storeContainerPolicyConfiguration(vcfg viper.Viper) {
+func (c *Config) storeContainerPolicyConfiguration(vcfg viper.Viper) error {
 	c.PyxisAPIToken = vcfg.GetString("pyxis_api_token")
 	c.Submit = vcfg.GetBool("submit")
 	c.PyxisHost = pyxisHostLookup(vcfg.GetString("pyxis_env"), vcfg.GetString("pyxis_host"))
 
 	// Strip the ospid- prefix from the project ID if provided.
 	certificationProjectID := vcfg.GetString("certification_project_id")
 	if strings.HasPrefix(certificationProjectID, "ospid-") {
```
komish, line 74: Does it make more sense to do these validations _before_ we get here? Like as a `PreRunE`. 

I can see it both ways; it won't take much to sway me one way or the other. Consider this non-blocking. Just thinking out loud.

certification/runtime/config.go
acornett21: We only have that info in our `help` output, I believe.

certification/runtime/config.go
acornett21: I don't think that the flags would be loaded into viper if there was a `PreRunE`, but I might be wrong.

@bcrochet Thoughts on this?

bcrochet: You can add a PreRunE. It should still run the PersistentPreRun.

acornett21: The flags do get loaded before `PreRunE` so I have removed the logic from `config.go` to `check_containter.go` PTAL.

cmd/check_container.go
```diff
@@ -208,3 +210,23 @@ func checkContainerPositionalArgs(cmd *cobra.Command, args []string) error {
 
 	return nil
 }
+
+// validateCertificationProjectId validates that the certification project id is in the proper format
+// and throws an error if the value provided is in a legacy format that is not usable to query pyxis
+func validateCertificationProjectId(cmd *cobra.Command, args []string) error {
+
```
bcrochet, line 217: Extra line.

bcrochet (commented):
Fix the gofumpt error, but otherwise lgtm

komish (approved):
/lgtm

acornett21:
Putting this on hold, there is another use case that we've never covered. ie a user never gives an `ospid` but provides a legacy project id of `123-456-789-000`. Might as well try to cover that case as well.

komish (approved):
/lgtm

bcrochet (approved):
/lgtm

jomkz (approved):

PR #776, opened 2022-08-26
check container type for scratch exceptions instead of os_content_type

komish:
According to a recent review of data, we may be checking the wrong field to determine if a scratch exception is required. This corrects that.


komish:
Got confirmation that container.Type should be the field evaluated for scratch. This should be good for review.

jomkz (approved):

PR #779, opened 2022-09-01
[RFC] Add policy to check for scratch images running privileged

hbrueckner:
The policy for scratch container images includes the check for `RunAsNonRoot` and does not consider the privileged setting in the container project. This PR is about to introduce a scratch container image policy to consider running privileged containers.

This PR arises from discussions on the RH support case 03250672 which triggered issue #766.

Testing the latest `preflight` (main branch) with issue #766 corrected, solved the parsing problem but introduced the `RunAsNonRoot` check again as part of the scratch policy. In preflight v1.4.0, the check has been skipped based on the project setting. This behavior might have been introduced by PR #776 which correctly detects scratch images.

Signed-off-by: Hendrik Brueckner <brueckner@linux.ibm.com>

### DIFF
--- a/certification/engine/engine.go
+++ b/certification/engine/engine.go
@@ -104,6 +104,13 @@ func initializeChecks(ctx context.Context, p policy.Policy, cfg certification.Co
 			&containerpol.HasRequiredLabelsCheck{},
 			&containerpol.RunAsNonRootCheck{},
 		}, nil
+	case policy.PolicyScratchRoot:
+		return []certification.Check{
+			&containerpol.HasLicenseCheck{},
+			containerpol.NewHasUniqueTagCheck(cfg.DockerConfig()),
+			&containerpol.MaxLayersCheck{},
+			&containerpol.HasRequiredLabelsCheck{},
+		}, nil
 	}
 
 	return nil, fmt.Errorf("provided policy %s is unknown", p)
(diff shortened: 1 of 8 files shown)


komish:
/ok-to-test

Thanks @hbrueckner! I'll start tests here for now, I have a few nits I'll throw into a review soon.

cmd/check_container.go
```diff
@@ -161,7 +161,11 @@ func getContainerPolicyExceptions(ctx context.Context, pc pyxisClient) (policy.P
 	}
 	log.Debugf("Certification project name is: %s", certProject.Name)
 	if certProject.Container.Type == "scratch" {
-		return policy.PolicyScratch, nil
+		if certProject.Container.Privileged {
+			return policy.PolicyScratchRoot, nil
+		} else {
+			return policy.PolicyScratchNonRoot, nil
+		}
```
acornett21, line 168:
```suggestion
		if certProject.Container.Privileged {
			return policy.PolicyScratchRoot, nil
		}
		
		return policy.PolicyScratchNonRoot, nil
		
```
We can remove the else and just return `nonroot` as the default

acornett21:
I don't think we have ever supported this use case, we'll check with the business to see if we should support it going forward.

cmd/check_container.go
hbrueckner: @acornett21 Thanks a lot. Done.

komish:
We very much appreciate the contributions here, but we've not heard any confirmation from the Certification policy makers that this combination needs to be handled within Preflight. For now, we'll mark this closed. If the policy stance changes, we'll revisit then.

hbrueckner:
/reopen

Rebased my branch to latest main: https://github.com/hbrueckner/openshift-preflight/tree/pr-scratch-privileged

jfrancin:
I'd absolutely like to see this moved forward. Preflight needs  to properly handle the situation where a container is privileged but is also 'scratch'.

Issue #709, opened 2022-06-24
Enhance Container Submission Flow to Better Handle Timeouts


acornett21:
## Is your feature request related to a problem? Please describe.
When we first build `check container` we only had REST calls to pyxis in the submission flow, now that we have calls to pyxis in the checks, it leaves for a potential undesirable experience for partners. If there is a timeout in the checks, we then post an image with the `certified: false` for the image, and no way to change that, even if the partner re-runs preflight and gets valid results from pyxis on the 2nd, 3rd, nth run.

## Describe the solution you'd like.
I'm not sure if the ideal solution is
- If there is a pyxis timeout in the check, exit preflight/stop the `submit` call
- Do a `POST`, `GET`, and if there is a diff do a `PATCH`  for `/images` (see additional comments below)

## Additional context.
`PUT`, `PATCH` was not available  when we built preflight, and are not in the REST implementation yet, but were added in later see [JIRA](https://issues.redhat.com/browse/ISV-1391)

jfrancin:
Another partner got bit by this same issue over the weekend. Case 03253477 - partner ran Preflight on three images; they all failed with a weird error:

```
time="2022-06-25T00:44:23-04:00" level=error msg="could not marshal GraphQL query: Message: Unexpected error occurred, Locations: []"
time="2022-06-25T00:44:23-04:00" level=error msg="Error when querying pyxis for uncompressed top layer idsMessage: Unexpected error occurred, Locations: []"
time="2022-06-25T00:44:23-04:00" level=error msg="Unable to verify layer hashesMessage: Unexpected error occurred, Locations: []"
time="2022-06-25T00:44:23-04:00" level=info msg="check completed: BasedOnUbi" ERROR="Message: Unexpected error occurred, Locations: []" result="Message: Unexpected error occurred, Locations: []"
```

Subsequent tests today (27 June) passed. Filed Jira https://issues.redhat.com/browse/ISVISSUE-149 to get results changed.

jfrancin:
Partner decided to rebuild (to cause different image SHAs), which allowed the containers to be re-submitted with passing grades, and are now published.

PR #782, opened 2022-09-15
adding a new updateImage method to enable partners to re-run

acornett21:
- Fixes: #709 
- Adding an updateImage method that sets the `certified` value to `true` for a given architecture
- Adding in login into the submit flow to call this value if the previous certified value we get when calling pyxis is `false` and the current execute's certified value is `true`
- Added tests and updated mock data to support new flow


certification/pyxis/pyxis.go
```diff
@@ -127,6 +127,53 @@ func (p *pyxisClient) getImage(ctx context.Context, dockerImageDigest string) (*
 	return &data.Data[0], nil
 }
 
+// updateImage updates the certification status to true for a given architecture of an image
+// this methods purpose is to account for pyxis errors in a given check (http 500 as an example), where the user
+// needs to rerun preflight to get a valid response from pyxis
```
bcrochet, line 132:
```suggestion
// needs to re-run preflight to get a valid response from pyxis
```

certification/pyxis/pyxis.go
```diff
@@ -127,6 +127,53 @@ func (p *pyxisClient) getImage(ctx context.Context, dockerImageDigest string) (*
...
+func (p *pyxisClient) updateImage(ctx context.Context, imageID string, architecture string) (*CertImage, error) {
```
bcrochet, line 133: Why not just pass in a *CertImage?

certification/pyxis/submit.go
```diff
@@ -68,6 +71,16 @@ func (p *pyxisClient) SubmitResults(ctx context.Context, certInput *Certificatio
 		if err != nil {
 			return nil, fmt.Errorf("could not get image: %v", err)
 		}
+
+		// checking to see if the original value is certified and the previous value is not certified,
+		// this would indicate that a partner is running preflight again, and during the first run there was a timeout/error
+		// in a check that interacts with pyxis and we need to correct the certified value for the image
+		if certified && !certImage.Certified {
+			certImage, err = p.updateImage(ctx, certImage.ID, certImage.Architecture)
```
bcrochet, line 79: I think it makes more sense to just pass in  &certImage. If more fields are added later, you won't have to change this, just updateImage.

bcrochet (commented):
Couple of nits in comments and a question about the updateImage signature.

certification/pyxis/submit.go
acornett21: Are you saying build the new struct outside of the `updateImage` method and pass that in? I thought about that, but in updateProject we build the struct in the method, so I was trying to be consistent. Both ways make sense to me, and I have no strong opinion either way.

certification/pyxis/pyxis.go
```diff
@@ -127,6 +127,53 @@ func (p *pyxisClient) getImage(ctx context.Context, dockerImageDigest string) (*
...
+// updateImage updates the certification status to true for a given architecture of an image
+// this method's purpose is to account for pyxis errors in a given check (HTTP 500 as an example), where the user
+// needs to re-run preflight to get a valid response from pyxis
+func (p *pyxisClient) updateImage(ctx context.Context, imageID string, architecture string) (*CertImage, error) {
+	// instantiating a new certImage struct, if updateImage is called we know that we need to set the certified status to true
+	certImage := &CertImage{
+		ID:           imageID,
+		Architecture: architecture,
+		Certified:    true,
+	}
```
komish, line 139: I don't love auto-setting `Certified: true`. If we're going to do this, we should rename this method.

I know it's a private method, but I would still rather the method name align with this predefined behavior.

certification/pyxis/pyxis.go
```diff
@@ -127,6 +127,53 @@ func (p *pyxisClient) getImage(ctx context.Context, dockerImageDigest string) (*
...
+	resp, err := p.Client.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("cannot update image in pyxis: %w", err)
+	}
+
+	defer resp.Body.Close()
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, fmt.Errorf("could not read body: %w", err)
```
komish, line 159:
```suggestion
		return nil, fmt.Errorf("could not read response body: %w", err)
```

certification/pyxis/pyxis.go
```diff
@@ -127,6 +127,53 @@ func (p *pyxisClient) getImage(ctx context.Context, dockerImageDigest string) (*
...
+			resp.StatusCode,
+			string(body))
+	}
+
+	var newCertImage CertImage
+	if err := json.Unmarshal(body, &newCertImage); err != nil {
+		return nil, fmt.Errorf("could not unmarshal body: %s: %w", string(body), err)
+	}
+
+	return &newCertImage, nil
```
komish, line 174:
```suggestion
	return &updatedCertImage, nil
```

certification/pyxis/pyxis.go
acornett21: Seems like you and @bcrochet agree this method should be refactored to just take a `&CertImage` and submit do the proper work. I'll make that change.

certification/pyxis/submit.go
komish: I left a comment elsewhere, but I'm of the opinion that `updateImage` is probably not how we want to refer to this method given its context, where by it only updates certified to true.

I'm not sure if passing in a certImage makes sense to me given that context. Do we not already send an Update call to Pyxis in this workflow? Or is the expectation that we're always sending a Create call? If we're looking to build a separate workflow for Update, then maybe we need to detect earlier in Submit whether we need to go down an Update (because the image already exists) or Create (because the image doesn't already exist) workflow and handle things independently?

Just a few thoughts. Not sure where I land on it, but I'm less close to the issue at hand.

komish (changes requested):

komish (approved):
My comments have been address.

/lgtm

certification/pyxis/submit.go
acornett21: Since we don't know if a user is running preflight for the first time or nth time for a given digest/arch combination, we built the submit flow for `images` and `rpm` manifest, to first call `create`, then if we get a `conflict` then we would call a `get`.

This is just an extension of that logic, with the addition that we need to correct the certification value from a previous run. Do we buy anything from switching to a `get` first? then falling back to create if `not found`? That seems like a larger refactor and a higher chance of getting something wrong, then we also should switch RPM manifest calls to follow that logic so we are consistent.

That's just my take, I could be being short sighted.

certification/pyxis/pyxis.go
```diff
@@ -127,6 +127,44 @@ func (p *pyxisClient) getImage(ctx context.Context, dockerImageDigest string) (*
 	return &data.Data[0], nil
 }
 
+// updateImage updates a given certification image based on how the image is built in the `submit` flow
+func (p *pyxisClient) updateImage(ctx context.Context, certImage *CertImage) (*CertImage, error) {
+	b, err := json.Marshal(certImage)
```
bcrochet, line 132: I still think you should create a new CertImage to pass to the PATCH call. Passing in the *CertImage was just to keep the signature from having to grow if we ever needed to add another field.

certification/pyxis/submit.go
```diff
@@ -68,6 +71,24 @@ func (p *pyxisClient) SubmitResults(ctx context.Context, certInput *Certificatio
...
 			return nil, fmt.Errorf("could not get image: %v", err)
 		}
+
+		// checking to see if the original value is certified and the previous value is not certified,
+		// this would indicate that a partner is running preflight again, and during the first run there was a timeout/error
+		// in a check that interacts with pyxis and we need to correct the certified value for the image
+		if certified && !certImage.Certified {
+			// instantiating an updateCertImage struct, when calling updateImage we need to provide the ID, architecture,
+			// and change the certified value to `true`
+			updatedCertImage := &CertImage{
```
bcrochet, line 81: I would still do this in the method, not here.

bcrochet (changes requested):

certification/pyxis/pyxis.go
acornett21: I initially had the struct to post in the updateImage method, but moved it out since I thought that's what we were going for, and was probably confused by the ask. I moved it back here.

certification/pyxis/submit.go
acornett21: The only thing that I kept here is changing the certified flag to come from the original certImage, ie the one we create during the current execution. Then I pass in the struct that we got from the `GET` call to pyxis.

certification/pyxis/submit_test.go
```diff
@@ -19,7 +19,9 @@ var _ = Describe("Pyxis Submit", func() {
 	mux.Handle("/api/v1/projects/certification/id/my-awesome-project-id/test-results", &pyxisTestResultsHandler{})
 	mux.Handle("/api/v1/projects/certification/id/my-image-project-id/images", &pyxisImageHandler{})
 	mux.Handle("/api/v1/projects/certification/id/", &pyxisProjectHandler{})
+	mux.Handle("/api/v1/images/id/updateImage", &pyxisImageHandler{})
 	mux.Handle("/api/v1/images/id/blah/", &pyxisRPMManifestHandler{})
+	mux.Handle("/api/v1/images/id/updateImage/", &pyxisImageHandler{})
```
skattoju, line 24: Is this a duplicate ?

acornett21: Not a duplicate, this is to force the new tests back to the `pyxisImageHandler` where the `PATCH` mock is at. Since, RPM manifest route is subset of the image route, I had to add this in. The below picture from swagger might make this make more sense.

image

bcrochet (approved):
/lgtm

komish:
Still LGTM!

PR #786, opened 2022-09-21
Introduce afero to simplify FS operations

bcrochet:
* Introduce afero in the artificts package
* Redo the signature of artifacts.WriteFile to accept a reader instead of a string. Updates all locations of WriteFile
* Rework a test that wasn't properly checking for failure in its BeforeEach

This proves the concept. More work can be done to use afero in unit tests to make them simpler.


certification/internal/engine/engine.go
```diff
@@ -293,7 +293,7 @@ func generateBundleHash(bundlePath string) (string, error) {
 		hashBuffer.WriteString(fmt.Sprintf("%s  %s\n", k, files[k]))
 	}
 
-	_, err := artifacts.WriteFile("hashes.txt", hashBuffer.String())
+	_, err := artifacts.WriteFile("hashes.txt", &hashBuffer)
```
acornett21, line 296: WriteFile doesn't take a pointer, so how does this work?

bcrochet: No, but it takes an interface. And *bytes.Buffer satisfies io.Reader. It has to be a *bytes.Buffer, and not a bytes.Buffer, since the receiver for bytes.Buffer's Read method is a pointer.

acornett21: I forgot io.Reader is an interface.

acornett21 (approved):

certification/artifacts/artifacts.go
```diff
@@ -20,59 +25,40 @@ const DefaultArtifactsDir = "artifacts"
 
 func init() {
 	// set the singleton to the default value.
-	ads = DefaultArtifactsDir
+	cwd, _ := os.Getwd()
+	ads = filepath.Join(cwd, DefaultArtifactsDir)
```
komish, line 29: Could we call `Reset()` here?

certification/artifacts/artifacts.go
```diff
@@ -20,59 +25,40 @@ const DefaultArtifactsDir = "artifacts"
...
-		// Fatal does an os.Exit. If we can't create the artifacts directory,
-		// we can't continue.
-		log.Fatal(fmt.Errorf("could not create artifact path: %v", err))
-	}
+func WriteFile(filename string, contents io.Reader) (string, error) {
 	fullFilePath := filepath.Join(Path(), filename)
 
-	if err := os.WriteFile(fullFilePath, []byte(contents), 0o644); err != nil {
+	if err := afero.SafeWriteReader(appFS, fullFilePath, contents); err != nil {
 		return fullFilePath, fmt.Errorf("could not write file to artifacts diretory: %v", err)
```
komish, line 57: Unrelated, but since we're here:

```suggestion
		return fullFilePath, fmt.Errorf("could not write file to artifacts directory: %v", err)
```

certification/artifacts/artifacts.go
```diff
@@ -20,59 +25,40 @@ const DefaultArtifactsDir = "artifacts"
...
 // Returns the full path (including the artifacts dir)
-func WriteFile(filename, contents string) (string, error) {
-	artifactDir := ads
-	_, err := createArtifactsDir(artifactDir)
-	if err != nil {
-		// Fatal does an os.Exit. If we can't create the artifacts directory,
-		// we can't continue.
-		log.Fatal(fmt.Errorf("could not create artifact path: %v", err))
-	}
+func WriteFile(filename string, contents io.Reader) (string, error) {
```
komish, line 53: We're no longer creating the artifacts directory on write. Is that intentional? If so, we need to update this comment.

komish: I'm reading `SafeWriteReader` which looks like it creates parent directories, so we may still be doing the same thing here. Just make sure I didn't misread that if you don't mind.

That said, it looks like this will throw an error if we try to overwrite a file:

https://github.com/spf13/afero/blob/v1.9.2/util.go#L84-L86

On a single iteration, I think this is fine. I wonder if subsequent iterations becomes more frustrating to deal with because it requires an `rm -rf artifacts/` before a re-execution.

komish (commented):
Just a few comments. Overall, it looks good.

I don't know enough about afero to determine exactly how it's interacting with the underlying FS on a given system, but at the API level, the logic LGTM.

certification/artifacts/artifacts.go
bcrochet: I had that same thought. I did it in a previous iteration, but changed it. But coming back to it, I think I will.

certification/artifacts/artifacts.go
bcrochet: Surprised the linter didn't catch that.

certification/artifacts/artifacts.go
bcrochet: You are correct on all counts. 

As for the need for an rm -rf... I would actually prefer that be the case. I don't think we really want artifacts from different runs mixed up anyway. 

We could use WriteReader instead. It skips the Exists check, but keeps all of the other conveniences.

komish: Leave it until it becomes a problem. Probably best folks clean up artifacts in between runs anyway, as you indicate.

komish (approved):
/lgtm

Issue #639, opened 2022-05-18
Propose external library for sharing logic for operator checks


sebrandon1:
## Is your feature request related to a problem? Please describe.
The [CNF Test Suite](https://github.com/test-network-function/cnf-certification-test) team and the Preflight team are both working to verify customer workloads and we would like to use some of the logic/checks that the preflight repo does for operators (and maybe even containers) from [here](https://github.com/redhat-openshift-ecosystem/openshift-preflight/tree/main/certification/internal/policy/operator) and bring that same logic into the CNF test suite.

## Describe the solution you'd like.
Some sort of consumable, external library (redhat-openshift-ecosystem/preflight-lib or some-name) where we can share the logic for performing checks against customer workloads.  We could/should also think about sharing the logic we have on our side as well for some additional collaboration.

sebrandon1:
Maybe it doesn't even need to be that extreme as the full check being broken out but possibly just the engine(s).

The wrapper around the operator-sdk commands would be cool to be able to consume for example:  https://github.com/redhat-openshift-ecosystem/openshift-preflight/blob/main/certification/internal/engine/operatorsdk.go

sebrandon1:
I have been working on implementing this the last couple of weeks and I think I'm at the point where I'm ready to demo/ask additional questions of the preflight team to help me complete the work.

Here is my WIP PoC PR for the work in the TNF suite: https://github.com/test-network-function/cnf-certification-test/pull/592
Here is my branch in my fork of the preflight repo where I'm building out the preflight "lib" functionality: https://github.com/sebrandon1/openshift-preflight/tree/preflight_lib_test

Essentially what I'm doing is trying to expose the functionality already living in preflight to be able to be consumed by the TNF suite.

![image](https://user-images.githubusercontent.com/4563082/196712181-416dba07-ce86-4123-b3fc-0c42c6d24bdc.png)

Currently the preflight repo uses its `cmd` package to interact with the user + build test runners based on parameters passed in via the CLI.

What I did is break the the cobra/viper logic away from the test builder logic and put that into the "lib" package and corresponding folder/files.  By doing this, I can reference the `lib` package directly from the TNF repo and pass in the parameters I want to kick off the tests.

The TNF repo also uses Ginkgo so I report the results in the claim file output and it looks like a native test was ran as part of the TNF code.

Outstanding questions I need some clarification on:
[...]

PR #809, opened 2022-10-20
Add lib to preflight

sebrandon1:
Work in progress!

Related to: 
- https://github.com/test-network-function/cnf-certification-test/pull/592
- https://github.com/redhat-openshift-ecosystem/openshift-preflight/issues/639

This PR provides some changes needed to implement the `lib`.  (Name is subject to change 😄)

I want to go through some of the files to describe my notable changes:

- certification/runtime/config.go
In this file, I added `NewManualContainerConfig` and `NewManualOperatorConfig` for being able to create config objects geared towards their individual purposes.

- cmd/check_container.go and cmd/check_operator.go
These files have had their "NewOperator/Container" funcs moved to the lib.

- fakes.go and types.go
Moved to the lib folder, because it contains funcs like `NewFakePyxisClientNoop` and `NewNoopSubmitter`.
The `NoopSubmitter` has had appropriate setters added.

- lib/check_container_test.go and cmd/check_container_test.go
I split these two files into files more geared towards their testing purposes.  The `cmd` package one tests stuff that has to do with the `cmd` package, for example.


cmd/check_test.go
```diff
@@ -43,30 +44,13 @@ var _ = Describe("cmd package check command", func() {
 			os.Unsetenv("PFLT_PYXIS_ENV")
 			os.Unsetenv("PFLT_PYXIS_HOST")
 		})
-		Context("Regular Connect URL", func() {
```
acornett21, line 46: I feel like these 2 tests should still be valid.

go.mod
```diff
@@ -47,8 +45,8 @@ require (
 	github.com/fsnotify/fsnotify v1.5.4 // indirect
 	github.com/go-logr/logr v1.2.3 // indirect
 	github.com/go-openapi/jsonpointer v0.19.5 // indirect
-	github.com/go-openapi/jsonreference v0.19.5 // indirect
-	github.com/go-openapi/swag v0.19.14 // indirect
+	github.com/go-openapi/jsonreference v0.20.0 // indirect
```
acornett21, line 48: If none of the direct dependencies are changing, how are the indirect ones changing? Should this and the sum file really be committed?

lib/fakes.go
```diff
@@ -1,4 +1,4 @@
-package cmd
+package lib
```
acornett21, line 1: NIT: Should we keep the name of this file as `fakes_test.go` for clarity?

acornett21 (changes requested):

cmd/check_test.go
sebrandon1: Yes, I agree.  I can't remember why they were removed.  I'll add them back.

go.mod
sebrandon1: Agreed as well.  I wasn't changing any of the dependencies so I was confused when these files changed.  I will make sure they match the `main` branch.

sebrandon1: Fixed.  Not sure why they were different.

cmd/check_test.go
sebrandon1: Added the tests back.

lib/fakes.go
sebrandon1: Actually I think this one should remain `fakes.go` and the file that references the fake funcs `lib/preflight_check.go` should be possibly renamed to `lib/preflight_check_test.go`.

sebrandon1: I renamed `lib/preflight_check.go` to `lib/preflight_check_test.go` to better reflect what's happening as it's just running some ginkgo unit tests.

lib/lib.go
```diff
@@ -0,0 +1,229 @@
...
+)
+
+// CheckContainerRunner contains all of the components necessary to run checkContainer.
+type CheckContainerRunner struct {
+	Cfg       *runtime.Config
+	Pc        PyxisClient
+	Eng       engine.CheckEngine
+	Formatter formatters.ResponseFormatter
+	Rw        ResultWriter
+	Rs        ResultSubmitter
```
komish, line 27: Not sure we want this exported (and for the equivalent operator struct)

lib/lib.go
```diff
@@ -0,0 +1,229 @@
...
+
+	return &CheckContainerRunner{
+		Cfg:       cfg,
+		Pc:        pyxisClient,
+		Eng:       engine,
+		Formatter: fmttr,
+		Rw:        &runtime.ResultWriterFile{},
+		Rs:        rs,
+	}, nil
+}
```
komish, line 65: Same here (and for the equivalent operator initializer)

lib/lib.go
```diff
@@ -0,0 +1,229 @@
...
+
+	return &CheckOperatorRunner{
+		Cfg:       cfg,
+		Eng:       engine,
+		Formatter: fmttr,
+		Rw:        &runtime.ResultWriterFile{},
+	}, nil
+}
+
+// resolveSubmitter will build out a resultSubmitter if the provided pyxisClient, pc, is not nil.
```
komish, line 99:
```suggestion
// ResolveSubmitter will build out a resultSubmitter if the provided pyxisClient, pc, is not nil.
```

lib/lib.go
```diff
@@ -0,0 +1,229 @@
...
+
+	// if a partner sets `Host Level Access` in connect to `Privileged`, enable RootExceptionContainerPolicy checks
+	if certProject.Container.Privileged {
+		return policy.PolicyRoot, nil
+	}
+	return policy.PolicyContainer, nil
+}
+
+// PreflightCheck executes checks, interacts with pyxis, format output, writes, and submits results.
+func PreflightCheck(
```
komish, line 138: Not sure we need this exported.

lib/types.go
```diff
@@ -13,23 +13,24 @@ import (
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification"
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/artifacts"
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/pyxis"
+	"github.com/spf13/viper"
 
 	log "github.com/sirupsen/logrus"
 )
 
 // resultWriter defines methods associated with writing check results.
```
komish, line 21:
```suggestion
// ResultWriter defines methods associated with writing check results.
```

lib/types.go
```diff
@@ -13,23 +13,24 @@ import (
...
-type resultWriter interface {
+type ResultWriter interface {
 	OpenFile(name string) (io.WriteCloser, error)
 	io.WriteCloser
 }
 
 // resultSubmitter defines methods associated with submitting results to Red HAt.
```
komish, line 27:
```suggestion
// ResultSubmitter defines methods associated with submitting results to Red HAt.
```

lib/types.go
```diff
@@ -13,23 +13,24 @@ import (
...
-type resultSubmitter interface {
+type ResultSubmitter interface {
 	Submit(context.Context) error
 }
 
 // pyxisClient defines pyxis API interactions that are relevant to check executions in cmd.
```
komish, line 32:
```suggestion
// PyxisClient defines pyxis API interactions that are relevant to check executions in cmd.
```

lib/types.go
```diff
@@ -155,38 +156,46 @@ func (s *containerCertificationSubmitter) Submit(ctx context.Context) error {
...
 }
 
-var _ resultSubmitter = &noopSubmitter{}
+func NewNoopSubmitter(emitLog bool, reason string, log *log.Logger) *NoopSubmitter {
+	return &NoopSubmitter{
+		emitLog: emitLog,
+		reason:  reason,
+		log:     log,
+	}
+}
```
komish, line 194: If we're going to export the struct, we may as well export the fields and remove the setters, I think?

komish (dismissed):
@sebrandon1 Thanks a ton for the work on pushing this forward! I've added a few review comments, but I'll definitely need to take another pass. Before I do - discussion points ⬇️ 

We briefly discussed this when we met, but I wanted to go ahead and capture it here publicly so that we can have folks comment/provide feedback.

I think what I'm looking to see here out of the gate is the ability to run `CheckContainer(...)` and `CheckOperator(...)` for starters.

The inputs are up for debate (my ideal would be just a context and maybe your image string), but the outputs here would be results, and somehow, log files where relevant (TBD, but I'm putting that aside for now).

The goal here would be to make sure that folks starting to consume preflight as a library don't have to go through the hurdles to set up the PreflightCheck configurations, etc.

To that end, I'm thinking (sorry @bcrochet) that we start by moving most of this lib work into a top-level `internal` package, accessible by both `cmd` and everything else, wire up `CheckOperator` and `CheckContainer` as callable functions/methods, and then if we need to pull more things out of this internal library, for your use case (and the use case of others), then we'd do that as well.

If we did that, though, we want to make sure that the exported functions are actually what you need to benefit from calling the Preflight library in your use cases. Does that sound like something that would work for you?

@sebrandon1 @acornett21 @bcrochet WDYT?

lib/lib.go
sebrandon1: This is how I kick off the tests right now.  But if we go with your proposal this could all be kept internal.

How I kick it off for example:
```
err = plib.PreflightCheck(context.TODO(), runner.Cfg, runner.Pc, runner.Eng, runner.Formatter, runner.Rw, runner.Rs)
		if err != nil {
			return err
		}
```

sebrandon1:
@komish Thanks for the review!  Yes, actually what you propose makes a lot of sense by pushing some of that logic farther into the code and only getting two functions `CheckContainer(...) and CheckOperator(...)` with appropriate parameters exposed that would make things much easier.  Those two functions could return your `runtime.Results` structs and handle everything internally versus me interpreting a JSON blob and unmarshalling it back into your structs.  Removes a lot of complexity on the whoever "consumes" the lib so I'm fine with that.

With my initial PoC approach I didn't want to _change_ too much to get this working but this makes more sense for future-proofing the effort.  Very exciting!

komish:
@sebrandon1 for sure. I’ll give some folks time to look and then we can chart the path to get this merged!

lib/types.go
acornett21: That makes logical sense to me as well.

cmd/check_container_test.go
```diff
@@ -574,4 +137,26 @@ certification_project_id: mycertid`
 			})
 		})
 	})
+
+	Context("When instantiating a checkContainerRunner", func() {
+		var cfg *runtime.Config
+
```
komish, line 143: The `cfg` in this block is missing initialization now, which causes tests to panic. I believe the block that is missing looks like this:

```
		BeforeEach(func() {
			cfg = &runtime.Config{
				Image:          "quay.io/example/foo:latest",
				ResponseFormat: formatters.DefaultFormat,
			}
		})
```

komish:
@sebrandon1 I'll have you move `lib` to `internal/lib`, and then rewire the import statements to use that new path. From there, we'll expose hooks for you to be able to run the checks. This temporarily breaks your use case, but we'll fast-follow this on our end and I'll ping you into the PR that adds the exported functionality for your use case.

lib/types.go
sebrandon1: I think this also could be kept internal now using @komish's proposal.  👍

sebrandon1:
@komish Sounds excellent!  I'll do that right now.

sebrandon1:
Okay in [7a84034](https://github.com/redhat-openshift-ecosystem/openshift-preflight/pull/809/commits/7a840340848c31f59c81fc2b31ab9086980bb95d) I moved the lib into a new "internal" folder in the top level and changed the references.

komish:
@sebrandon1 

The CI failure reports:

```
internal/lib/lib.go:10                goimports  File is not `goimports`-ed with -local github.com/redhat-openshift-ecosystem/openshift-preflight
internal/lib/types.go:12              goimports  File is not `goimports`-ed with -local github.com/redhat-openshift-ecosystem/openshift-preflight
certification/internal/lib/lib.go:10  goimports  File is not `goimports`-ed with -local github.com/redhat-openshift-ecosystem/openshift-preflight
```

Given that's the only failure I see right at this moment, I think running this and committing the changes should fix the issue:

```
./bin/golangci-lint run --fix
```

sebrandon1:
Okay fixed.  I ran `./bin/golangci-lint run --fix`.

cmd/check_container_test.go
sebrandon1: Fixed.

certification/runtime/config.go
```diff
@@ -56,6 +56,28 @@ func NewConfigFrom(vcfg viper.Viper) (*Config, error) {
...
+		Image:          image,
+		Submit:         false, // operator results are not submitted
+		WriteJUnit:     writeJUnit,
+		ResponseFormat: responseFormat,
+		Artifacts:      artifactsDir,
+		Bundle:         false,
+		Scratch:        false,
+	}
+}
+
```
komish, line 80: @sebrandon1 let's go ahead and remove these for now. I believe this is part of what's driving down coverage.

sebrandon1: Yup sounds good.  They won't be needed in the future implementation.

komish:
Added on two commits to fix the coverage issues and fix a couple of case issues in comments for exported functions.

Coverage issues were fixed by:
- adding suite_test.go to the `internal/lib`. They were being skipped.
- moving NewCheckOperatorRunner-related tests to `internal/lib` where that function now lives
- moving fakes.go to fakes_test.go which excludes it in coverage evaluation.

komish:
https://github.com/redhat-openshift-ecosystem/openshift-preflight/pull/809/commits/0d08e685ab63ee3e980cf0a75274382d7ab19239 removes a duplicate bit of code and moves the test to the correct location.

komish:
@sebrandon1 @bcrochet @acornett21 as I've made changes here, PTAL.

komish:
> 



> * FAILURE https://www.distributed-ci.io/jobs/84825c44-a20d-4c21-845d-0aec82dd6322/jobStates
>   from [Add lib to preflight #809](https://github.com/redhat-openshift-ecosystem/openshift-preflight/pull/809)

@tkrishtop Just a heads up that it seems like this is a problem in the infrastructure's ability to RW files on the filesystem (permission denied) and doesn't seem related to the PR's changes at first glance.

acornett21:
@komish I think this addresses everything, the only thing we'd need is the commits squashed.

acornett21 (approved):

bcrochet (approved):
/lgtm

lib/fakes.go
sebrandon1: Works for me.

PR #810, opened 2022-10-24
Use constructor-style for NoopSubmitter

sebrandon1:
Inspired by #809 

This change can be separated from the rest of the `lib` work, but using a constructor-style to creating the `noopSubmitter` objects.

### DIFF
--- a/cmd/check_container_test.go
+++ b/cmd/check_container_test.go
@@ -158,7 +158,7 @@ certification_project_id: mycertid`
 			AfterEach(func() {
 				submit = origSubmitValue
 			})
-			It("should return a noopSubmitter ResultSubmitter", func() {
+			It("should return a NoopSubmitter ResultSubmitter", func() {
 				runner, err := lib.NewCheckContainerRunner(context.TODO(), cfg, false)
 				Expect(err).ToNot(HaveOccurred())
 				_, rsIsCorrectType := runner.Rs.(*lib.NoopSubmitter)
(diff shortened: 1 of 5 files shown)


acornett21:
I don't think the linter will allow us to switch to this style of coding since for current use cases we are never setting a `reason`, meaning the linter wants us to remove that parameters from the method signature completely.

cmd/types.go
```diff
@@ -184,7 +184,14 @@ type noopSubmitter struct {
 	log     *log.Logger
 }
 
-var _ resultSubmitter = &noopSubmitter{}
+func newNoopSubmitter(emitLog bool, log *log.Logger) *noopSubmitter {
+	return &noopSubmitter{
+		emitLog: emitLog,
+		log:     log,
```
acornett21, line 190: This gets past the linter issue, but I think we'd want to keep the `message` as an option, but I'll let @bcrochet or @komish also chime in.

sebrandon1: The `reason` field is still available in the struct.  It's just default to an empty string.

tkrishtop:
hi @sebrandon1, sorry for the failed job, we got an issue with the folder permissions in Dallas, but it is now fixed. Everything should work once your branch is rebased.

sebrandon1:
Fully rebased on top of the `lib` changes from yesterday.  Thanks for the reviews!

acornett21 (approved):

PR #813, opened 2022-10-25
move main.go to be within the cmd directory structure and push all subcommands down one layer

komish:
This PR is in preparation for exposing code at the top layer of the directory structure to be library code. We have to move main.go into `cmd` so that the import path can be correct for library code, and also probably should anyway as it doesn't hurt anything.


acornett21 (commented):
Can't `main` live in `/cmd` without having a sub folder? cmd/prflight/cmd seems like a weird package path.

acornett21:
> Can't `main` live in `/cmd` without having a sub folder? cmd/prflight/cmd seems like a weird package path.

Looking at a few other repos, it looks like this pattern is actually common.

komish:
> Can't `main` live in `/cmd` without having a sub folder? cmd/prflight/cmd seems like a weird package path.

I don’t think so. The package for the entry point must be main, but the cobra commands probably shouldn’t live in the main package.

tkrishtop:
hi @komish sorry for the failed job, we got an issue with the folder permissions in Dallas, it is now fixed

sebrandon1 (approved):

bcrochet (approved):
/lgtm

Awesome that the changes were pretty limited and just a 'git mv' in most cases.

skattoju4 (approved):
/lgtm

bcrochet (approved):
/lgtm

PR #819, opened 2022-11-02
Update Go to v1.19.3

sebrandon1:
https://groups.google.com/g/golang-announce/c/mbHY1UY3BaM?pli=1

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -23,9 +23,9 @@ Vagrant.configure("2") do |config|
     containers-common \
     openscap-containers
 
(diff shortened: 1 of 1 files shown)


acornett21 (commented):
Would we also want to change our Dockerfile as well to point to the latest golang builder image?

bcrochet (approved):
/lgtm

acornett21:
> Would we also want to change our Dockerfile as well to point to the latest golang builder image?

It looks like golang always points the major version to the latest minior version, so no change is needed in the docker file.

```
v1.19 -> v1.19.3 all point to the same sha
```

Issue #807, opened 2022-10-20
Support manifest list


lucacome:
## Is your feature request related to a problem? Please describe.
I want to run `preflight check container` once for my multi-arch manifest

## Describe the solution you'd like.
`preflight` certifies all the images in a manifest list.

## Describe alternatives you've considered.
Manually certifying every architecture

## Additional context.
Maybe I'm missing a flag or something? It seems like `preflight` is able to read the manifest list but defaults to using the image for `amd64`.

acornett21:
Hi @lucacome preflight uses a library `crane` to pull the images, get their configs, and send them to our backend, `crane` defaults to uses the architecture value for the system that it is run on. 

Currently, the workaround is what you mentioned, to certify every architecture individually on the appropriate base system.

With that said, multi-arch/manifest list is something that has been talked about being added to the container certification process. I'm not sure what the priority of that is, but it would not be just a change in preflight, backend systems at Red Hat would also need to be changed as well.

lucacome:
Hi @acornett21 thanks for the response and the explanation. 

>  ... crane defaults to uses the architecture value for the system that it is run on.

It seemed to default to `amd64` 🤔 

> Currently, the workaround is what you mentioned, to certify every architecture individually on the appropriate base system.

For the latest round of certifications I was able to do everything from one system, just specifying different SHAs...I hope everything worked 😅 

> but it would not be just a change in preflight, backend systems at Red Hat would also need to be changed as well.

Is there a better place for me to open an issue to ask for the backend systems to support this?

acornett21:
> It seemed to default to `amd64` 🤔

If you are on an `amd64` system using our `amd64` binary that would be the default, `arm64` would be `arm64`, and so on...

For your other questions I've messaged you on slack.

lucacome:
Sorry I forgot to mention that I'm on `arm64` and defaulted to `amd64`.

Thanks!

bcrochet:
> Sorry I forgot to mention that I'm on `arm64` and it defaulted to `amd64`.
> 
> Thanks!

I just tried on arm64, and you are correct. It pulled the amd64 image. I'll take a look at why that's happening.

lucacome:
> Hi @lucacome preflight uses a library `crane` to pull the images, get their configs, and send them to our backend, `crane` defaults to uses the architecture value for the system that it is run on.
> 
[...]

PR #821, opened 2022-11-03
Grab the image that matches the arch of the binary

bcrochet:
Crane defaults to "linux/amd64" as the platform, unless an option is passed with WithPlatform. This defaults to an OS of "linux", and an architecture of runtime.GOARCH, which is basically set at build time.

Additional work will be required to process all arches of an image in one go.

Related #807

### DIFF
--- a/certification/internal/engine/engine.go
+++ b/certification/internal/engine/engine.go
@@ -13,6 +13,7 @@ import (
 	"path"
 	"path/filepath"
 	"regexp"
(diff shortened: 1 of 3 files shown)


acornett21 (approved):

skattoju4 (approved):
/lgtm

skattoju (approved):
/lgtm

PR #825, opened 2022-11-08
publish library and refactor to use it

komish:
This PR publishes a `container` and `operator` package, allowing users to programmatically run Preflight in their code vs. calling out the the binary. Thanks (and cc) @sebrandon1 for kicking this effort off.

This takes a different direction than the original implementation PoC'd by @sebrandon1 , which used our existing `Check<Container/Operator>Runners`, which themselves internally called `PreflightCheck` with all the appropriate options. 

The reason for this dramatic of a change is because those implementations exhibited an issue where that call stack really intermixed logic required to execute checks with the logic required to run via the CLI (e.g. writing files to the filesystem, etc).

This PR should accomplish a separation of responsibilities, where the logic to execute preflight checks is decoupled from the logic to accept input from the user via CLI, write log files to the filesystem, etc. The result is that the CLI now is able to call the library itself.

This PR is large, so I'll try to add notes. to help reviewers get through it. I tried to partition it into component PRs, but all of the various parts of the refactor were dependent on each other. Apologies in advance.

Here are some major call outs:

## certification/artifacts

- This package was effectively a simplified way to write an artifact file to the filesystem. Our checks are written in such a way that they reach into this package to write a file at will. For library invocations, I didn't think it made sense to do this, so I've unplugged the global-ness of this package, and instead have checks reaching into the context required throughout execution to write an artifact. This global package import just did not fit our new approach.

- The artifacts package now has two new APIs, a FilesystemWriter and a MapWriter, that allow a caller to decide how their artifact files will be written. The CLI uses the FilesystemWriter, which should produce the same result as before.

### Side Effects
- a ton of our tests reached into Artifacts as well to configure the writing of artifacts to temp directories. As a result, I had to reconfigure a lot of tests to instead configure artifact writing structs and stuff them into the context. The result should be roughly the same (temp dirs receiving artifact files that are cleaned up).

- we treat log writing and artifact writing differently. Artifact writing was done by calling the artifact library itself, but log writing depends on logrus' global structs. This caused a problem for library callers, as they would not receive the preflight.log due to the logger being configured by the cmd package. As a result, there are helper functions in the internal/lib package that are intended to help bridge the gap until we consider logging from context as well (or some other solution). See **internal/lib/logging.go**

## runtime.Config && certification.Config

- These structs effectively help us translate from a viper config to the config we actually want to use in execution (e.g. to help inform check instantiation). The certification Config was also a read-only configuration to prevent a caller downstream from making a change to the configuration at runtime. It helped solve issues that seem less relevant today as we look to separate concerns between what our CLI invocation does, and the actual check execution. We also ended up with a global config struct that contained values for both operator and container checks, regardless of which one was being called.

- To that end, As we no longer have PreflightCheck, CheckContainerRunner, CheckOperatorRunner, etc, the key places where we need configuration values is at check instantiation. I've pulled out a majority of functions that required the full configuration in its signature, and instead focused in only on raw values necessary to produce the same output. So as an example, InitializeChecks now has been split to operator/container specific functions, and they each accept a smaller config struct containing the values they need to build out their checks.

- There's probably an opportunity in the future to remove these larger configs in favor of the smaller one when translating from viper to what we need to execute

## internal/cli

- The key thing to note here is that the previous PreflightCheck logic has mostly migrated here, albeit in a reduced form that wraps our library. The goal here was to further disambiguate between "tasks that need to happen to run via the CLI", and "tasks necessary to execute checks".

- Another key thing here is that we don't accept the engine here. Now we accept a function declaration that returns results or an error (i.e. the engine's Run function).  This lets us test a bit more flexible and seems to produce the same result.

## container, operator, and errors

- These are the library functions that we're making available. The goal was to provide as basic of an interface as we could. There is room for improvement (e.g. it would be better if these accepted io.Readers for things instead of filesystem paths), and I've tried to name things that can be improved in such a way that we can later deprecate them, or roll out the new feature without conflict.

- These are called internally by the CLI now 

- The errors package has been created just because this is an API and users may want to make decisions based on the errors they received. I know we've deleted an errors package/file in the past, so I tried not to be too overzealous with it. 

## Checks

- I mentioned this earlier, but there are several checks that were touched because they wrote artifacts. I adjusted their logic so they could write by pulling the artifact writer from context. I've also done this in a fail-safe way, which means if the context artifact writer is nil, we don't write. that's not ideal - but in some cases felt better than throwing an error and/or panicking.

## Final Thoughts

- I'll be looking at the coveralls report to see what I've missed. I had to shuffle a lot around, so it's possible I missed some coverage points we had previously.
- I'd love to test this extensively to make sure things aren't breaking in new and unknown ways. When released, we should release this as an alpha. 


komish:
[5bd0f64](https://github.com/redhat-openshift-ecosystem/openshift-preflight/commit/5bd0f64cf570784991c8503faaa3994fa5031e2f) fixes linting issues preventing **Go /build (pull_request)** from completing

certification/internal/operatorsdk/operatorsdk.go
```diff
@@ -163,9 +163,13 @@ func (o operatorSdk) BundleValidate(ctx context.Context, image string, opts Oper
 	return &bundleValidateData, nil
 }
 
-func (o operatorSdk) writeScorecardFile(resultFile, stdout string) error {
-	_, err := artifacts.WriteFile(resultFile, strings.NewReader(stdout))
-	return err
+func (o operatorSdk) writeScorecardFile(ctx context.Context, resultFile, stdout string) error {
```
sebrandon1, line 166: `o` is unused here.

komish:
[308be35](https://github.com/redhat-openshift-ecosystem/openshift-preflight/commit/308be358f0ad5c39bdc8cdc1aa3f304ec50472af) fixes a mistaken context used in a test that was causing **Go / build (pull_request)** to fail.

certification/internal/policy/operator/deployable_by_olm.go
```diff
@@ -65,6 +65,10 @@ func (p *DeployableByOlmCheck) initClient() error {
 	if err := openshift.AddSchemes(scheme); err != nil {
 		return fmt.Errorf("could not add new schemes to client: %w", err)
 	}
+
+	// TODO(): GetConfig generates a rest config from environment paths. We already have
```
sebrandon1, line 69: This might be what you are looking for: https://github.com/test-network-function/cnf-certification-test/pull/644

certification/internal/policy/operator/deployable_by_olm.go
```diff
@@ -575,7 +579,7 @@ func (p *DeployableByOlmCheck) cleanUp(ctx context.Context, operatorData operato
 	_ = p.openshiftClient.DeleteNamespace(ctx, operatorData.TargetNamespace)
 }
 
-func (p *DeployableByOlmCheck) writeToFile(data interface{}) error {
+func (p *DeployableByOlmCheck) writeToFile(ctx context.Context, data interface{}) error {
```
sebrandon1, line 582: I don't think `p` is used here.

sebrandon1 (changes requested):
Some initial comments, will re-review again tomorrow.

certification/engine/engine.go
```diff
@@ -30,33 +30,36 @@ type CheckEngine interface {
...
+		IsScratch:    isScratch,
+	}, nil
+}
 
-	return engine, nil
+// OperatorCheckConfig contains configuration relevant to an individual check's execution.
+type OperatorCheckConfig struct {
+	ScorecardImage, ScorecardWaitTime, ScorecardNamespace, ScorecardServiceAccount string
+	IndexImage, DockerConfig, Channel, Kubeconfig                                  string
 }
```
komish, line 53: [Note To Reviewers]: This is an example of the smaller task-focused configuration that's needed to initialize checks. This is what the checks need today, and we can modify this if checks change (as opposed to modifying the larger configs). This is possible because `InitializeChecks` has been replaced by individual policy-based initializing logic.

certification/internal/engine/engine.go
```diff
@@ -39,7 +39,9 @@ import (
 // CraneEngine implements a certification.CheckEngine, and leverage crane to interact with
 // the container registry and target image.
 type CraneEngine struct {
-	Config certification.Config
+	// DockerConfig is the credential required to pull the image.
+	DockerConfig string
+	// Config certification.Config
```
komish, line 44: [Note To Reviewers]: This is an example of a place where we passed the full-sized configuration vs. specific components we needed previously, and this PR changes that.

certification/internal/engine/engine.go
```diff
@@ -302,9 +300,12 @@ func generateBundleHash(bundlePath string) (string, error) {
...
-	_, err := artifacts.WriteFile("hashes.txt", &hashBuffer)
-	if err != nil {
-		return "", fmt.Errorf("could not write hash file to artifacts dir: %w", err)
+	artifactsWriter := artifacts.WriterFromContext(ctx)
+	if artifactsWriter != nil {
+		_, err := artifactsWriter.WriteFile("hashes.txt", &hashBuffer)
+		if err != nil {
+			return "", fmt.Errorf("could not write hash file to artifacts dir: %w", err)
+		}
 	}
```
komish, line 309: [Note To Reviewers]: Here's an example of the new way to write artifacts. This function did not have a context, so it needed to be added in order to implement the new approach.

certification/internal/engine/engine_test.go
```diff
@@ -46,7 +47,9 @@ var _ = Describe("Execute Checks tests", func() {
 		tmpDir, err := os.MkdirTemp("", "preflight-engine-test-*")
 		Expect(err).ToNot(HaveOccurred())
 		DeferCleanup(os.RemoveAll, tmpDir)
-		artifacts.SetDir(tmpDir)
+		aw, err := artifacts.NewFilesystemWriter(artifacts.WithDirectory(tmpDir))
+		Expect(err).ToNot(HaveOccurred())
+		testcontext = artifacts.ContextWithWriter(context.Background(), aw)
```
komish, line 52: [Note To Reviewers]: Here's an example of what needed to change in tests (throughout). We still write to the filesystem where required and clean up after ourselves, but we facilitate the writes through the context. Some test also needed the previous `artifacts.Path()` logic so in some cases the artifact writer is a test-block-wide variable that can be called for this, as the helper functions that pull the writer from the context only return the interface.

certification/internal/operatorsdk/operatorsdk.go
komish: Think we should move this as a function vs. a method, or something else?

certification/internal/policy/operator/scorecard_basic_check_spec.go
```diff
@@ -69,8 +68,7 @@ func (p *ScorecardBasicSpecCheck) Help() certification.HelpText {
 		}
 	}
 	return certification.HelpText{
-		Message: "Check ScorecardBasicSpecCheck encountered an error. Please review the " +
-			artifacts.Path() + "/" + scorecardBasicCheckResult + " file for more information.",
+		Message:    "Check ScorecardBasicSpecCheck encountered an error. Please review the " + scorecardBasicCheckResult + " file in your execution artifacts for more information.",
```
komish, line 71: These messages referenced a filesystem path in the past. I'm not sure this makes sense anymore.

komish:
[51df40f](https://github.com/redhat-openshift-ecosystem/openshift-preflight/commit/51df40fe113c71109b1ec967462a1886872d5775) just removes an extra word in the code that I forgot to delete.

internal/lib/logging.go
```diff
@@ -0,0 +1,47 @@
...
+		return false
+	}
+}
+
+// SetCallerToCLI sets the caller as the CLI. NOTE: This is a temporary
+// workaround for internal CLI executions and will be removed at a later
+// date.
+func SetCallerToCLI(ctx context.Context) context.Context {
+	return context.WithValue(ctx, executionEnvIsCLI, true)
+}
```
komish, line 47: [Note To Reviewers]: This file is what lets us include some logic in the Library implementation of operator/container checks that allows us to work around our log configuration issues, where the CLI configures the logger early in the execution process.

certification/artifacts/filesystem_writer.go
```diff
@@ -0,0 +1,58 @@
...
+
+// FilesystemWriter is an ArtifactWriter that targets a particular directory on
+// the underlying filesystem.
+type FilesystemWriter struct {
+	dir string
+	fs  afero.Fs
+}
+
+// NewFilesystemWriter creates an artifact writer which writes to the filesystem.
+func NewFilesystemWriter(opts ...FilesystemWriterOption) (*FilesystemWriter, error) {
```
sebrandon1, line 19:
```suggestion
func NewFilesystemWriter(opts ...FilesystemWriterOption) *FilesystemWriter {
```

Error is never returned.

certification/artifacts/map_writer.go
```diff
@@ -0,0 +1,34 @@
...
+
+var ErrFileAlreadyExists = errors.New("file already exists")
+
+// MapWriter implements an ArtifactWriter storing contents in a map.
+type MapWriter struct {
+	files map[string]io.Reader
+}
+
+// NewMapWriter creates an artifact writer in memory using a map.
+func NewMapWriter() (*MapWriter, error) {
```
sebrandon1, line 16:
```suggestion
func NewMapWriter() *MapWriter {
```
Error never returned.

certification/internal/operatorsdk/operatorsdk.go
sebrandon1: Probably just a normal function.

sebrandon1 (commented):
Some more comments

certification/artifacts/filesystem_writer.go
komish: I'm open to the idea, but if these should need to do something that might return an error in the future, it's easier to return nil now than it is to add an error return value to the signature later.

sebrandon1: Ah I see, okay I figured it was an oversight.  This works for me.

operator/check_operator.go
```diff
@@ -0,0 +1,157 @@
...
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/internal/lib"
+)
+
+type Option = func(*operatorCheck)
+
+// TODO(): replace this value when the default in package cmd is moved to a central location
+const defaultScorecardWaitTime = "240"
+
+// NewCheck is a check runner that executes the Operator Policy.
+func NewCheck(image, kubeconfig, indeximage string, opts ...Option) *operatorCheck {
```
komish, line 21: [Feedback Request]: This signature contains the kubeconfig and is expecting a file. Ideally, we would want to accept an io.Reader but there are other changes that need to happen here before we do that. I'm taking suggestions about what to do here so we don't have to change this signature once we do make the changes necessary to use an io.Reader and cause potential breakages when consumers upgrade versions of this library.

I could accept the Kubeconfig as an option instead, and differentiate similarly to how I'm currently doing the DockerConfigJSON, but a Kubeconfig isn't optional.

certification/internal/engine/engine.go
acornett21: should like 44 be removed? it is commented out correct?

certification/internal/engine/engine.go
acornett21: Do we need to worry about `artifactsWriter` being nil in any case? I think the intent here is for when callers do not care about what we write to disk, but what about the cases where we are trying to write to disk, would an exception happen before we even get there?

certification/internal/openshift/openshift_version.go
```diff
@@ -17,7 +17,7 @@ func GetOpenshiftClusterVersion() (runtime.OpenshiftClusterVersion, error) {
 	if _, ok := os.LookupEnv("KUBECONFIG"); !ok {
 		return runtime.UnknownOpenshiftClusterVersion(), fmt.Errorf("KUBECONFIG not specified")
 	}
-	kubeConfig, err := ctrl.GetConfig()
+	kubeConfig, err := ctrl.GetConfig() // TODO(): Identify how to get the config/rest client using an io.Reader or similar instead of this.
```
acornett21, line 20: Is this todo for a future pr?

certification/internal/operatorsdk/operatorsdk.go
acornett21: If we move this to a normal function, I'd probably push it down past `createScorecardConfigFile` just for clarity and to avoid confusion as to why all the other functions have `o` and this does not.

certification/internal/policy/operator/scorecard_basic_check_spec.go
acornett21: Is `artificats.Path()` a thing still? I'm guessing the answer is in some cases yes, in some cases no....so it's probably safe to remove this to avoid `nil` pointer panics, even if the log is less clear now for some use cases.

internal/lib/lib.go
```diff
@@ -1,112 +1,19 @@
...
-		Cfg:       cfg,
-		Eng:       engine,
-		Formatter: fmttr,
-		Rw:        &runtime.ResultWriterFile{},
-	}, nil
-}
-
-// ResolveSubmitter will build out a ResultSubmitter if the provided pyxisClient, pc, is not nil.
-// The pyxisClient is a required component of the submitter. If pc is nil, then a noop submitter
-// is returned instead, which does nothing.
```
acornett21, line 102: Should we keep this comment?

certification/internal/openshift/openshift_version.go
komish: Yep. I didn't want to include it here just because it changed even more things than I'm already changing.

certification/internal/engine/engine.go
komish: Yep, I'll remove it. Good catch

certification/internal/openshift/openshift_version.go
acornett21: That makes sense, and I see the larger TODO where you call this out explicitly.

certification/internal/policy/operator/scorecard_basic_check_spec.go
komish: Technically No.

`artifacts.Path()` was a function in the Artifacts package that would return the singleton's configured directory (previously set by artifacts.SetDir()`. 

The FilesystemWriter does have a `Path()` method, and that's what we use in CLI invocations. This result, however, is returned to the user regardless of their context (e.g. library calls, cli exec, whatever).

Checks have access to an ArtifactWriter interface via the context, and that interface only defines the `WriteFile` method. I initially had the `Path` method defined to solve this use case, but other artifact writers don't necessarily have this concept of a path, so I did away with it here to make sure that our results are caller agnostic.

Hope that makes sense.

certification/internal/engine/engine.go
komish: Yep, you're spot on. The `preflight` CLI execution _needs_ the writer to be there, so it throws an error if one is not found before we ever hit this logic. See --> https://github.com/redhat-openshift-ecosystem/openshift-preflight/blob/51df40fe113c71109b1ec967462a1886872d5775/internal/cli/cli.go#L36-L40

For library callers, in theory, it's optional. I say "in theory", because it's almost certainly something that the library caller will want to add to the context to get more information for cases where things are breaking, but that's still up to the caller.

It's designed to support the case where the caller didn't provide an artifact writer gracefully. Our CLI execution enforces the artifact Writer must be in the context before it hits this.

certification/artifacts/artifacts_test.go
```diff
@@ -1,41 +1,30 @@
-package artifacts
+package artifacts_test
```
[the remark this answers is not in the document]
komish: It can be, but I made it `artifacts_test` because the tests I executed don't require internal knowledge of the workings of the package. This is the Go implementation of "black box testing".

certification/engine/engine.go
```diff
@@ -30,33 +30,36 @@ type CheckEngine interface {
...
 
-// initializeChecks configures checks for a given policy p using cfg as needed.
-//
-//nolint:unparam // ctx is unused. Keep for future use.
-func initializeChecks(ctx context.Context, p policy.Policy, cfg certification.Config) ([]certification.Check, error) {
+// InitializeOperatorChecks returns opeartor checks for policy p give cfg.
+func InitializeOperatorChecks(ctx context.Context, p policy.Policy, cfg OperatorCheckConfig) ([]certification.Check, error) {
```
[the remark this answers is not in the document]
komish: I'll wait for others to chime in to see opinions. For what it's worth, a consumer doesn't need to initialize an engine or checks to call preflight check operator or check container via a library. That's what _we_ offer in the container and operator packages.

internal/lib/lib.go
komish: Yep, sorry - I moved the old functions into different places so I could use the same function names and migrate to them. Some comments got caught in the crossfire.

komish:
[def2e43](https://github.com/redhat-openshift-ecosystem/openshift-preflight/commit/def2e43816fbc1ffaf1abb316b144858de5b6902) Resolves two code review comments from @acornett21 re: comments I left/took out inadvertently. No code changes here.

certification/internal/engine/engine.go
acornett21: That makes sense that in the CLI package we have an error if `artifactsWriter` is nil, this is the code I was looking for, thanks for pointing to it!.

certification/internal/policy/operator/scorecard_basic_check_spec.go
acornett21: That makes perfect sense.

komish:
[3ac0654](https://github.com/redhat-openshift-ecosystem/openshift-preflight/commit/3ac0654e02bc06d7db93094c324760b2b003d65b) rebases to the commit by @bcrochet [to add platform setting support](https://github.com/redhat-openshift-ecosystem/openshift-preflight/commit/f24ae18453eead8cb4b6da295e369eb8d4c2eb77) and adds the appropriate library options for the feature.

bcrochet (commented):
Not going to put 'request changes' yet.

komish:
[42091d0](https://github.com/redhat-openshift-ecosystem/openshift-preflight/commit/42091d0c14c52421949368b53a8d21ec71e9bd1a) rebases on @bcrochet's PR that adds the Insecure flag, and adds the functionality to the new library-based workflow.

certification/internal/openshift/openshift_version.go
komish: Resolving this.

certification/artifacts/artifacts.go
```diff
@@ -5,60 +5,33 @@
...
-	return fullFilePath, nil
-}
+const artifactWriterContextKey contextKey = "ArtifactWriter"
 
-// Path will return the artifacts directory.
-func Path() string {
-	return ads
+// ArtifactWriter is the functionality required by all implementations.
+type ArtifactWriter interface {
+	WriteFile(filename string, contents io.Reader) (string, error)
```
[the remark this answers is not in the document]
komish: Fair. Oversight on my part. I suppose technically this is subjective to the implementation, but the spirit of it is that the returned value is the fullpath. I've added return parameter identifiers here.

komish:
[c5d9ee5](https://github.com/redhat-openshift-ecosystem/openshift-preflight/commit/c5d9ee50cb8496bb7239410651f7dfa8bd681d90) adds a hint into the ArtifactWriter interface's string return value to inform the caller as to what it is intended to be.

certification/artifacts/filesystem_writer.go
```diff
@@ -0,0 +1,58 @@
...
+func WithDirectory(dir string) FilesystemWriterOption {
+	return func(w *FilesystemWriter) {
+		if dir == "" {
+			return
+		}
+		w.dir = resolveFullPath(dir)
+	}
+}
+
+type FilesystemWriterOption = func(*FilesystemWriter)
```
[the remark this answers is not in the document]
komish: Discussed this with you because I could go either way. I'm going to go with no error here, simply because I don't want to inadvertently encourage the possibility of doing option evaluations here, and returning an error if there are two exclusive options. Option ordering would cause this to be flaky.

To your point, the intended use of an error in an option is for things that are invalid (nil values to interface parameters, etc). That **IS** a valid use case that will need to be handled prior internally before any logic is called. You have an example of this in another project, so having no error returned here will serve as a bit of an A/B test.

certification/artifacts/filesystem_writer.go
```diff
@@ -0,0 +1,58 @@
...
+type FilesystemWriter struct {
+	dir string
+	fs  afero.Fs
+}
+
+// NewFilesystemWriter creates an artifact writer which writes to the filesystem.
+func NewFilesystemWriter(opts ...FilesystemWriterOption) (*FilesystemWriter, error) {
+	w := FilesystemWriter{
+		dir: resolveFullPath(DefaultArtifactsDir),
+		fs:  afero.NewOsFs(),
```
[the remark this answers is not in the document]
komish: I didn't have a use case for it, but I left the struct field this way just in case. Didn't know what the future held for our Afero use, and I initially PoC'd a MemMapFS ArtifactWriter using this logic that didn't quite work out.

certification/internal/engine/engine.go
komish: @bcrochet for some reason I have to reply to your 7LoC question in this thread. 

In theory, yep, as long as we want to do a "WriteIfNotNil". I imagined this would be a library. I should have included it here, but didn't. I could do that here if we feel strongly about it, or in a future PR if we're open to that.

komish:
[4e5e6a2](https://github.com/redhat-openshift-ecosystem/openshift-preflight/commit/4e5e6a205197e95367e249a734eb81656b776e8d) adds **docs/LIBRARY.md** which demos the usage of the library.

cmd/preflight/cmd/check_container.go
```diff
@@ -74,23 +77,41 @@ func checkContainerRunE(cmd *cobra.Command, args []string) error {
...
 	}
 
+	opts := generateContainerCheckOptions(cfg)
+
+	checkcontainer := container.NewCheck(
+		containerImage,
+		opts...,
+	)
+
+	pc := lib.NewPyxisClient(ctx, cfg.CertificationProjectID, cfg.PyxisAPIToken, cfg.PyxisHost)
```
acornett21, line 100: The order of this creates a bug where we can no longer test in lower envs. NewCheck can either take `pc` or at the very least `cfg.PyxisHost` and have NewCheck no longer default to the default pyxis host

komish: Oh! I think the problem here is that I never actually pass the values to the containerCheck. I believe I already added the options, I just didn't call them. I'll double check. Good catch

bcrochet (commented):
Couple more comments.

komish:
[58662ff](https://github.com/redhat-openshift-ecosystem/openshift-preflight/commit/58662fff20d3a382444b329649342bfc0564be80) @acornett21 this should fix the issue where I wasn't passing pyxis and certification information from the viper configuration to the container check options.

errors/errors.go
```diff
@@ -0,0 +1,12 @@
+package errors
+
+import "errors"
+
+// Library-wide error messages are here.
+var (
+	ErrKubeconfigEmpty              = errors.New("kubeconfig value is empty")
```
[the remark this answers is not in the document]
komish: Yep. I'd be on board with a good motivation factor to _not_ include them. But I figured we have to emit _something_ from the library functions?

container/check_container.go
```diff
@@ -0,0 +1,159 @@
...
+	if err := eng.ExecuteChecks(ctx); err != nil {
+		return runtime.Results{}, err
+	}
+
+	return eng.Results(ctx), nil
+}
+
+// HasPyxisData returns true of the values necessary to make a pyxis
+// API call are not empty. This does not check the validity of the input values.
+func (c *containerCheck) HasPyxisData() bool {
```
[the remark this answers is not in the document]
komish: Excellent question. I've unexported it in [3d6498f](https://github.com/redhat-openshift-ecosystem/openshift-preflight/commit/3d6498ff1eb226df787c1f3dec1819b5a87b643e)

komish:
[3d6498f](https://github.com/redhat-openshift-ecosystem/openshift-preflight/commit/3d6498ff1eb226df787c1f3dec1819b5a87b643e) Unexports `containerCheck.HasPyxisData`

cmd/preflight/cmd/check_container.go
komish: Should be resolved. I'll let you resolve this conversation when confirmed.

komish:
[2dc22ae](https://github.com/redhat-openshift-ecosystem/openshift-preflight/commit/2dc22aee44c0708c57f501ec1d6dcd97d1ccb8c5) Just fixes a typo in docs/LIBRARY.md ([3] --> [4])

komish:
[2810591](https://github.com/redhat-openshift-ecosystem/openshift-preflight/commit/2810591d82764021543e2b0f8c9aacd966d7b7a1) Sets logrus to discard log writes to the standard logger if there is no artifact writer configured in the context. This is only called when not being invoked by the CLI and fixes a bug where simply calling either container or operator check functions with no artifact writer configured would write to stderr.

certification/artifacts/artifacts.go
```diff
@@ -5,60 +5,33 @@
...
-	return fullFilePath, nil
-}
+const artifactWriterContextKey contextKey = "ArtifactWriter"
 
-// Path will return the artifacts directory.
-func Path() string {
-	return ads
+// ArtifactWriter is the functionality required by all implementations.
+type ArtifactWriter interface {
+	WriteFile(filename string, contents io.Reader) (fullpathToFile string, err error)
```
bcrochet, line 36: Great! Now the interface is clear.

cmd/preflight/cmd/check_container.go
acornett21: I retested this on Friday and it is fixed. Thanks!

errors/errors.go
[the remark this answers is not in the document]
acornett21: @komish Even if we keep these internal, they are emitted, right? or did you mean something else?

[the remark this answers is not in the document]
komish: Sure, that's fair. In practice, however, if a user is calling our API, receives an error, and wants to perform some action based on that error, these errors would provide the caller something to match against.

bcrochet (commented):
A few more comments.

certification/artifacts/filesystem_writer_test.go
```diff
@@ -0,0 +1,50 @@
...
+	"path/filepath"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Filesystem Artifact Writer", func() {
+	var tempdir string
+	BeforeEach(func() {
+		var err error
```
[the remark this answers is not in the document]
komish: Not sure I follow this. Even if I created a new aferoFS and just used the same FilesystemWriter, I'd still have to create the directories in a BeforeEach, right?

certification/artifacts/artifacts_test.go
```diff
@@ -1,41 +1,30 @@
...
-			Expect(err).ToNot(HaveOccurred())
-			Expect(dirInfo.IsDir()).To(BeTrue())
+			ctx = artifacts.ContextWithWriter(ctx, aw)
+			awRetrieved := artifacts.WriterFromContext(ctx)
+			Expect(awRetrieved).ToNot(BeNil())
+			Expect(awRetrieved).To(BeEquivalentTo(aw))
 		})
 	})
+	It("Should return nil when there is no ArtifactWriter found in the context", func() {
+		ctx := context.Background()
```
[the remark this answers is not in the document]
komish: Making it one line means I no longer need the variable here 👍

komish:
[46bfc95](https://github.com/redhat-openshift-ecosystem/openshift-preflight/commit/46bfc95ddc7166e593806929bf04cac71fd717cf) This addresses artifacts writer test feedback by @bcrochet

operator/check_operator.go
bcrochet: Let's iterate on this in the branch.

bcrochet (approved):
/lgtm

PR #834, opened 2022-11-16
add support for kubeconfig being pulled in from an io.Reader and stored as a []byte

komish:
This PR converts all logic that relies on the existence of the `KUBECONFIG` environment variable to instead rely on either the raw data (as `[]byte` or an `io.Reader` depending on the case.

### DIFF
--- a/certification/internal/engine/engine.go
+++ b/certification/internal/engine/engine.go
@@ -38,6 +38,8 @@ import (
 // CraneEngine implements a certification.CheckEngine, and leverage crane to interact with
 // the container registry and target image.
 type CraneEngine struct {
+	// Kubeconfig is a byte slice containing a valid Kubeconfig to be used by checks.
+	Kubeconfig []byte
 	// DockerConfig is the credential required to pull the image.
 	DockerConfig string
 	// Image is what is being tested, and should contain the
@@ -175,7 +177,7 @@ func (c *CraneEngine) ExecuteChecks(ctx context.Context) error {
 
 	if c.IsBundle {
 		// Record test cluster version
-		c.results.TestedOn, err = openshift.GetOpenshiftClusterVersion()
+		c.results.TestedOn, err = openshift.GetOpenshiftClusterVersion(c.Kubeconfig)
 		if err != nil {
 			log.Errorf("could not determine test cluster version: %v", err)
 		}
--- a/certification/internal/openshift/openshift_version.go
+++ b/certification/internal/openshift/openshift_version.go
@@ -3,25 +3,26 @@ package openshift
 import (
 	"context"
 	"fmt"
-	"os"
 
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification/runtime"
 
 	configv1Client "github.com/openshift/client-go/config/clientset/versioned/typed/config/v1"
 	log "github.com/sirupsen/logrus"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	ctrl "sigs.k8s.io/controller-runtime"
+	"k8s.io/client-go/tools/clientcmd"
 )
 
-func GetOpenshiftClusterVersion() (runtime.OpenshiftClusterVersion, error) {
-	if _, ok := os.LookupEnv("KUBECONFIG"); !ok {
-		return runtime.UnknownOpenshiftClusterVersion(), fmt.Errorf("KUBECONFIG not specified")
+func GetOpenshiftClusterVersion(kubeconfig []byte) (runtime.OpenshiftClusterVersion, error) {
+	if len(kubeconfig) == 0 {
+		return runtime.UnknownOpenshiftClusterVersion(), fmt.Errorf("kubeconfig was not provided")
 	}
-	kubeConfig, err := ctrl.GetConfig() // TODO(): Identify how to get the config/rest client using an io.Reader or similar instead of this.
+
+	restconfig, err := clientcmd.RESTConfigFromKubeConfig(kubeconfig)
 	if err != nil {
 		return runtime.UnknownOpenshiftClusterVersion(), fmt.Errorf("unable to load the config, check if KUBECONFIG is set correctly: %v", err)
 	}
-	configV1Client, err := configv1Client.NewForConfig(kubeConfig)
+
+	configV1Client, err := configv1Client.NewForConfig(restconfig)
 	if err != nil {
 		return runtime.UnknownOpenshiftClusterVersion(), fmt.Errorf("unable to create a client with the provided kubeconfig: %v", err)
 	}
@@ -30,7 +31,7 @@ func GetOpenshiftClusterVersion() (runtime.OpenshiftClusterVersion, error) {
 		return runtime.UnknownOpenshiftClusterVersion(), fmt.Errorf("unable to get openshift-apiserver cluster operator: %v", err)
 	}
 
-	log.Debug(fmt.Sprintf("fetching operator version and openshift-apiserver version %s from %s", openshiftAPIServer.Status.Versions, kubeConfig.Host))
+	log.Debug(fmt.Sprintf("fetching operator version and openshift-apiserver version %s from %s", openshiftAPIServer.Status.Versions, restconfig.Host))
 	return runtime.OpenshiftClusterVersion{
 		Name:    "OpenShift",
 		Version: openshiftAPIServer.Status.Versions[1].Version,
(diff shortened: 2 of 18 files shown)


operator/check_operator.go
```diff
@@ -19,7 +19,7 @@ type Option = func(*operatorCheck)
 const defaultScorecardWaitTime = "240"
 
 // NewCheck is a check runner that executes the Operator Policy.
-func NewCheck(image, kubeconfig, indeximage string, opts ...Option) *operatorCheck {
+func NewCheck(image string, kubeconfig io.Reader, indeximage string, opts ...Option) *operatorCheck {
```
acornett21, line 22: Nit: Do we also want to re-order the params to make it more readable? 
```suggestion
func NewCheck(image,  indeximage string, kubeconfig io.Reader, opts ...Option) *operatorCheck {
```

acornett21:
How does someone now run this? How would this work in the pipeline(s)?

operator/check_operator.go
komish: Hmm I'm open to this. I figured "order of importance", ranking Kubeconfig > Index Image - but I mean... they're both required so maybe that's silly of me 😆

certification/internal/openshift/openshift_version.go
```diff
@@ -3,25 +3,31 @@ package openshift
...
-	ctrl "sigs.k8s.io/controller-runtime"
+	"k8s.io/client-go/tools/clientcmd"
 )
 
-func GetOpenshiftClusterVersion() (runtime.OpenshiftClusterVersion, error) {
-	if _, ok := os.LookupEnv("KUBECONFIG"); !ok {
-		return runtime.UnknownOpenshiftClusterVersion(), fmt.Errorf("KUBECONFIG not specified")
+func GetOpenshiftClusterVersion(kubeconfig []byte) (runtime.OpenshiftClusterVersion, error) {
+	if len(kubeconfig) == 0 {
+		return runtime.UnknownOpenshiftClusterVersion(), fmt.Errorf("kubeconfig was not received")
```
skattoju, line 17: nit kubeconfig was not provided

komish:
> How does someone now run this? How would this work in the pipeline(s)?

That's a great question. Nothing should change. Users invoking the preflight binary (e.g. the pipeline) still need to pass KUBECONFIG=/my/path/to/kubeconfig.

Internally, the CLI invocation will take that path and turn it into an what's needed by operator.NewCheck (e.g. io.Reader).

The internal engines turn this reader into a byte slice, and consume that as needed. 

So for ClusterVersion checks, we pass the byte slice straight into client-go and end up with the rest config needed by the cluster calls we make.

For Scorecard, we (unfortunately) _need_ a file on disk because we exec operator-sdk. Instead of building support in the engines for both a file on disk AND the byte slice, I just rewrite the byte slice to disk in a temp file. This happens for each buildout of our scorecard structs, so that means twice in our case. It's not particularly efficient, but I thought it felt better than passing down the kubeconfig file path down for one use case, and using the byte slice for the other. In effect, I favored a consistent passing of data to the engine, and let the individual engine components manipulate that to their use case.

certification/internal/engine/engine.go
```diff
@@ -38,6 +38,8 @@ import (
 // CraneEngine implements a certification.CheckEngine, and leverage crane to interact with
 // the container registry and target image.
 type CraneEngine struct {
+	// Kubeconfig is a byte slice containing a valid Kubeconfig to be used ny checks.
```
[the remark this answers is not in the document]
komish: Fixed

certification/internal/openshift/openshift_version.go
komish: Fixed

certification/internal/openshift/openshift_version.go
```diff
@@ -3,25 +3,31 @@ package openshift
...
-func GetOpenshiftClusterVersion() (runtime.OpenshiftClusterVersion, error) {
-	if _, ok := os.LookupEnv("KUBECONFIG"); !ok {
-		return runtime.UnknownOpenshiftClusterVersion(), fmt.Errorf("KUBECONFIG not specified")
+func GetOpenshiftClusterVersion(kubeconfig []byte) (runtime.OpenshiftClusterVersion, error) {
+	if len(kubeconfig) == 0 {
+		return runtime.UnknownOpenshiftClusterVersion(), fmt.Errorf("kubeconfig was not provided")
 	}
-	kubeConfig, err := ctrl.GetConfig() // TODO(): Identify how to get the config/rest client using an io.Reader or similar instead of this.
+
+	cc, err := clientcmd.NewClientConfigFromBytes(kubeconfig)
```
bcrochet, line 20: Can you use this instead, and not have to do the next block? 
https://github.com/kubernetes/client-go/blob/master/tools/clientcmd/client_config.go#L120

It looks like it literally does both of these.

cmd/preflight/cmd/check_operator.go
```diff
@@ -93,7 +93,12 @@ func checkOperatorRunE(cmd *cobra.Command, args []string) error {
 
 	opts := generateOperatorCheckOptions(cfg)
 
-	checkoperator := operator.NewCheck(operatorImage, cfg.Kubeconfig, cfg.IndexImage, opts...)
+	kubeconfig, err := os.Open(cfg.Kubeconfig)
+	if err != nil {
+		return fmt.Errorf("unable to open provided kubeconfig file: %s", err)
+	}
+	defer kubeconfig.Close()
```
bcrochet, line 100: Do we really want to defer closing this for the entire run? Kind of a nit. Up to you.
Kind of goes with the point below about just passing the byte slice instead of an io.Reader.

operator/check_operator.go
bcrochet: I don't think it matters if they're both required. :) I'd opt for readability.

operator/check_operator.go
```diff
@@ -52,9 +52,9 @@ func (c operatorCheck) Run(ctx context.Context) (runtime.Results, error) {
 		lib.LogThroughArtifactWriterIfSet(ctx)
 	}
 
-	// NOTE(from Jose): Workaround for DeployableByOLM which relies on ctrl.GetConfig()
-	if _, isSet := os.LookupEnv("KUBECONFIG"); !isSet {
-		os.Setenv("KUBECONFIG", c.kubeconfig)
+	kubeconfig, err := io.ReadAll(c.kubeconfig)
```
bcrochet, line 55: Is there any reason to delay this read to here? What do we gain by passing the io.Reader instead of just the byte slice? I think if it's ultimately just going to be the byte slice, passing the io.Reader seems an unnecessary abstraction. It doesn't change the testability.

bcrochet (commented):
A few comments. Not going to mark as request changes. Most of them are discussion points.

certification/internal/openshift/openshift_version.go
komish: Sure thing. The clientcmd package has way too many ways to do what feels like "the same thing". Totally missed this convenience function.

komish: Fixed

operator/check_operator.go
komish: Order changed!

operator/check_operator.go
komish: Fair. I've changed it to accept just the byte slice instead.

cmd/preflight/cmd/check_operator.go
komish: This open is unrelated - the CLI invocation _always_ has to open a file. That said, might as well close it as soon as we're done with it. I've removed the `defer` and closed it explicitly.

cmd/preflight/cmd/check_operator.go
```diff
@@ -93,7 +94,18 @@ func checkOperatorRunE(cmd *cobra.Command, args []string) error {
 
 	opts := generateOperatorCheckOptions(cfg)
 
-	checkoperator := operator.NewCheck(operatorImage, cfg.Kubeconfig, cfg.IndexImage, opts...)
+	kubeconfigFile, err := os.Open(cfg.Kubeconfig)
```
bcrochet, line 97: So, you can still use a defer if you put this all in an anonymous function.

```suggestion
        kubeConfig, err := func(kubeconfig string) ([]byte, err) {
            kubeconfigFile, err := os.Open(cfg.Kubeconfig)
            defer kubeconfigFile.Close()
            if err != nil { return nil, err }
            kubeconfig, err := io.ReadAll(kubeconfigFile)
            if err != nil { return nil, err }
            return kubeconfig, nil
        }
```

komish: Fixed

bcrochet (approved):
/lgtm

skattoju (approved):
/lgtm

PR #835, opened 2022-11-17
Rebase lib on main

komish:
This rebases the lib branch on main which has 4 dependabot commits merged that didn't make it to lib. I'm not sure this is necessary, but figured I'd try it anyway.


acornett21:
> These go.mod conflicts are kind of odd. I don't have these conflicts locally. 🤷

These are the two dependencies that I had to manually change to get the project to build, since they needed to be done at one time, and the bot was only doing one at a time. Then the bot incremented the versions with the next release. I get the same merge conflict locally. LMK if you need help.

bcrochet:
Are we sure this works like this? I haven't really done a rebase like this in GitHub. I would cherry-pick the dependabot commits individually. Or, I would just ignore them until we are ready to actually rebase lib ONTO main to then be commited to main.

komish:
I'm fairly sure if I cherry-pick the dependabot commits onto lib, it'll mess with order of those commits in the history and then rebasing will be more difficult. Try and and shoot a PR out here.

Since rebase seems apparently finicky in this case, which almost certainly means we're doing something wrong here, I'm thinking we could just take main, cherry-pick the lib commits over the top of the latest main, then deal with conflicts (probably minimal).

PR #838, opened 2022-11-18
Move even more from certification to internal

bcrochet:
This starts to move some of the "public" things in certification intor internal, as some of these never really needed to be part of the public API, but were only made so because of how internal is handled.

* Moves certification/engine
* Moves certification/formatters
* Moves certification/pyxis
* Moves certification/runtime
* Moves certification/policy


komish:
How does a user leverage our formatters to take results from library calls and write them wherever they want?

komish:
If you want to keep formatters out of the exported surface, you'll need to update the LIBRARY doc https://github.com/redhat-openshift-ecosystem/openshift-preflight/blob/lib/docs/LIBRARY.md.

komish:
From what I can tell, library callers cannot write logic around runtime.Results now because the type is defined in the internal package.

bcrochet:
> How does a user leverage our formatters to take results from library calls and write them wherever they want?

Assume you mean a user of the library. As it stands, the formatters, to me, are a preflight implementation detail. If someone requests that we expose the formatters, then we cross that bridge.

bcrochet:
> From what I can tell, library callers cannot write logic around runtime.Results now because the type is defined in the internal package.

They would still get a struct back. They just could not create a runtime.Results directly.

komish:
They also cannot create functions with runtime.Results as a parameter.

bcrochet:
> They also cannot create functions with runtime.Results as a parameter.

That one I don't have an answer for. Is that necessary?

bcrochet:
> Sure. Having a struct of data that I want to manipulate becomes a lot harder when the struct's type isn't exported.

Ok. Since it's in the signature of *Check.Run, I can see exposing that.

bcrochet:
Now in #840

PR #839, opened 2022-11-18
Move the last package out of certification

bcrochet:
Only one package remained in certification. There can be only none.
    
* Moves artifacts to internal
    

komish:
How would a library caller retrieve their artifacts if the artifact writer definitions are all internal

bcrochet:
> How would a library caller retrieve their artifacts if the artifact writer definitions are all internal

Will address in #840

PR #840, opened 2022-11-18
Complete more of the move from certification to internal

bcrochet:
* Moves certification/engine
* Moves certification/formatters
* Leaves the FormatterFunc type exported
* Moves certification/pyxis
* Moves certification/runtime
* Moves certification/policy
* Creates runtime top-level package    


komish:
As a library caller, this PR would not allow me to extract artifacts, such as the preflight.log, from my execution. The artifacts package needs to be made available so that a library caller can manipulate the context with the appropriate ArtifactWriter.

bcrochet:
> How would a library caller retrieve their artifacts if the artifact writer definitions are all internal

Should library users with to retrieve the artifacts, I would propose that a (future) API could be added to reach internal and get a tar to get the artifacts.

komish:
> Should library users wish to retrieve the artifacts, I would propose that a (future) API could be added to reach internal and get a tar to get the artifacts. I currently view artifacts as a preflight-specific implementation detail, and not a general purpose thing for library users.

This doesn't work for me. If a library caller gets an error or failing result, (i.e. they don't pass), they have nothing they can do to troubleshoot the problem.

I don't see it as strictly internal functionality because of this. I also don't see it as a CLI-specific implementation detail.

komish:
I also don't see a reason to remove the formatters. The API isn't complicated, and it does a single thing - and guides developers on how they can implement their own.

With this said, I don't think it's strictly necessary. If you want it removed, you'll need to update doc/LIBRARY.md to remove the reference to it.

bcrochet:
> > Should library users wish to retrieve the artifacts, I would propose that a (future) API could be added to reach internal and get a tar to get the artifacts. I currently view artifacts as a preflight-specific implementation detail, and not a general purpose thing for library users.
> 
> This doesn't work for me. If a library caller gets an error or failing result, (i.e. they don't pass), they have nothing they can do to troubleshoot the problem.
> 
> I don't see it as strictly internal functionality because of this. I also don't see it as a CLI-specific implementation detail.

I think I can meet you halfway here. While I don't think the current implementation is necessary to be exposed, I do think that some mechanism to retrieve the artifacts would be useful to a library user. So, what I would propose is moving forward here, but then being thoughtful about what API we want to expose. Does that work for you? I do agree that the library user should have the opportunity to review them, I just don't think they necessarily need to *write* them, which is more what this artifacts implementation currently does.

bcrochet:
> I also don't see a reason to remove the formatters. The API isn't complicated, and it does a single thing - and guides developers on how they can implement their own.
> 
> With this said, I don't think it's strictly necessary. If you want it removed, you'll need to update doc/LIBRARY.md to remove the reference to it.

I can come along with this. I'll pull them back out.

certification/results.go
```diff
@@ -1,20 +1,20 @@
...
-	certification.Check
+	internal.Check
 	ElapsedTime time.Duration
 }
 
 type Results struct {
 	TestedImage       string
 	PassedOverall     bool
-	TestedOn          OpenshiftClusterVersion
+	TestedOn          string
```
acornett21, line 17: I'm not sure why we changed this implementation detail, and added a `String()` method in runtime.go. Does this buy us something I'm not seeing?

acornett21: Also, this doesn't seem to work. This is always `unknown/unknown` in the json that gets written to disk, even though internal results struct is correct. I assume it's the genericFormatter that's causing the issue, but didn't look further.

bcrochet: > I'm not sure why we changed this implementation detail, and added a `String()` method in runtime.go. Does this buy us something I'm not seeing?

There was no reason to expose the OpenshiftClusterVersion struct. I'll admit that where it stands right now is completely naive and may (probably?) doesn't work as it stands. I haven't looked at it further either.

bcrochet:
> @bcrochet
> 
> Re: Artifacts
> 
> > So, what I would propose is moving forward here, but then being thoughtful about what API we want to expose
> 
> This is fine - but we've got no roadmap planned on this. So, what do we do to bridge the gap until then?
> 
> I think I would have less of an issue with moving the artifacts into an internal package _if_ executing Preflight (either policy) and having a failing or erroring result could be debugged without it. But as of today, the results for certain checks _themselves_ tell the user top refer to the [artifacts for information on how to troubleshoot.](https://github.com/redhat-openshift-ecosystem/openshift-preflight/blob/main/certification/internal/policy/operator/scorecard_olm_suite.go#L74). Let alone the embedding of the preflight.log as an artifact for library callers.
> 

As I proposed elsewhere, I think the API to artifacts should just be a 'SaveXXX', like 'SaveTar', or 'SaveFs', 'SaveDir'. Or, 'Save(opt ArtifactOption)'. I don't think the writing of artifacts is something that the library user should be relying on the library to do. If a library user wants to add artifacts, they do so to their own artifacts payload without having to reach into the library to do it. Now, to GET the artifacts that the library generates, that makes sense. 

> > I do agree that the library user should have the opportunity to review them, I just don't think they necessarily need to write them, which is more what this artifacts implementation currently does.
> 
> While some of the implementation makes reference to "Write", It's really just a semantic. We can call these methods anything we want. The FilesystemWriter "writes" to the fs, but the MapWriter is just building a map. The `WriteFile` method in the ArtifactWriter interface just maintains the verbiage used historically. IIRC, the previous artifacts directory had the same function.
> 

I get that, however, I don't think it's necessary to expose that from the library. As I previously stated, I think the library should provide the capability to get to the artifacts generated, but not be responsible for writing to the library user's artifacts.

certification/results.go
acornett21: We do 'expose' OpenshiftClusterVersion struct though, in the sense that it is written to disk, and written to the console. And as it stands that does not work for the reasons I stated above.

komish:
> I can get on board with this. Expose our current formatters that we've implemented, but still allow for creating your own formatter.
> 
> This tells me that these should probably be separate PRs.

However you want to do this is fine by me.

komish:
> I get that, however, I don't think it's necessary to expose that from the library. As I previously stated, I think the library should provide the capability to get to the artifacts generated, but not be responsible for writing to the library user's artifacts.

Maybe we're not in the seeing each other regarding this point. The ArtifactWriter isn't something that the library caller _consumes_. It's effectively a choice the library caller makes, telling preflight how to write its artifacts in a way that the library caller can consume. "Write my artifacts here, and I'll retrieve them later".

As an example, if a library caller wanted to... say write an S3 Uploader ArtifactWriter - they could do that, then embed the artifact writer in the context and preflight would write its artifacts however the caller wishes.

komish:
Just to close the discussion loop here, I believe we landed on leaving the artifacts implementation on this branch as is for the moment, while @bcrochet reaches an alternative implementation that may be a bit more clearly defined for our use case. 

The artifacts implementation as it exists in _this_ branch would, therefore, move to the top level and outside of the certification module (as it is not needed).

Regarding logging, I believe we landed on exposing a few of our formatting functions as opposed to the internal formatter logic itself. Formatting, in itself, can be done however the user sees fit, and so we only intend to provide the execution our results formatting that we use as a baseline

@bcrochet please correct me if I've captured this incorrectly, it's been several days since our discussion!

certification/results.go
```diff
@@ -0,0 +1,25 @@
+package certification
+
+import (
+	"time"
+
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/internal/check"
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/internal/runtime"
+)
+
+type openshiftClusterVersion = runtime.OpenshiftClusterVersion
```
komish, line 10: Feel free to ignore - does this still need to be here? It just looks like an alias, but I'm not sure why we would have used it.

komish (approved):
/lgtm

Thanks for this heavy lift!

certification/results.go
bcrochet: It's an alias that reaches into internal so it can still be used as a type here, without having to duplicate it, and so that it isn't exposed as an actual part of the API.

acornett21 (approved):

PR #841, opened 2022-11-21
Replace usage of global logrus instance to instead use local instance

komish:
All of our internal code makes use of `log "github.com/sirupsen/logrus"`'s global instance of `logrus.Logger`, which allowed us to just import the library and immediately call useful functions (e.g. `Debugf(...)`).

If we did nothing, the library caller would have preflight code writing the preflight.log to `os.Stderr` with an unconfigured Logger, so instead the library has special tooling that would discard writes being made in this way, and instead redirect the contents of preflight.log to the artifacts (e.g. the MapWriter).

This presented an issue for consumers who also used Logrus, because this was accomplished by calling `logrus.SetOutput(...)` which would reconfigure the caller's logger as well, causing it to stop sending logs altogether. 

As a workaround, this PR creates an internal instance of the logger and reconfigures existing code to use it. The library caller can now use the global instance of logrus however they see fit without any side effect caused by our code.

IMPORTANT: I know we intend on reconfiguring our logging functionality, but as of this moment we don't have any proposals or work towards that. This PR is just intended to be a stop-gap for the library to merge until we're able to go through and design/refactor to the new implementation.

CC @sebrandon1 - This should resolve the issue you encountered.

### DIFF
--- a/internal/log/log.go
+++ b/internal/log/log.go
@@ -1,0 +1,22 @@
+// Package log is a preflight-local logrus bridge, removing the
+// dependency on logrus' global logger instance.
+package log
+
+import (
+	"github.com/sirupsen/logrus"
+)
+
+var l *logrus.Logger
+
+// Logger returns the configured logger, or a new unconfigured logger
+// if one has not already been configured.
+func Logger() *logrus.Logger {
+	if l == nil {
+		l = logrus.New()
+	}
+
+	return l
+}
+
+// L is a convenience alias to Logger.
+var L = Logger
--- a/certification/pyxis/builder.go
+++ b/certification/pyxis/builder.go
@@ -7,7 +7,7 @@ import (
 	"io"
 	"net/http"
 
-	log "github.com/sirupsen/logrus"
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/internal/log"
 )
 
 // certificationInputBuilder facilitates the building of CertificationInput for
@@ -68,7 +68,7 @@ func (b *certificationInputBuilder) Finalize() (*CertificationInput, error) {
 // Errors are logged, but will not halt execution.
 func (b *certificationInputBuilder) WithCertImage(r io.Reader) *certificationInputBuilder {
 	if err := b.storeCertImage(r); err != nil {
-		log.Error(fmt.Errorf("cert image could not be stored: %v", err))
+		log.L().Error(fmt.Errorf("cert image could not be stored: %v", err))
 	}
 
 	return b
@@ -78,7 +78,7 @@ func (b *certificationInputBuilder) WithCertImage(r io.Reader) *certificationInp
 // Errors are logged, but will not halt execution.
 func (b *certificationInputBuilder) WithPreflightResults(r io.Reader) *certificationInputBuilder {
 	if err := b.storePreflightResults(r); err != nil {
-		log.Error(fmt.Errorf("preflight results could not be stored: %v", err))
+		log.L().Error(fmt.Errorf("preflight results could not be stored: %v", err))
 	}
 
 	return b
@@ -88,7 +88,7 @@ func (b *certificationInputBuilder) WithPreflightResults(r io.Reader) *certifica
 // Errors are logged, but will not halt execution.
 func (b *certificationInputBuilder) WithRPMManifest(r io.Reader) *certificationInputBuilder {
 	if err := b.storeRPMManifest(r); err != nil {
-		log.Error(fmt.Errorf("rpm manifest could not be stored: %v", err))
+		log.L().Error(fmt.Errorf("rpm manifest could not be stored: %v", err))
 	}
 
 	return b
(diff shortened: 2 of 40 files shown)


internal/log/log.go
```diff
@@ -0,0 +1,22 @@
+// Package log is a preflight-local logrus bridge, removing the
+// dependency on logrus' global logger instance.
+package log
```
acornett21, line 3: If we go with a log package now, we are kind of stuck with this forever right?

internal/log/log.go
```diff
@@ -0,0 +1,22 @@
...
+func Logger() *logrus.Logger {
+	if l == nil {
+		l = logrus.New()
+	}
+
+	return l
+}
+
+// L is a convenience alias to Logger.
+var L = Logger
```
skattoju4, line 22: nit: this `log.L().Stuff()` looks ugly ... may be not worth implementing an interface so that we can consume logging like `log.Stuff()` but looks less ugly imho like in [opcap](https://github.com/opdev/opcap/blob/main/internal/logger/logger.go)

internal/log/log.go
komish: Why would we be stuck with it forever?

internal/log/log.go
skattoju: actually would say not a nit 🧐

komish: I considered this - but It feels a bit unnecessary, and like quite a bit of overhead, to add all of `logrus.Logger`'s methods as package-level functions just to achieve the same API we were getting with the import of logrus itself.

I suppose I can make it so if we deem it truly beneficial, but I do feel this achieves the same result without having N (where N == the number of methods `logrus.Logger` has) additional functions in our own code. We would need something like Warn/Info/Error/Debug/Trace + their formatted string variants (e.g. `Infof`), SetOutput, SetLevel, SetFormatter... and each of these would then technically need their own test to not impact coverage.

In theory, I could have just allowed access to the `l` variable itself, but then I'd have to mess with init functions. Ultimately, this felt cleaner (of the options).

Let me know.

acornett21: Going the opcap route seems overkill, and also confusing for our use case. Why add overhead and methods that we have to maintain?

internal/log/log.go
acornett21: I missed that this is internal, so I guess we are free to change it at will and not affect the API layer.

komish: Yep. That's the idea. I expect it to change in the not-so-distant future, but the level of change to rearchitect this requires changing all of our check code in non-trivial ways. I considered this PR to be effectively a trivial change since the underlying behavior doesn't change.

internal/log/log.go
skattoju: Does not have to be everything just what we use. I was thinking from the angle of the rest of code being decoupled from Logrus so that we can swap in a different logger in the future with minimal impact. However, I imagine most loggers have a similar interface so may not be worth it. On the other hand having a way to get a reference to the logger like `logger = log.GetLogger()` would let consumers do logger.Stuff() too right ?

komish: Because this package is internal, we would be the only consumers.

skattoju: By consumers I don't mean users, just things that use the log package. It could be "decoupled" if we only referenced Logrus specific things in the log package and everything else (all consumers) only consume what the log package exposes. Yeah its internal but we still have a "global" logger i.e. everything that consumes logging from the log package uses the same instance regardless of `log.L().Stuff();` vs `logger = log.Logger(); logger.stuff();` because of the "singleton" initialization.

acornett21: > Does not have to be everything just what we use.
We have a flag exposed to the caller for them to use how they would like, which means we'd have to code everything logrus supports, which is overkill IMO.

komish: The definition of a logging interface that we then consume definitely has merit - and very well may be a part of the logging refactor I described in the original message. That said, this is not intended to be that refactor, but is rather just a bugfix.

skattoju: Fair points but what is the case for `log.L().Stuff()`; vs `logger = log.Logger(); logger.stuff();` ?

komish: The former option `log.L().Stuff()` saves a variable assignment. Otherwise, those are identical.

skattoju: One is more ugly 😝. But can safely downgrade this conversation back to nit. I didn't read the PR description fully 😅.

komish: I think the core of what you were asking is 100% on the money! Make our internal/log package the _only_ import of our code. We did similar for viper a while back.

I just don't know if I want to do it _here_. We're on a branch here that hasn't merged to main yet. I'm trying to make changes only scoped to the end-goal of the library becoming published. The logging refactor is imminent, don't you worry.

bcrochet: I still think the library should use https://github.com/go-logr/logr and leave the actual log implementation as a detail for the user. As long as the logger given supports that interface, win.

bcrochet: > For what it's worth - we're not actually decoupled from logrus here - we're still very dependent on logrus. This just lets us avoid clobbering another consumer's global logrus.Logger configuration because we _internally_ refer to our own.

I'm good with this being a stop-gap, knowing that we still want a major log refactor.

bcrochet (approved):
/lgtm

acornett21 (approved):

PR #843, opened 2022-11-28
Refactor logging to use the logr.Logger interface

bcrochet:
By using logr.Logger, it will allow for a library user or the cli to inject a logr that conforms to the logr.Logger interface. This frees up the library from having to do any kind of strange log tricks in order to handle logging.


internal/bundle/bundle.go
```diff
@@ -67,7 +68,7 @@ func Validate(ctx context.Context, imagePath string) (*Report, error) {
 			return nil, fmt.Errorf("%v", err)
 		}
 		if k8sVer, found := ocpToKubeVersion[targetVersion]; found {
-			log.L().Debugf("OpenShift %s detected in annotations. Running with additional checks enabled.", targetVersion)
+			logger.V(log.DEBUGLEVEL).Info("running with additional checks enabled", "version", targetVersion)
```
komish, line 71: Did we mean to remove the context related to the OpenShift version, and how that triggered additional checks?

cmd/preflight/cmd/root.go
```diff
@@ -48,7 +51,13 @@ func rootCmd() *cobra.Command {
 }
 
 func Execute() error {
-	return rootCmd().ExecuteContext(context.Background())
+	zapLog, err := zap.NewDevelopment()
```
komish, line 54: Is this supposed to be `zap.NewProduction()`? I'm not super familiar with Zap.

cmd/preflight/cmd/root.go
```diff
@@ -48,7 +51,13 @@ func rootCmd() *cobra.Command {
...
+	if err != nil {
+		panic(fmt.Sprintf("who watches the watchmen (%v)?", err))
```
komish, line 56: I imagine this is here for WIP purposes, but just wanted to point it out to be sure.

komish:
Left a couple of comments - but I also want to get into some of the detail here to make sure I'm understanding.

The key difference is that this intends to pull a logger from context. Our internal code makes direct calls to our internal logrus instance, and so to prevent our library from writing to os.Stderr, we would detect that the caller is NOT our own CLI, and then effectively discard messages that were headed to os.Stderr.

For clarity, we still wrote those log lines - just to an artifact "preflight.log". No need to get into those semantics, but I wanted to make sure it was clear that we didn't simply ignore those log lines, we just stored them in a different way for library callers.

What feels kind of strange is, for example, this bit here - 
https://github.com/redhat-openshift-ecosystem/openshift-preflight/blob/a40d38b91b27baf24916cbd06d95194fa8b9e39e/internal/bundle/bundle.go#L35-L37

We effectively need to ask the context for a logger or... discard those lines? This reads to me like it's effectively a no-op, so I agree with the intention. It just feels like a weird place to be calling something like this. I'm not really sure what I expect instead, so I guess it's fair that this comment may not be super helpful.

At its core, this approach makes sense. I do find myself wondering if we want really want to introduce logr AND zap (where logr is our interface, and zap is our CLI's implementation) as dependencies in order to accomplish this, vs. defining our own in potentially less LOC. With that said, I do suppose it's no different than only accepting a logrus instance, which is kind of what we do now.

The above point goes along with the idea that @skattoju suggested, which was that we consider using our own logging interfaces and allowing the implementation details to be whatever want, such as zap. Gives us a little bit more control. With that said, I've heard of logr before. I'm not sure of its general popularity and consumption - and what we'd be doing here is making folks implement that interface and use its context helpers - but I'm not opposed to the idea.

Just some discussion points! Looking forward to hear everyone's thoughts.

internal/log/log.go
```diff
@@ -20,3 +24,54 @@ func Logger() *logrus.Logger {
 
 // L is a convenience alias to Logger.
 var L = Logger
```
skattoju, line 26: do we still need this ?

internal/bundle/bundle.go
bcrochet: Something in a Debug statement triggered additional checks?

cmd/preflight/cmd/root.go
bcrochet: It's just PoC. This will absolutely change to set up logging how preflight cli wants it.

cmd/preflight/cmd/root.go
bcrochet: Yes. Copy pasta.

internal/log/log.go
bcrochet: @komish is correct. I just wanted to get the conversation started without breaking the whole thing.

internal/bundle/bundle.go
komish: The original log message said contained `OpenShift %s detected in annotations` but it doesn't exist in the newly formatted log statement message.

```suggestion
			logger.V(log.DEBUGLEVEL).Info("OpenShift %s detected in annotations. Running with additional checks enabled", "version", targetVersion)
```

This suggestion doesn't take into account whether this is a valid way to do string interpolation for this `Info` method.

bcrochet: Oh... It's not string interpolation. It's structured logging. Instead of interpolation, you can use "msg, key, value, key, value, ..."

If you want to keep interpolation, then you would use fmt.Sprintf()

internal/log/log.go
```diff
@@ -20,3 +24,54 @@ func Logger() *logrus.Logger {
...
+
+func (s bufferSink) Enabled(level int) bool {
+	return true
+}
+
+func (s bufferSink) Error(err error, msg string, keysAndValues ...interface{}) {
+	s.buffer.WriteString(fmt.Sprintf("%s %v %s %v\n", s.name, err.Error(), msg, keysAndValues))
+}
+
+func (s bufferSink) Info(level int, msg string, keysAndValues ...interface{}) {
```
skattoju, line 61: Super weird that log.Info() takes a level param I guess its a go thing ? Would it be better to have Error, Warn, Info Debug and Trace funcs ?

bcrochet: It's kind of a philosophical choice. The idea being... Logs are either two things. Info, or Error. If you think about it, it makes sense. By increasing the level of a log, you are saying "This is probably not something you want to see all the time". Hence, debug and trace. Errors on the whole would always be visible in the log, and probably would not increase the level. Warning is just an alias for Info really.

internal/policy/operator/deployable_by_olm.go
```diff
@@ -507,55 +523,57 @@ func (p *DeployableByOlmCheck) installedCSV(ctx context.Context, operatorData op
...
+		logger.Info("warning: unable to retrieve the target namespace")
 	} else {
 		if err := p.writeToFile(ctx, targetNamespace); err != nil {
-			log.L().Errorf("could not write target namespace to storage")
+			logger.Error(err, "could not write target namespace to storage")
 		}
 	}
 
-	log.L().Trace("Deleting the resources created by DeployableByOLM Check")
+	logger.V(2).Info("deleting the resources created by DeployableByOLM Check")
```
komish, line 576: `V(2)` is Debug? Trace? I somewhat prefer a constant/variable value so I don't have to guess. `DBG` and `TRC`?

internal/policy/operator/deployable_by_olm.go
```diff
@@ -362,7 +370,7 @@ func (p *DeployableByOlmCheck) generateOperatorGroupData(operatorData *operatorD
 	case operatorsv1alpha1.InstallModeTypeAllNamespaces:
 		targetNamespaces = []string{}
 	}
-	log.L().Debugf("The OperatorGroup's TargetNamespaces is %s", targetNamespaces)
+	logger.V(1).Info("OperatorGroup TargetNamespaces", "namespace", targetNamespaces)
```
komish, line 373: There are a couple of spots where we lose the phrasing of the log line in order to utilize the k/v structure. I'm not sure how I feel about it, so I'm just putting this here as a reminder to see if it seems off to anyone else.

internal/policy/operator/deployable_by_olm.go
```diff
@@ -267,7 +273,7 @@ func (p *DeployableByOlmCheck) setUp(ctx context.Context, operatorData *operator
 			return err
 		}
 	} else {
-		log.L().Debug("No docker config file is found to access the index image in private registries. Proceeding...")
+		logger.V(1).Info("no docker config file is found to access the index image in private registries")
```
komish, line 276: +1 to removing `Proceeding...`, but should we instead consider saying something like `using anonymous auth`?

internal/policy/operator/deployable_by_olm.go
```diff
@@ -128,7 +130,7 @@ func (p *DeployableByOlmCheck) Validate(ctx context.Context, bundleRef certifica
 		return false, fmt.Errorf("%v", err)
 	}
 
-	log.L().Debugf("The operator Metadata is %+v", *operatorData)
```
komish, line 131: Do you know off hand if we'll get the same output from the "%+v" formatter using the k/v setup in the new logger?

internal/policy/container/runs_as_nonroot.go
```diff
@@ -33,20 +33,22 @@ func (p *RunAsNonRootCheck) getDataToValidate(image cranev1.Image) (string, erro
...
-		log.L().Info("USER value must be provided and be a non-root value for this check to pass")
+		logger.Info("detected empty USER. Presumed to be running as root")
+		logger.Info("USER value must be provided and be a non-root value for this check to pass")
 		return false, nil
 	}
 
 	if user == "0" || user == "root" {
-		log.L().Infof("detected USER specified as root: %s", user)
-		log.L().Info("USER other than root is required for this check to pass")
+		logger.Info(fmt.Sprintf("detected USER specified as root: %s", user))
```
komish, line 46: Maybe we don't need to Infof this? Seems like the log message tells us exactly what... user we found.

internal/policy/container/max_layers.go
```diff
@@ -25,15 +25,15 @@ func (p *MaxLayersCheck) Validate(ctx context.Context, imgRef certification.Imag
...
 }
 
 func (p *MaxLayersCheck) getDataToValidate(image cranev1.Image) ([]cranev1.Layer, error) {
 	return image.Layers()
 }
 
-func (p *MaxLayersCheck) validate(layers []cranev1.Layer) (bool, error) {
-	log.L().Debugf("detected %d layers in image", len(layers))
+func (p *MaxLayersCheck) validate(ctx context.Context, layers []cranev1.Layer) (bool, error) {
+	logr.FromContextOrDiscard(ctx).V(5).Info("number of layers detected in image", "layerCount", len(layers))
```
komish, line 36: This needs to change to be `V(1)`?

internal/policy/container/has_required_labels.go
```diff
@@ -24,15 +24,17 @@ func (p *HasRequiredLabelsCheck) Validate(ctx context.Context, imgRef certificat
 		return false, fmt.Errorf("could not retrieve image labels: %v", err)
 	}
 
-	return p.validate(labels)
+	return p.validate(ctx, labels)
```
komish, line 27: Would it make sense to reach into the struct for the context instead of accepting it in the internal validate function?

That would require passing a context into the check at init time instead of at Validate() time... Maybe that's worse.

internal/engine/engine.go
```diff
@@ -127,26 +128,26 @@ func (c *CraneEngine) ExecuteChecks(ctx context.Context) error {
...
-		log.L().Debugf("writing container filesystem to output dir: %s", containerFSPath)
+		logger.V(1).Info("writing container filesystem", "outputDirectory", containerFSPath)
 		err = crane.Export(img, w)
 		if err != nil {
 			// TODO: Handle this error more effectively. Right now we rely on
 			// error handling in the logic to extract this export in a lower
 			// line, but we should probably exit early if the export encounters
 			// an error, which requires watching multiple error streams.
-			log.L().Error("unable to export and flatten container filesystem:", err)
+			logger.V(5).Error(err, "unable to export and flatten container filesystem")
```
komish, line 144:
```suggestion
			logger.V(1).Error(err, "unable to export and flatten container filesystem")
```

internal/engine/engine.go
```diff
@@ -178,23 +179,23 @@ func (c *CraneEngine) ExecuteChecks(ctx context.Context) error {
 
 	if c.IsBundle {
 		// Record test cluster version
-		c.results.TestedOn, err = openshift.GetOpenshiftClusterVersion(c.Kubeconfig)
+		c.results.TestedOn, err = openshift.GetOpenshiftClusterVersion(ctx, c.Kubeconfig)
 		if err != nil {
-			log.L().Errorf("could not determine test cluster version: %v", err)
+			logger.V(5).Error(err, "could not determine test cluster version")
```
komish, line 184:
```suggestion
			logger.V(1).Error(err, "could not determine test cluster version")
```

internal/engine/engine.go
```diff
@@ -228,7 +229,7 @@ func (c *CraneEngine) ExecuteChecks(ctx context.Context) error {
 		// hash the contents of the bundle.
 		md5sum, err := generateBundleHash(ctx, c.imageRef.ImageFSPath)
 		if err != nil {
-			log.L().Errorf("could not generate bundle hash: %v", err)
+			logger.V(5).Error(err, "could not generate bundle hash")
```
komish, line 232:
```suggestion
			logger.V(1).Error(err, "could not generate bundle hash")
```

internal/engine/engine.go
```diff
@@ -107,10 +108,10 @@ func (c *CraneEngine) ExecuteChecks(ctx context.Context) error {
 	if err != nil {
 		return fmt.Errorf("failed to create temporary directory: %v", err)
 	}
-	log.L().Debug("temporary directory is ", tmpdir)
+	logger.V(1).Info("temporary directory is ", tmpdir)
 	defer func() {
 		if err := os.RemoveAll(tmpdir); err != nil {
-			log.L().Errorf("unable to clean up tmpdir %s: %v", tmpdir, err)
+			logger.V(5).Error(err, "unable to clean up tmpdir", "tempDir", tmpdir)
```
komish, line 114:
```suggestion
			logger.V(1).Error(err, "unable to clean up tmpdir", "tempDir", tmpdir)
```

internal/engine/engine.go
```diff
@@ -524,9 +532,10 @@ func getBgName(srcrpm string) string {
 }
 
 func writeRPMManifest(ctx context.Context, containerFSPath string) error {
+	logger := logr.FromContextOrDiscard(ctx)
 	pkgList, err := rpm.GetPackageList(ctx, containerFSPath)
 	if err != nil {
-		log.L().Errorf("could not get rpm list, continuing without it: %v", err)
+		logger.V(5).Error(err, "could not get rpm list, continuing without it")
```
komish, line 538:
```suggestion
		logger.V(1).Error(err, "could not get rpm list, continuing without it")
```

internal/policy/container/max_layers.go
komish: I marked this in other places where I also saw `V(5)`, but the fact that it exists in multiple places has me unsure that we actually want to swap it.

komish (commented):
Just for starters. I'll have to give this another pass and some real-world testing.

internal/policy/operator/scorecard_check.go
```diff
@@ -20,11 +21,13 @@ type scorecardCheck struct {
 
 //nolint:unparam // ctx is unused. Keep for future use.
 func (p *scorecardCheck) validate(ctx context.Context, items []operatorsdk.OperatorSdkScorecardItem) (bool, error) {
+	logger := logr.FromContextOrDiscard(ctx)
+
```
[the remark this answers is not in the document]
bcrochet: If it were BEFORE the line, yes.

internal/policy/operator/scc_info.go
```diff
@@ -43,17 +44,19 @@ func (p *securityContextConstraintsInCSV) dataToValidate(ctx context.Context, im
...
 		return true, nil
 	}
 
 	if len(requestedSccList) != 1 {
 		return false, fmt.Errorf("only one scc should be requested at a time")
 	}
 
-	log.L().Infof("A custom scc was specified: %s , action may be needed by a cluster admin, please check the "+
-		"operator documentation for more information", requestedSccList[0])
+	logger.Info(fmt.Sprintf("A custom scc was specified: %s , action may be needed by a cluster admin, please check the "+
```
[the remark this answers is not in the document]
bcrochet: If anything, it would probably increase the readability. I'll change it.

internal/policy/operator/deployable_by_olm_test.go
```diff
@@ -124,7 +124,7 @@ var _ = Describe("DeployableByOLMCheck", func() {
 
 	DescribeTable("Image Registry validation",
 		func(bundleImages []string, expected bool) {
-			ok := checkImageSource(bundleImages)
+			ok := checkImageSource(context.Background(), bundleImages)
```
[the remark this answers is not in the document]
bcrochet: My thought on this is that if we need to add logging for a test, we can always add for individual tests (or even just for debugging). But I'd like to default to not specifying a logger in tests. I need to figure out if we can even turn off the logging that comes from operator-sdk.

internal/policy/operator/deployable_by_olm.go
bcrochet: V(2) is trace.

Yeah... log.DBG or log.DEBUG. I do like the smaller one. I'll add that back

internal/policy/operator/deployable_by_olm.go
bcrochet: I get that. For this particular one, I don't think that's the case. But if you do find others that are candidates, point them out.

internal/policy/operator/deployable_by_olm.go
bcrochet: I do not. Will need to test that specifically.

internal/policy/container/runs_as_nonroot.go
bcrochet: Could make it say specified as root or UID 0... Agreed.

internal/policy/container/max_layers.go
bcrochet: That was missed since I did a replace all.. And it didn't match my search term.

internal/policy/container/has_required_labels.go
bcrochet: I'd rather pass at the point of use. I don't think it's good practice to store a context like that.

internal/policy/operator/validate_operator_bundle.go
```diff
@@ -33,14 +33,16 @@ func (p *ValidateOperatorBundleCheck) dataToValidate(ctx context.Context, imageP
...
 }
 
-func (p *ValidateOperatorBundleCheck) validate(ctx context.Context, report *bundle.Report) (bool, error) { //nolint:unparam // save context for future use
+func (p *ValidateOperatorBundleCheck) validate(ctx context.Context, report *bundle.Report) (bool, error) {
+	logger := logr.FromContextOrDiscard(ctx)
+
 	if !report.Passed || len(report.Results) > 0 {
 		for _, output := range report.Results {
 			for _, result := range output.Errors {
-				log.L().Error(result.Error())
```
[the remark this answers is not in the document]
bcrochet: Changing to errors.New

internal/policy/operator/scorecard_olm_suite.go
```diff
@@ -33,9 +34,10 @@ func NewScorecardOlmSuiteCheck(operatorSdk operatorSdk, ns, sa string, kubeconfi
 }
 
 func (p *ScorecardOlmSuiteCheck) Validate(ctx context.Context, bundleRef certification.ImageReference) (bool, error) {
-	log.L().Trace("Running operator-sdk scorecard Check for ", bundleRef.ImageURI)
+	logger := logr.FromContextOrDiscard(ctx)
+
+	logger.V(2).Info("running operator-sdk scorecard check", "image", bundleRef.ImageURI)
 	selector := []string{"suite=olm"}
-	log.L().Tracef("--selector=%s", selector)
```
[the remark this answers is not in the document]
bcrochet: IIRC my motivation for removing it was that what it is printing out is literally in the line above it. I can put it back if you feel strongly about it.

internal/engine/engine.go
bcrochet: I'm removing the V().

internal/policy/operator/deployable_by_olm.go
bcrochet:
```
time="2022-12-02T08:18:25-05:00" level=debug msg="operator metadata" metadata="{\"CatalogImage\":\"quay.io/opdev/simple-demo-operator-catalog:v0.0.6\",\"Channel\":\"stable\",\"PackageName\":\"simple-demo-operator\",\"App\":\"simple-demo-operator\",\"InstallNamespace\":\"simple-demo-operator\",\"TargetNamespace\":\"simple-demo-operator-target\",\"InstallModes\":{\"AllNamespaces\":{\"type\":\"AllNamespaces\",\"supported\":true},\"MultiNamespace\":{\"type\":\"MultiNamespace\",\"supported\":false},\"OwnNamespace\":{\"type\":\"OwnNamespace\",\"supported\":true},\"SingleNamespace\":{\"type\":\"SingleNamespace\",\"supported\":false}},\"CsvNamespaces\":null,\"InstalledCsv\":\"\"}"
```

internal/engine/engine.go
bcrochet: Removed the V()

internal/engine/engine.go
bcrochet: Removed the V()

internal/engine/engine.go
bcrochet: Removed the V()

internal/log/log.go
```diff
@@ -3,20 +3,56 @@
...
+	"bytes"
+	"fmt"
+
+	"github.com/go-logr/logr"
 )
 
-var l *logrus.Logger
+const (
+	DBG int = 1
+	TRC int = 2
```
acornett21, line 14: Are these numbers just random? Looking at the implementation it doesn't seem like it's a typical linear scale ie 1-5 (trace, debug, warn, info, error) for the levels. And seems that it can be any number.

internal/bundle/bundle.go
```diff
@@ -67,7 +68,7 @@ func Validate(ctx context.Context, imagePath string) (*Report, error) {
 			return nil, fmt.Errorf("%v", err)
 		}
 		if k8sVer, found := ocpToKubeVersion[targetVersion]; found {
-			log.L().Debugf("OpenShift %s detected in annotations. Running with additional checks enabled.", targetVersion)
+			logger.V(log.DBG).Info("running with additional checks enabled because of the OpenShift version detected", "version", targetVersion)
```
acornett21, line 71: It seems we no longer have a formatter, is this syntax doing a literal replacement of the word `version` in the sentence with the value `targetVersions`?

cmd/preflight/cmd/root.go
```diff
@@ -6,15 +6,14 @@ import (
 	"io"
 	"os"
 
-	"github.com/sirupsen/logrus"
-
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/artifacts"
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/version"
 
+	"github.com/bombsimon/logrusr/v4"
```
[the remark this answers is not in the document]
bcrochet: It adds the bridge from logrus to logr, as logrus itself does not implement go-logr/logr. It's a small adapter lib.

internal/bundle/bundle.go
komish: This is a variadic method accepting an arbitrary number of values at the end, and it associates each odd parameter as a key with the following even parameters as a value (excluding the very first parameter). See: https://github.com/go-logr/logr/blob/master/logr.go#L270

acornett21: @komish I looked at this more after my question, and I see things how you explained them. I just don't love this syntax, but internally `logr` ends up internally converting to `fmt.Sprintf("Openshift %s, version)` anyways, and that's all that matters I guess.

docs/LIBRARY.md
```diff
@@ -88,6 +88,9 @@ artifacts in a map of filenames to `io.Reader`. The `preflight.log` is written
 to the ArtifactWriter for library callers, and can be accessed once checks have
 been executed.
 
+A logger of choice should be added to the context with logr.NewContext. Otherwise,
+no logs will be available from the library.
+
```
komish, line 93: The code snippet likely also needs to be updated, because it currently reflects a preflight.log being written to the artifact writer and I don't think that's the case anymore with this patch.

internal/authn/keychain.go
```diff
@@ -28,17 +30,21 @@ func WithDockerConfig(dockercfg string) PreflightKeychainOption {
...
 // This operates as a singleton. If provided an option, that option overwrites
 // the single instance of PreflightKeychain. If provided no option, the keychain
 // is returned as already configured.
-func PreflightKeychain(opts ...PreflightKeychainOption) craneauthn.Keychain {
+func PreflightKeychain(ctx context.Context, opts ...PreflightKeychainOption) craneauthn.Keychain {
 	for _, opt := range opts {
 		opt(&keychain)
 	}
 
+	keychain.ctx = ctx
```
komish, line 46: Should we check `ctx` for nil here? Otherwise, there's no real point in initializing a few lines up because it will _always_ get replaced here.

internal/log/log.go
```diff
@@ -3,20 +3,56 @@
...
+)
 
-// Logger returns the configured logger, or a new unconfigured logger
-// if one has not already been configured.
-func Logger() *logrus.Logger {
-	if l == nil {
-		l = logrus.New()
+func NewBufferSink(buffer *bytes.Buffer) logr.LogSink {
```
komish, line 17: Wow, we had to generate a net-new buffer sink? There was not an existing one we could have leveraged from the logr package?

internal/pyxis/pyxis_suite_test.go
```diff
@@ -44,129 +39,138 @@ func mustWrite(w io.Writer, s string) {
...
-		response.WriteHeader(http.StatusConflict)
-	case request.Method == http.MethodGet && request.Header["X-Api-Key"][0] == "my-bad-rpmmanifest-401-api-token":
-		response.WriteHeader(http.StatusUnauthorized)
-	case request.Header["X-Api-Key"][0] == "my-bad-rpmmanifest-api-token":
-		response.WriteHeader(http.StatusUnauthorized)
-	case request.Method == http.MethodPost && request.Header["X-Api-Key"][0] == "my-update-image-success-api-token":
-		mustWrite(response, `{"_id":"updateImage"}`)
-	default:
-		mustWrite(response, `{"_id":"blah"}`)
+func pyxisRPMManifestHandler(ctx context.Context) func(http.ResponseWriter, *http.Request) {
```
komish, line 122: NIT: could the return value have been http.HandlerFunc?

komish (approved):
A single nitpick. Disregard if you want!. Logs seem to flow how I expect. So -

/lgtm

internal/log/log.go
skattoju: Would tend to think that Error is also Info 🤷 I guess that the difference is you can't turn errors off ? .. Feels more verbose and less clear for the logging functionality consumer (log lib user) .. 

` logger.V(log.TRACELEVEL).Info("unnecessary detail for general scenarios") 
`vs
` logger.Trace("extra detail for execution tracing") 
`

skattoju: but if its a standard i guess its ok.. so will resolve this

internal/log/log.go
bcrochet: I didn't see one, but I also didn't look that hard. It took all of about 3 minutes to create.

internal/pyxis/pyxis_suite_test.go
bcrochet: Yes. It could.

internal/authn/keychain.go
bcrochet: I initialized it to support the tests. There's really no harm in initializing the default then always replacing with the supplied one here.

komish (approved):
This all still looks good to me.

/lgtm

acornett21 (approved):

PR #848, opened 2022-12-05
Fix HasModifiedFiles

### LINKED ISSUE #846: Preflight does not detect the modification of the package and throw the warning or error in the results
## Bug Description
Infinidat certified an openstack image, they replaced a file via the command line  for this package python3-requests-2.20.0-2.1.el8_1.noarch in their  [dockerfile](https://github.com/Infinidat/cinder/commit/dc044c87791fb9d7caed7658d0cdd15d4520286c#diff-e58f4ed034f19f0d4c01795b2255eb8860f9a9bf478f90d4bb18e58d832197f8)
here is the command line :
 RUN curl -Ls [ https://raw.githubusercontent.com/psf/requests/v2.20.0/requests/__init__.py|https://raw.githubusercontent.com/psf/requests/v2.20.0/requests/_init.py] > /usr/lib/python3.6/site-packages/requests/init_.py

Whent they ran the preflight test for this image, it passed.
1.4.1

## Steps to Reproduce:
1) build the image refer to the [dockerfile](https://github.com/Infinidat/cinder/commit/dc044c87791fb9d7caed7658d0cdd15d4520286c#diff-e58f4ed034f19f0d4c01795b2255eb8860f9a9bf478f90d4bb18e58d832197f8)
2) run the preflight tool with the image

## Expected Result
Test name: HasModifiedFiles -should fail
Policy guide: https://access.redhat.com/documentation/en-us/red_hat_software_certification/8.53/html-single/red_hat_openshift_software_certification_policy_guide/index#assembly-requirements-for-container-images_openshift-sw-cert-policy-introduction
*************************
Container images must not change content provided by Red Hat packages or layers except for files that both you or the customers can change, such as configuration files.

## Actual Result
The test results showed pass
 https://connect.redhat.com/projects/5f018e64ab0698a38db07f54/overview
sha256:1aebded18e9317ab3338d8f34ba8bcc5aebe98e566a51eae6775389dda059f68 -> pass

bcrochet:
* (almost) Complete rewrite of the check
* Ignore any path that starts with "etc/" or "var/"
* Ignore any RPM file that has config, doc, or config(missingok) flags
* Check each layer against the last, rather than all the flattened layers at once
    
Fixes #846


internal/policy/container/has_modified_files.go
```diff
@@ -45,118 +48,155 @@ func (p *HasModifiedFilesCheck) Validate(ctx context.Context, imgRef image.Image
...
 // modified/added via layers in the image.
 func (p *HasModifiedFilesCheck) getDataToValidate(ctx context.Context, imgRef image.ImageReference) (*packageFilesRef, error) {
-	logger := logr.FromContextOrDiscard(ctx)
-
-	// Get a list of packages from the RPM database. This avoids having to rely on
-	// rpm, dnf, yum, etc. being installed in the image.
-	pkgList, err := rpm.GetPackageList(ctx, imgRef.ImageFSPath)
+	fs := afero.NewOsFs()
+	layerDir, err := afero.TempDir(fs, "", "rpm-layers-*")
+	defer func() {
```
komish, line 53: Two Items here:

---
Bit of a nit

Let's move this deferral below the error checking.

Logically, I know that fs.RemoveAll does not return an error if the underlying fs doesn't exist. But layerDir and err are assigned together, and I will always do a double-take if the assigned variable is used before the error assigned in the same line is checked.

---
I'm sure golangci-lint told you to explicitly ignore this error using `_`, but I would probably justify ignoring that lint enforcement here. It's not too uncommon to do what you had before `defer thingThatMightErrorButWeDoNotCare()`

internal/policy/container/has_modified_files.go
```diff
@@ -45,118 +48,155 @@ func (p *HasModifiedFilesCheck) Validate(ctx context.Context, imgRef image.Image
...
-		return nil, fmt.Errorf("could not get image layers: %w", err)
+		return nil, err
 	}
 
-	files := make([][]string, 0, len(layers))
+	layerIDs := make([]string, 0, len(layers))
+
+	files := make(map[string][]string, len(layers))
+	layerPackages := make(map[string]map[string]struct{}, len(layers))
+	layerPackageFiles := make(map[string]map[string]string, len(layers))
```
komish, line 73: Wondering if we should create a few local types to avoid using multiple layers of maps and slices, even if just for readability?

bcrochet: I was thinking that. Your comment just confirms it.

internal/policy/container/has_modified_files.go
bcrochet: > Two Items here:
> 
> Bit of a nit
> 
> Let's move this deferral below the error checking.
> 
> Logically, I know that fs.RemoveAll does not return an error if the underlying fs doesn't exist. But layerDir and err are assigned together, and I will always do a double-take if the assigned variable is used before the error assigned in the same line is checked.

linting told me to do that anyway, so it should already be moved.

> 
> I'm sure golangci-lint told you to explicitly ignore this error using `_`, but I would probably justify ignoring that lint enforcement here. It's not too uncommon to do what you had before `defer thingThatMightErrorButWeDoNotCare()`

I actually do prefer the explicit ignoring of the error. It makes it clear that the error is being ignored on purpose, rather than just being an oversight.

internal/policy/container/has_modified_files.go
```diff
@@ -45,118 +48,155 @@ func (p *HasModifiedFilesCheck) Validate(ctx context.Context, imgRef image.Image
...
-		var wg sync.WaitGroup
-		wg.Add(1)
-		go func() {
-			// For each path in the pathChan, add it to the layer's
-			// list of files.
-			layerFiles := make([]string, 0)
-			for path := range pathChan {
-				layerFiles = append(layerFiles, path)
+
+		if err := func() error {
```
komish, line 87: Why an anonymous function here, vs. a standalone?

internal/policy/container/has_modified_files.go
```diff
@@ -45,118 +48,155 @@ func (p *HasModifiedFilesCheck) Validate(ctx context.Context, imgRef image.Image
...
-	files := make([][]string, 0, len(layers))
+	layerIDs := make([]string, 0, len(layers))
+
+	files := make(map[string][]string, len(layers))
+	layerPackages := make(map[string]map[string]struct{}, len(layers))
+	layerPackageFiles := make(map[string]map[string]string, len(layers))
 
-	// Uncompress each layer and build a slice containing the files
-	// modified by each layer.
 	for _, layer := range layers {
```
komish, line 75: We removed the comments here, but the logic is still fairly dense to follow. Can we rewrite the comment in an applicable way?

internal/policy/container/has_modified_files.go
```diff
@@ -45,118 +48,155 @@ func (p *HasModifiedFilesCheck) Validate(ctx context.Context, imgRef image.Image
...
 
-	return &packageFilesRef{files, packageFiles}, nil
+	return &packageFilesRef, nil
 }
 
-// dropFirstLayerIfEmpty will evaluate the length of the first layer and will drop its entry in files if it's empty.
-// This avoids establishing a baseline of files installed in the original layer with an empty list. An example
-// of when this should occur is in cases where the image was built FROM scratch, such as ubi-micro. This is required
-// to ensure we don't attempt to validate against an empty layer.
-func (p *HasModifiedFilesCheck) dropFirstLayerIfEmpty(files [][]string) ([][]string, bool) {
```
komish, line 119: Do we no longer need to handle empty first layers?

internal/policy/container/has_modified_files.go
```diff
@@ -45,118 +48,155 @@ func (p *HasModifiedFilesCheck) Validate(ctx context.Context, imgRef image.Image
...
+				continue
+			}
+			currentPackageVer, found := layerPackageFiles[layerID][layerFile]
+			if !found {
+				// Package was removed? This is bad too. Set modified files, and exit
 				modifiedFilesDetected = true
+				break
 			}
+			if lastPackageVer == currentPackageVer {
+				// This means the file was changed, but the package was not. Stinky.
```
komish, line 186: Stinky!

acornett21: I liked that too, and the other comments in this block. It covers the things we've talked about nicely IMO.

internal/policy/container/has_modified_files.go
```diff
@@ -212,35 +252,61 @@ func untar(pathChan chan<- string, r io.Reader) error {
...
+		filenames, err := pkg.InstalledFiles()
 		if err != nil {
 			return nil, err
 		}
 		for _, file := range filenames {
 			// A struct is used here, but it is unimportant and
 			// should not have value.
-			installedFiles[file] = struct{}{}
+			// Also trims any leading "." or "/"
+			installedFiles[strings.TrimLeft(file.Path, "./")] = fmt.Sprintf("%s-%s", pkg.Name, pkg.Version)
```
acornett21, line 309: Is this comment valid anymore? its not a `map[string]struct{}` anymore it's a `map[string]string` now, so should the comment be `this is a string that is trimmed`?

acornett21:
@bcrochet Aside from the comments left, was this tested against the image (or Dockerfile) where this issue was reported? And is a similar unit test to that Dockerfile included as well? (I realize the unit test might not be a 1-to-1, hence the word `similar`)

acornett21:
Also, before this lands in a release we need to let Cert Ops know that `dnf update` (and the like) will no longer be valid for partners when there are vulnerabilities in UBI, correct?

internal/policy/container/has_modified_files.go
bcrochet: It's a closure. I could pull it out into it's own standalone function though.

internal/policy/container/has_modified_files.go
bcrochet: Since we are going layer by layer, it basically does it without having to explicitly remove it.

internal/policy/container/has_modified_files.go
bcrochet: I'll modify it.

bcrochet:
> Also, before this lands in a release we need to let Cert Ops know that `dnf update` (and the like) will no longer be valid for partners when there are vulnerabilities in UBI, correct?

It shouldn't. As long as the version of the rpm changes, it's a valid operation for a file to change.

bcrochet:
> @bcrochet Aside from the comments left, was this tested against the image (or Dockerfile) where this issue was reported? And is a similar unit test to that Dockerfile included as well? (I realize the unit test might not be a 1-to-1, hence the word `similar`)

I will generate some Dockerfiles (and commit them to the repo) that hopefully exercises each of the scenarios.

internal/policy/container/has_modified_files.go
```diff
@@ -24,17 +26,26 @@ var _ check.Check = &HasModifiedFilesCheck{}
...
 type packageFilesRef struct {
-	// LayerFiles contains a slice of files modified by layer.
-	LayerFiles [][]string
-	// PackagesFiles is a map of modified files. the anonymous struct
-	// here is just used to allow us to search package files by name
-	// as the key of this map, and doesn't have any value.
-	PackageFiles map[string]struct{}
+	// LayerFiles contains a slice of files created/modified in layer
+	LayerFiles layerFiles
+	// LayerPackages is a map of a map of packages for each layer
```
skattoju, line 38: nit: can be simplified as "maps packages in each layer"

skattoju (approved):

komish:
> Also, before this lands in a release we need to let Cert Ops know that `dnf update` (and the like) will no longer be valid for partners when there are vulnerabilities in UBI, correct?

Just re-iterating this so we don't miss it.

komish (changes requested):
Uhm. I'm not sure where this changed, but the name of this check is `HasModifiedfiles` - and that seems incorrect to me? Shouldn't it be `HasNotModifiedFiles`?

internal/policy/container/has_modified_files.go
```diff
@@ -24,17 +26,26 @@ var _ check.Check = &HasModifiedFilesCheck{}
...
+	LayerPackages layerPackages
+	// LayerPackageFiles maps files to a package name-version
+	LayerPackageFiles layerPackageFiles
+	// Layers is the list of layers in the image
+	Layers []string
```
komish, line 43: I'm not sure this is any more clear with these type definitions. I think what I expected to see instead was something like...

```go
type layerFiles struct {
    id string
    files []string
}
```

Feel free to disregard, though if you do, I'm not sure if the type definitions as written here add value, so you can remove them from my point of view. Up to you.

komish (changes requested):
Overall, LGTM

I went ahead and generated an image that modified `/usr/bin/sha256sum` and the logic caught it.

I think what I'd love to see is log output that tells me which layer saw the change that caused the failure. Without that, we'll have to rely on "can you provide the Dockerfile" which may not always be possible. 

I think I saw that go-logr was removed as an import in this module, so I would assume that means there's absolutely no logging happening here. I'm fairly confident we want _something_ here to help users self-diagnose, or report logic flaws.

---

> I will generate some Dockerfiles (and commit them to the repo) that hopefully exercises each of the scenarios.

@bcrochet I'd love to see this as well. Are you looking to do this after merge or with this PR?

acornett21:
> Uhm. I'm not sure where this changed, but the name of this check is `HasModifiedfiles` - and that seems incorrect to me? Shouldn't it be `HasNotModifiedFiles`?

@komish This check probably should have been what you say above from the get go, IDK if we can change now. We'd need to change docs, then the UI in connect my be confusing for users. But yeah `HasModifiedFiles:PASS` reads incorrectly.

bcrochet:
> Overall, LGTM
> 
> I went ahead and generated an image that modified `/usr/bin/sha256sum` and the logic caught it.
> 
> I think what I'd love to see is log output that tells me which layer saw the change that caused the failure. Without that, we'll have to rely on "can you provide the Dockerfile" which may not always be possible.
> 
> I think I saw that go-logr was removed as an import in this module, so I would assume that means there's absolutely no logging happening here. I'm fairly confident we want _something_ here to help users self-diagnose, or report logic flaws.

Yes, absolutely.

> 
> > I will generate some Dockerfiles (and commit them to the repo) that hopefully exercises each of the scenarios.
> 
> @bcrochet I'd love to see this as well. Are you looking to do this after merge or with this PR?

Probably after. Otherwise, this will never land and will be huge.

internal/policy/container/has_modified_files.go
bcrochet: At this point, it's an implementation detail. It will not affect the outside world, so if we want to change it in the future, we can. For now, this is how my brain is interpreting the data structure. I know there is a (probably) a better way, but I don't really want to focus on that atm

jfrancin:
> Also, before this lands in a release we need to let Cert Ops know that `dnf update` (and the like) will no longer be valid for partners when there are vulnerabilities in UBI, correct?

Why won't this be valid any more? I'm only aware of this not working for containers that use ubi-micro (since there's no dnf/microdnf available).

komish:
I'm holding this for the moment. We're making some decisions on how we want to handle this particular check moving forward, and the current ideology differs a bit from how this is implemented, so we think it best to just hold here for now.

internal/policy/container/has_modified_files.go
```diff
@@ -24,140 +28,158 @@ var _ check.Check = &HasModifiedFilesCheck{}
...
-			}
-			// also add it to the overall list of files.
-			files = append(files, layerFiles)
-			wg.Done()
-		}()
-		// add paths to the pathChan
-		err = untar(pathChan, r)
+			return nil, nil, fmt.Errorf("unable to retrieve diff id for layer: %w", err)
+		}
+		layerID := layerIDHash.String()
```
acornett21, line 88: Question: This is the `sha256` value correct? not like it's indecency (0, 1, 2,....) correct?

internal/policy/container/has_modified_files.go
```diff
@@ -24,140 +28,158 @@ var _ check.Check = &HasModifiedFilesCheck{}
...
+	disallowedModifications := map[string]string{}
+	for idx, layerID := range layerIDs {
+		ref := packageFiles[layerID]
+		// if ref.HasRPMDB {
+		// 	// TODO: bpc
+		// 	// Something else needs to be other than just skipping because we have an RPM db in this layer.
+		// 	// It works for the *first* layer with an RPM db, since everything would be flagged, but we need to
+		// 	// figure out WHY a new RPM db is present.
+		// 	continue
+		// }
```
acornett21, line 154: Is this still valid to be commented out? I think you covered this in other logic where you are finding the RPM's.

internal/policy/container/has_modified_files.go
```diff
@@ -24,140 +28,158 @@ var _ check.Check = &HasModifiedFilesCheck{}
...
+				if previousPackageVersion != currentPackageVersion {
+					// This appears like an update. This is allowed.
+					continue
+				}
+				modFound = true
+				disallowedModifications[modifiedFile] = layerID
 			}
 		}
+		if modFound {
+			logger.V(log.DBG).Info("found disallowed modification in layer")
```
acornett21, line 179: Does logging the layer buy us anything here? Or are we logging the layer/modification/etc elsewhere?

internal/policy/container/has_modified_files.go
```diff
@@ -180,69 +202,242 @@ func (p HasModifiedFilesCheck) Metadata() check.Metadata {
...
+	pkgNameList := make(map[string]struct{}, len(pkgList))
+	for _, pkg := range pkgList {
+		pkgNameList[fmt.Sprintf("%s-%s-%s", pkg.Name, pkg.Version, pkg.Release)] = struct{}{}
+	}
+	return pkgNameList
+}
 
-	for {
-		header, err := tr.Next()
+// FindRPMDB attempts to extract a valid RPMDB from layers in the order
```
acornett21, line 213:
```suggestion
// findRPMDB
```

acornett21, line 213:
```suggestion
// findRPMDB attempts to extract a valid RPMDB from layers in the order
```
Nit

acornett21: This is like this in other places too, not sure if it's the linter doing that, or if that's how we want it.

internal/policy/container/has_modified_files.go
```diff
@@ -180,69 +202,242 @@ func (p HasModifiedFilesCheck) Metadata() check.Metadata {
...
+		"etc": {},
+		"var": {},
+		"run": {},
+	}
 
-		// if the header is nil, just skip it (not sure how this happens)
-		case header == nil:
-			continue
+	for k := range excl {
+		if strings.HasPrefix(s, filepath.Clean(k+"/")) || k == s {
```
acornett21, line 239: Is the `+"/"` necessary? Won't clean take care of the case where the trailing `/` is left off? Maybe I read clean incorrectly. Or you are trying to force a slash to insure clean doesn't error, since `//` will be taken care of anyways.

internal/policy/container/has_modified_files.go
```diff
@@ -180,69 +202,242 @@ func (p HasModifiedFilesCheck) Metadata() check.Metadata {
...
-		target := header.Name
+	return false
+}
 
-		// the following switch could also be done using fi.Mode(), not sure if there
-		// a benefit of using one vs. the other.
-		// fi := header.FileInfo()
+// PathIsExcluded checks if s is excluded explicitly as written.
+func pathIsExcluded(ctx context.Context, s string) bool {
+	excl := map[string]struct{}{
```
acornett21, line 251: Do we need `var` and `/var` here as well?

internal/engine/engine.go
```diff
@@ -224,7 +224,7 @@ func (c *CraneEngine) ExecuteChecks(ctx context.Context) error {
 		checkElapsedTime := time.Since(checkStartTime)
 
 		if err != nil {
-			logger.WithValues("result", "ERROR", "err", err).Info("check completed", "check", check.Name())
+			logger.WithValues("result", "ERROR", "err", fmt.Sprintf("%v", err)).Info("check completed", "check", check.Name())
```
komish, line 227: Should be able to just call the Error() method here.

```suggestion
			logger.WithValues("result", "ERROR", "err", err.Error()).Info("check completed", "check", check.Name())
```

internal/policy/container/has_modified_files.go
```diff
@@ -24,140 +28,158 @@ var _ check.Check = &HasModifiedFilesCheck{}
...
+	layerIDs, packageFiles, err := p.gatherDataToValidate(ctx, imgRef, fs)
 	if err != nil {
 		return false, fmt.Errorf("could not generate modified files list: %v", err)
 	}
-	return p.validate(ctx, packageFiles)
+	return p.validate(ctx, layerIDs, packageFiles)
 }
 
-// getDataToValidate returns the list of files (packageFilesRef.PackageFiles)
+// gatherDataToValidate returns the list of files (packageFilesRef.PackageFiles)
```
komish, line 53:
```suggestion
// gatherDataToValidate returns the list of files (packageFilesRef.LayerFiles)
```

internal/policy/container/has_modified_files.go
komish: Correct. It's the hash value, not an index value.

https://pkg.go.dev/github.com/google/go-containerregistry@v0.13.0/pkg/v1#Layer

internal/policy/container/has_modified_files.go
```diff
@@ -24,140 +28,158 @@ var _ check.Check = &HasModifiedFilesCheck{}
...
+			return nil, nil, err
+		}
 
-	return &packageFilesRef{files, packageFiles}, nil
-}
+		found, pkgList := findRPMDB(ctx, layer)
+		if !found {
+			logger.V(log.TRC).Info("could not find rpm database")
+			if idx > 0 {
+				// Just make this is the same as last layer, since the RPM db was not modified
```
komish, line 107: This is a bit confusing. We're suggesting that the Layer packages are the same as the previous layer in cases where we haven't found the RPMDB? So... empty? I'm finding it a bit hard to follow what this solves.

internal/policy/container/has_modified_files.go
```diff
@@ -180,69 +202,242 @@ func (p HasModifiedFilesCheck) Metadata() check.Metadata {
...
+	for _, pkg := range pkgList {
+		pkgNameList[fmt.Sprintf("%s-%s-%s", pkg.Name, pkg.Version, pkg.Release)] = struct{}{}
+	}
+	return pkgNameList
+}
 
-	for {
-		header, err := tr.Next()
+// FindRPMDB attempts to extract a valid RPMDB from layers in the order
+// they are provided. If found is not set to true, foundIndex and pkglist should
```
komish, line 214: NIT:
```suggestion
// they are provided. If found is false, foundIndex and pkglist should
```

internal/policy/container/has_modified_files.go
bcrochet: Oh. Meant to remove that.

internal/policy/container/has_modified_files.go
```diff
@@ -24,140 +28,151 @@ var _ check.Check = &HasModifiedFilesCheck{}
...
+		files, err := generateChangesFor(layer)
+		if err != nil {
+			return nil, nil, err
+		}
 
-	return &packageFilesRef{files, packageFiles}, nil
-}
+		found, pkgList := findRPMDB(ctx, layer)
+		if !found {
+			logger.V(log.TRC).Info("could not find rpm database")
```
komish, line 105: This runs on every layer right? I think adding layer context might be useful here.

```suggestion
			logger.V(log.TRC).Info("could not find rpm database in layer", "layer", layerID)
```

That code may not be correct. Making the assumption that this accepts keys/values. Either way, you get the idea.

internal/policy/container/has_modified_files.go
komish: +1 might as well log the layer.

bcrochet: Not really. I took this straight from what @komish PoC'd. If we want to keep it, I should probably add the actually disallowed modification.

internal/policy/container/has_modified_files_test.go
```diff
@@ -0,0 +1,214 @@
...
+				pkgSecondLayer.LayerPackageFiles = pkgSecondLayerPackageFiles
+				pkgs["secondlayer"] = pkgSecondLayer
+			})
+			It("should pass validate", func() {
+				ok, err := hasModifiedFiles.validate(context.Background(), layers, pkgs)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(ok).To(BeTrue())
+			})
+		})
+		When("a package is removed", func() {
```
acornett21, line 127: Do we want a test for adding a package?

internal/policy/container/has_modified_files.go
bcrochet: I changed it to keep the comment, but also match the actual function name.

internal/policy/container/has_modified_files.go
bcrochet: @komish wrote this part. IIUC, it's so that no matter what, at least '/' will be there. If k was cleaned completely, it would return '.'

internal/engine/engine.go
bcrochet: Done

internal/policy/container/has_modified_files_test.go
```diff
@@ -0,0 +1,214 @@
...
+					DirNames:   []string{dirname},
+				},
+			}
+
+			badPkgList = []*rpmdb.PackageInfo{
+				{
+					BaseNames:  []string{basename},
+					DirIndexes: []int32{dirindex},
+					DirNames:   []string{dirname, "extra"},
+				},
```
acornett21, line 188: I get that the intent is to not have a `BeforeEach` in each test, but I'm not sure I love this style where we setup all the data, then only use one half in one test and other half in another. Maybe it's okay in this use case since we are testing individual methods, and not an entire flow.

internal/policy/container/has_modified_files.go
bcrochet: Var is caught in the directoryIsExcluded

internal/policy/container/has_modified_files.go
bcrochet: Thanks. I'll fix.

internal/policy/container/has_modified_files.go
bcrochet: It should just copy the rpm info from one layer to the next, once an RPMdb has been found. This keeps from having to track "last RPM db"

internal/policy/container/has_modified_files.go
bcrochet: Sure. I'll add that context.

internal/policy/container/has_modified_files_test.go
bcrochet: I think we already have that?

acornett21: Maybe that's covered in the modified and update. I was thinking we might have a test where we add a 3rd layer, but if that doesn't buy anything, maybe it's not needed.

internal/policy/container/has_modified_files.go
```diff
@@ -180,69 +196,242 @@ func (p HasModifiedFilesCheck) Metadata() check.Metadata {
...
+// findRPMDB attempts to extract a valid RPMDB from layers in the order
+// they are provided. If found is false, pkglist should
+// be disregarded as any value there will be invalid.
+func findRPMDB(ctx context.Context, layer v1.Layer) (found bool, pkglist []*rpmdb.PackageInfo) {
+	logger := logr.FromContextOrDiscard(ctx)
+	var err error
+	pkglist, err = extractRPMDB(ctx, layer)
+	if err == nil {
+		id, _ := layer.Digest()
+		logger.V(log.TRC).Info("findRPMDB found an RPM db", "layer", id)
```
komish, line 216: Something weird is happening with the quoting here. I'm seeing. Is `id` a struct/interface? Do you need to `id.String()`?
```
time="2023-03-07T13:37:23-06:00" level=trace msg="findRPMDB found an RPM db" layer="\"sha256:0ac9bf1c61484d864e69dc9405a28bfcd479493b343029402cb6779a4b163cf5\""
```

internal/policy/container/has_modified_files.go
komish: Confirmed ^

komish: Alternatively, could we use `normalize`

internal/policy/container/has_modified_files.go
komish: I'd have to look closer, but the PoC definitely spat out a list of things we considered to be disallowed after doing this evaluation, and the layer.

```json
13:54:21 ~/projects/hasModifiedFilesTest/hasmodfiles-run-quay_io_komish_deleteme_test_mod_files_0_0_1
 ⤿ cat disallowedmods.json 
{
    "usr/bin/sha256sum": "sha256:2997d243150350c3d23695cf50cbfcbfa291a18553b7eaff006799e4dc718acf"
}
```

internal/policy/container/has_modified_files_test.go
bcrochet: It's covered in the main test. If you look at the data, you will see that a package is "added" between layer one and two.

internal/policy/container/has_modified_files_test.go
bcrochet: I see what you mean. Yeah, I can fix that.

internal/policy/container/has_modified_files.go
bcrochet: Fixed.

internal/policy/container/has_modified_files.go
bcrochet: This should be the case again.

internal/policy/container/has_modified_files.go
```diff
@@ -24,140 +28,149 @@ var _ check.Check = &HasModifiedFilesCheck{}
...
-// of when this should occur is in cases where the image was built FROM scratch, such as ubi-micro. This is required
-// to ensure we don't attempt to validate against an empty layer.
-func (p *HasModifiedFilesCheck) dropFirstLayerIfEmpty(files [][]string) ([][]string, bool) {
-	var dropped bool
-	if len(files[0]) == 0 {
-		files = files[1:] // shift the empty layer out.
-		dropped = true
+			// If it's the first layer, just make the pkgList empty.
+			pkgList = make([]*rpmdb.PackageInfo, 0)
+		}
```
komish, line 121: I'm still not understanding this block. We're checking _every_ layer for an RPMDB? I assume this is because we generate a package list again, make sure that all packages have the right info, etc. I guess my concern is that the log output ends up looking like this.

```
time="2023-03-08T12:16:05-06:00" level=trace msg="findRPMDB found an RPM db" layer="sha256:296e14ee24149e14d573a1fbf5b5a625c7bb0cc22f5b2a8b180e833258187948"
time="2023-03-08T12:16:05-06:00" level=trace msg="could not find rpm database in layer" layer="sha256:8f26704f753cbe8b20afc01e97a4cb652befe26791b4d33193afc99ce544ebf5"
time="2023-03-08T12:16:05-06:00" level=trace msg="could not find rpm database in layer" layer="sha256:96915ad71c5c10538d1f3f5dda36a8d90dc8151805acb3b263406cb9e50036ff"
time="2023-03-08T12:16:05-06:00" level=trace msg="could not find rpm database in layer" layer="sha256:d6f7de083499e5152f183f885248de0ec370972fd9956bf3c8e48091984e750a"
time="2023-03-08T12:16:05-06:00" level=trace msg="could not find rpm database in layer" layer="sha256:1b8d55546b61cab3c1ecccdd581f3a184fb6c8b1916b254a406c6a395b0e880d"
time="2023-03-08T12:16:05-06:00" level=trace msg="could not find rpm database in layer" layer="sha256:e979289ce7d578b603bada6f7eeca5fc9f13b55b43d631e1b9626166ffcc778d"
time="2023-03-08T12:16:05-06:00" level=trace msg="findRPMDB found an RPM db" layer="sha256:63b8c90a3decf1db1660fbf451af06ce62ba843d46a77ce2dc0a0c8faeac9e50"
time="2023-03-08T12:16:05-06:00" level=trace msg="could not find rpm database in layer" layer="sha256:c8ec39fea4bb24e466b6fc2c5919b0d3f670097acb44635768fa388f0bcd9ac9"
time="2023-03-08T12:16:05-06:00" level=trace msg="could not find rpm database in layer" layer="sha256:0f39727bac0d55d024d8524e66a68a621eca212e8a6718f2accb3452ada55506"
time="2023-03-08T12:16:06-06:00" level=trace msg="could not find rpm database in layer" layer="sha256:7eb228728e47c149e57edaf5e822d6814d9be390c1535beabe32369067373390"
time="2023-03-08T12:16:07-06:00" level=trace msg="could not find rpm database in layer" layer="sha256:bc3e54ef33b1f796a97e73040c1323daedf2135ee559369ef3ba726d2b8a9f78"
time="2023-03-08T12:16:07-06:00" level=debug msg="found disallowed modification in layer" file=usr/sbin/nginx layer="sha256:c8ec39fea4bb24e466b6fc2c5919b0d3f670097acb44635768fa388f0bcd9ac9"
time="2023-03-08T12:16:07-06:00" level=debug msg="found disallowed modification in layer" file=usr/sbin/nginx-debug layer="sha256:c8ec39fea4bb24e466b6fc2c5919b0d3f670097acb44635768fa388f0bcd9ac9"
```

What's throwing me is the subsequent RPMDB being found. In trace logs, this doesn't help me debug unless I already know the implementation inside and out. Maybe a nit. I'm not sure.

internal/policy/container/has_modified_files.go
```diff
@@ -24,140 +28,149 @@ var _ check.Check = &HasModifiedFilesCheck{}
...
-			}
-			// also add it to the overall list of files.
-			files = append(files, layerFiles)
-			wg.Done()
-		}()
-		// add paths to the pathChan
-		err = untar(pathChan, r)
+			return nil, nil, fmt.Errorf("unable to retrieve diff id for layer: %w", err)
+		}
+		layerID := layerIDHash.String()
```
komish, line 89: This is a DiffID, not a LayerID. The layer value we spit out from here (on line 94) doesn't exist in an image's RootFS.Layers array.

internal/policy/container/has_modified_files.go
bcrochet: Trace is for developers. Right? I wouldn't want a partner to have to use trace in order to know what failed. It's most certainly an implementation detail. 

But to answer your initial question... Yes. Every layer is being check for an RPM db.

internal/policy/container/has_modified_files.go
bcrochet: This has been changed to use the LayerID.

komish (approved):
Overall

/lgtm

The pending thing is we're failing an edge case, but we've discussed it internally and @bcrochet will look to resolve or add a case around it.

bcrochet:
Closing to work on this in a branch.

PR #856, opened 2022-12-15
Merge newly-rebased lib into main

komish:
Just three things:
- Rebased Lib onto Main
- Fixed a bug in our logging where we were inadvertently not emitting variables because we didn't use `sprintf` https://github.com/redhat-openshift-ecosystem/openshift-preflight/commit/f5852ef50be9ac4cd19b77fbdae19b6395a09e8f
- Updated the LIBRARY.md doc to comply with the new logging changes. https://github.com/redhat-openshift-ecosystem/openshift-preflight/commit/2899cb0cfa6f1237bc61044262a37c5073832bd9

I left the original **lib** branch intact so we can use it as a reference, if needed. The first 6 commits here should be aligned with it, barring merge conflicts (which were mostly go mod related)


komish:
golangci-lint was upset about my import ordering so I fixed that and made it its own commit, simply because it touched 17 files :roll_eyes:

komish:
Re E2E Failures: Prow seems to be failing because some logic is creating a directory `artifacts` at the root of the source, and we already have that directory as a library.

sebrandon1 (approved):

.gitignore
```diff
@@ -31,8 +31,6 @@ results-junit.xml
 # E2E customzied env log file.
 sgol.txt
 
-# Testing dirs
-/artifacts
 # E2E customized env artifacts dir
 stcafitra/
```
skattoju, line 35: 🤨

acornett21 (approved):

skattoju (approved):
/lgtm

komish:
:partying_face:

sebrandon1:
Great to see this merged into main!  Thanks!

PR #858, opened 2022-12-16
Add Optional Restricted Network Check

komish:
/do-not-merge

This PR is built off of the lib-->main rebase because that rebase has moved a majority of the library code into an internal package. This PR does the same for new code to avoid rework.

---

This PR adds a check that attempts to review a CSV for hints that the developer has implemented guidelines to support clusters that have restricted networks. This boils down to:

- Does the operator have related images defined in their CSV spec. There's already a related image check, so we do not replicate that work, but instead just do a soft check to make sure there are values in that section.
- Does this operator have the "disconnected" infrastructure-feature annotation.
- Does this operator include environment variables in any of its deployment containers that are prefixed with RELATED_IMAGE_, indicating that the developer intends to pass through (to the controller) what images it should use for its managed workloads.

There's no technical way to validate that the third bullet results in a workload that respects this value. With that being said, OperatorSDK and operator-manifest-tools will automatically pin tagged images to their digest values if they're prefixed with this string, and so it makes sense to at least look for them to try to increase the confidence that things will work as expected. 

This check is marked as **optional**. It may fail, but that failure does not block overall results, or reflect as a failed check.

---

Note that a majority of this check's logic was thrown into a library (internal/csv) just to make it easier to test.


acornett21:
My comments ended up on the commit and not this PR. :(

komish:
https://github.com/redhat-openshift-ecosystem/openshift-preflight/commit/357e60cc429bf2e4788ea2d6b4c1dee104928410 address golangci-lint complaints

komish:
The 2.5% coverage decrease is _also_ coming from the original commits related to the rebase - not this PR.

komish:
Removing the hold on this as lib-->main has now merged, and this should no longer reflect those changes in this PR's commit history. Just the relevant changes.

internal/csv/csv.go
```diff
@@ -0,0 +1,97 @@
...
+	"github.com/operator-framework/operator-manifest-tools/pkg/imagename"
+	appsv1 "k8s.io/api/apps/v1"
+)
+
+const InfrastructureFeaturesAnnotation = "operators.openshift.io/infrastructure-features"
+
+// SupportsDisconnected accepts a stringified list of supported features
+// and returns true if "disconnected" is listed as a supported feature.
+//
+// E.g. '["disconnected"]'.
```
bcrochet, line 17: Big super small nit: This should probably be backticks, and not single quotes, to signify a JSON literal.

internal/policy/operator/restricted_network_aware.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	// If the CSV does not claim to support disconnected environments, there's no reason to check that it followed guidelines.
+	if !libcsv.HasInfrastructureFeaturesAnnotation(csv) {
+		logger.Info("warning: this operator does not have the infrastructure-features annotation, unable to determine if it supports disconnected environments.")
+		return false, nil
+	}
+
+	if !libcsv.SupportsDisconnected(csv.Annotations[libcsv.InfrastructureFeaturesAnnotation]) {
+		logger.Info("warning: the infrastructure-features enabled for this operator did not include the \"disconnected\" identifier.")
```
bcrochet, line 47: Why is this a warning? If I'm not expecting to have a disconnected operator, I'd be confused by this. I'd be ok with the log message, just don't think it should be specified as "warning:"

internal/policy/operator/restricted_network_aware.go
```diff
@@ -0,0 +1,99 @@
...
+	// All related images must be pinned. No tag references.
+	if !libcsv.RelatedImagesArePinned(csv.Spec.RelatedImages) {
+		logger.Info("warning: a related image is not pinned to a digest reference of the same image, and this is required.")
+		return false, nil
+	}
+
+	// Some environment variables should be passed into the environment using the RELATED_IMAGE_ prefix.
+	// This isn't the only way to pass this kind of information to the controller, but it is the suggested way in our
+	// documentation.
+	deploymentSpecs := make([]appsv1.DeploymentSpec, len(csv.Spec.InstallStrategy.StrategySpec.DeploymentSpecs))
```
bcrochet, line 66: You need to add a 0 in the middle. This creates a slice with capacity len(DeploymentSpecs) AND length len(DeploymentSpecs). Since you are using append, you want a capacity of len(DeploymentSpecs) and a length of 0. It probably hasn't caused an issue because the zero value for DeploymentSpec is not causing a problem, but you will end up with a slice that is populated with half zero value DeplymentSpecs at the beginning, and the second half being the ones you are appending.

bcrochet (changes requested):
A nit, a question, and a bug.

internal/csv/csv.go
komish: I don't think this is the case in the YAML manifests? If we wanted to represent the value in pure-go, it'd probably be 

```
`'["disconnected"]'`
```

But in the yaml, it's just single quotes. Example: https://github.com/redhat-openshift-ecosystem/community-operators-prod/blob/544d90d878a0c1dc7b194f004e00f57802578279/operators/pcc-operator/0.2.0/manifests/pcc-operator.v0.2.0.clusterserviceversion.yaml#L63

internal/policy/operator/restricted_network_aware.go
komish: That's fair. Not sure what I was thinking re: the warning. Thinking it may have been a copy-pasta. 

Now that I think about it, should I log this at all? :thinking:

internal/policy/operator/restricted_network_aware.go
komish: Derp, you're right - should be fixed.

internal/csv/csv.go
bcrochet: I see. Interesting. Like I said, super small nit. But I didn't realize that's how the field was actually represented in the yaml. It seems to be so that it is escaped, and not parsed as an array in the unmarshalling. Carry on.

internal/policy/operator/restricted_network_aware.go
bcrochet: I think it should be logged. Just not as "warning:"..

Let's think this through. As a user of a normal environment, if I see this message, what do I think? "Am I supposed to have 'disconnected' there?"

As a disconnected user, if I *don't* see this message, will I know that it's not working? 

I think the compromise might be this:

the infrastructure-features enabled for this operator did not include the "disconnected" identifier. If the operator is not targeted for a disconnected environment, one can safely ignore this message.

acornett21: I had similar thoughts on all of this logic as well (and commented on the commit and not the PR, no idea where that is at anymore), we are catering to the few, to confuses the masses.

bcrochet: @acornett21 I totally get where you are coming from there. I think the important thing here is to make sure that we fail as fast as possible, and communicate to those that don't care about it... "You can ignore this". So, we just have to make sure that the first check is the one that fails the fastest. 

Also, it might make sense to bookend these messages. Log at the beginning of the function "You can ignore this if you are not a disconnected operator", and defer logger.Info("Ignoring complete")

komish: @bcrochet there's an initial phase that checks that the operator has the annotation and that it contains the expected value. We don't actually check for readiness unless you pass both of those checks.

To that end, I've added "safe to ignore" to both of their log lines, and removed all the `warning:` text in the latest force-push https://github.com/redhat-openshift-ecosystem/openshift-preflight/commit/2e4f1be46c633407da33dfef9325245895f35078

acornett21: @bcrochet All that makes sense and I think @komish is doing that, my comment was more about the verbiage specifically. I think the latest commit clears/cleans things up nicely.

acornett21 (approved):

internal/csv/csv.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+// RelatedImageReferencesInEnvironment returns a slice of environment variables that are
+// prefixed with RELATED_IMAGE_, which is the typical way to push these values into
+// the controller's environment.
+func RelatedImageReferencesInEnvironment(deployments ...appsv1.DeploymentSpec) []string {
+	values := []string{}
+
+	for _, depl := range deployments {
+		for _, container := range depl.Template.Spec.Containers {
```
bcrochet, line 77: Nit (you can ignore), but to DRY this up, this inner loop could be in a function that is passed each of the container slices.

bcrochet (approved):
Just a small nit. I'll drop an approve in case you want to ignore it.

internal/csv/csv.go
komish: Good catch. Fixed!

bcrochet (approved):
/lgtm

Issue #827, opened 2022-11-10
Artifacts folder


lucacome:
## Bug Description
Every time I run `preflight check container` I need to change the artifacts folder with `--artifacts=`.

```
Error: could not write file to artifacts directory: /tmp/artifacts/results.json already exists
```

## Version and Command Invocation
```
$ preflight --version
preflight version 1.4.2 <commit: f9cff772837132149df69f8ae251d3caf81c49ac>
```

## Steps to Reproduce:
(How can we reproduce this?)
1) Run `preflight check container` with an image
2) Run `preflight check container` with an image a second time
3) See the error `Error: could not write file to artifacts directory`

## Expected Result
The ideal would be for me to not have to specify the artifacts folder. I'm not even sure why that's needed.

## Actual Result
I have to change the folder in the command line for every architecture of every image that I want to certify.

acornett21:
Hi @lucacome This is working as designed, since we want to make sure the user clears out the artifacts folder for before each run to ensure we are using the latest information from the most recent run. Preflight by default writes to `./artifacts` by defaults so if you clean that folder each run this shouldn't be an issue.

Preflight uses this folder to save off information gathered in the checks/policies that are then later used during the submission process to send to the backend system.

lucacome:
Hi @acornett21 thanks for the explanation.

Maybe I'm missing some context, but I'm not sure how the artifacts folder is useful to the end user (like me).
Let's say I run `preflight check` and the outputs of the check are saved in `artifacts`. Then if I want to submit those results, since I don't have a specific command to do that without re-running the check, I just do `preflight check --submit` and get the error about the files already existing.
Another scenario is the one I described above about submitting multiple images/architectures (even without the two steps, just using the `--submit` flag right away). I would have to either change the folder name or delete it all together every time I run the command.

So in the end I'm not sure what benefit I'm getting from the artifacts folder (given that the `results.json` is also in the output of the command). Maybe I'm missing something and I'm supposed to use the files in the folder?

IMHO, a better experience (at least for me) is that if somebody uses `--artifacts string` means that they care about those files and should behave the same way it does today, but if they don't pass the flag it should default to a random folder in `/tmp`.

komish:
> This change is now creating support cases.

@jfrancin if able, could you describe/paraphrase what the support case's reported issue is, and what the recommended resolution is (most commonly)? I ask simply because it would seem the issue preflight detects, and its resolution should be derived from the error message it returns.

[...]

PR #866, opened 2023-01-10
allow results and log writing to overwrite existing content on re-run of preflight

komish:
Workaround for #827 

In [version 1.4.0](https://github.com/redhat-openshift-ecosystem/openshift-preflight/blob/1.4.0/certification/artifacts/artifacts.go#L51), Preflight would overwrite existing results.json, etc. within the artifacts directory on re-execution. This behavior changed in version 1.4.1 with the introduction of Afero which is causing some friction for users.

This PR should return the original behavior back until we decide what we want to do re: artifacts writing in the future.


bcrochet (approved):
/lgtm

acornett21 (approved):

PR #877, opened 2023-01-30
fix deployable_by_olm to prioritize installmodes

acornett21:
It seemed this break statement was overlooked, and is causing an issue with some operators. Adding this back so we prioritize `OwnNamespace` over `SingleNamespace` again.


jomkz (approved):
/lgtm

skattoju (approved):
/lgtm

internal/policy/operator/deployable_by_olm.go
```diff
@@ -370,6 +370,7 @@ func (p *DeployableByOlmCheck) generateOperatorGroupData(ctx context.Context, op
 	for _, v := range prioritizedInstallModes {
 		if operatorData.InstallModes[v].Supported {
 			installMode = operatorData.InstallModes[v].Type
+			break
```
itroyano, line 373: IMO the missing break could be intended.   in case the CSV declares more than one install mode as supported - this picks the "highest" one.
ref - https://olm.operatorframework.io/docs/advanced-tasks/operator-scoping-with-operatorgroups/#targetnamespaces-and-their-relationship-to-installmodes

The issue leading to this PR is IBM ODF Operator, who declare both OwnNamespace and SingleNamespace support, but do not actually support SingleNamespace (operator fails).

acornett21: @itroyano This break was missed in a refactor, and `priortizedInstallModes` is an ordered list.

itroyano: That's ok, Opcap covers the all install modes scenario.
If this is a bug I would consider it a "good bug".  failed in the pipeline / testing rather than getting to the partner's actual customer and failing there.
if they mark "true" for an InstallMode, it means it was tested and works. so theoretically it shouldn't matter if Preflight picks the first mode they set "true", or the last.

itroyano: no strong objections for the fix though.

bcrochet: While I get where you are coming from @itroyano, it's not preflight's job to find bugs in the operator.

bcrochet (approved):
/lgtm

komish (approved):
/lgtm

PR #883, opened 2023-02-03
configure the insecure flag to allow access to registries with self-signed certificate

komish:
There's a bit of a behavioral difference in the crane library execution compared to its CLI that causes us to have a bug in how we handle our insecure flag. See https://github.com/google/go-containerregistry/issues/1553 for full context.

In effect, we allowed access to registries available via HTTP, but not HTTPS registries with untrusted certificates.

In testing, I also found a bug where we didn't fully initialize the the CraneEngine by passing it the insecure flag when received by the caller.

This PR resolves both issues. 

Signed-off-by: Jose R. Gonzalez <jrg@redhat.com>

### DIFF
--- a/internal/engine/engine.go
+++ b/internal/engine/engine.go
@@ -5,6 +5,7 @@ import (
 	"bytes"
 	"context"
 	"crypto/md5"
+	"crypto/tls"
 	"encoding/json"
 	"fmt"
 	"io"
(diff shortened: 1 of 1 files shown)


acornett21 (approved):

skattoju (approved):
/lgtm

bcrochet (approved):
/lgtm

PR #891, opened 2023-02-16
adding CLoseWithError to crane.Export for cases when reader returns EOF

acornett21:
This fixes an error where the reader does not close on an EOF. This is similar to how crane handles the export. There are also a ton of different ways to solve this. 

For example if we want to keep the waitgroup, we could add a count to that and wrap the untar as well.

### DIFF
--- a/internal/engine/engine.go
+++ b/internal/engine/engine.go
@@ -18,7 +18,6 @@ import (
 	"regexp"
 	"sort"
 	"strings"
-	"sync"
 	"time"
 
 	"github.com/go-logr/logr"
@@ -150,28 +149,27 @@ func (c *CraneEngine) ExecuteChecks(ctx context.Context) error {
 
 	// export/flatten, and extract
 	logger.V(log.DBG).Info("exporting and flattening image")
-	wg := sync.WaitGroup{}
-	wg.Add(1)
 	r, w := io.Pipe()
 	go func() {
-		defer w.Close()
 		logger.V(log.DBG).Info("writing container filesystem", "outputDirectory", containerFSPath)
-		err = crane.Export(img, w)
-		if err != nil {
-			// TODO: Handle this error more effectively. Right now we rely on
-			// error handling in the logic to extract this export in a lower
-			// line, but we should probably exit early if the export encounters
-			// an error, which requires watching multiple error streams.
-			logger.Error(err, "unable to export and flatten container filesystem")
-		}
-		wg.Done()
+
+		// Close the writer with any errors encountered during
+		// extraction. These errors will be returned by the reader end
+		// on subsequent reads. If err == nil, the reader will return
+		// EOF.
+		w.CloseWithError(crane.Export(img, w))
 	}()
 
 	logger.V(log.DBG).Info("extracting container filesystem", "path", containerFSPath)
 	if err := untar(ctx, containerFSPath, r); err != nil {
 		return fmt.Errorf("failed to extract tarball: %v", err)
 	}
-	wg.Wait()
+
+	// explicitly discarding from the reader for cases where there is data in the reader after it sends an EOF
+	_, err = io.Copy(io.Discard, r)
+	if err != nil {
+		return fmt.Errorf("failed to drain io reader: %v", err)
+	}
 
 	reference, err := name.ParseReference(c.Image)
 	if err != nil {
--- a/internal/engine/engine_test.go
+++ b/internal/engine/engine_test.go
@@ -1,6 +1,8 @@
 package engine
 
 import (
+	"archive/tar"
+	"bytes"
 	"context"
 	"errors"
 	"fmt"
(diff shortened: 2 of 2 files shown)


komish (commented):
Strictly speaking, we do not have to both remove the waitgroup configuration AND close the reader on completion of tar. We can do both, but if we want to reduce the scope of this PR, we can just convert the Reader to ReadCloser (as you've done, and defer closing the reader in that untar function _because_ io.Pipe is synchronous.

acornett21:
> Strictly speaking, we do not have to both remove the waitgroup configuration AND close the reader on completion of tar. We can do both, but if we want to reduce the scope of this PR, we can just convert the Reader to ReadCloser (as you've done, and defer closing the reader in that untar function _because_ io.Pipe is synchronous.

Correct, we really can do 'anything' in this PR like I mentioned. We can add another counter to the wg as well, and solve this with a `reader.Close()` and it's own `defer` for the untar as well. Whatever direction the team is more comfortable with, we can go that route.

komish:
**Note for reviewers:**

Finally diagnosed this - the issue here is that the image layers that encountered this hanging behavior contained padding after the EOF is sent. As a result, The PipeWriter (i.e. what's passed to crane.Export), being synchronous/unbuffered, is waiting for reads on the PipeReader (i.e. what's passed to untar) which never come because the `untar` function received its EOF and exited its function scope. The test image that was provided has an additional ~6600 bytes in the PipeReader.

Because we use a WaitGroup, we wait indefinitely for reads that will never come. 

The solutions are:
1) Don't use the WaitGroup (technically unnecessary because of the synchronous nature of these reads/writes)
2) Drain the remaining bytes from the PipeReader (e.g. after the untar completes) to force the Writer to close out and complete. E.g. `bytesRead, err := io.ReadAll(r)`.
3) Use an io.ReadCloser and close it once the `untar` function reaches EOF (this is what Adam has in this PR).

Reference: https://github.com/golang/go/issues/53110

bcrochet:
>     2. Drain the remaining bytes from the PipeReader (e.g. after the untar completes) to force the Writer to close out and complete. E.g. `bytesRead, err := io.ReadAll(r)`.

One could also do `_, err := io.Copy(io.Discard, r)`

internal/engine/engine.go
```diff
@@ -150,28 +149,21 @@ func (c *CraneEngine) ExecuteChecks(ctx context.Context) error {
...
+		// extraction. These errors will be returned by the reader end
+		// on subsequent reads. If err == nil, the reader will return
+		// EOF.
+		w.CloseWithError(crane.Export(img, w))
 	}()
 
 	logger.V(log.DBG).Info("extracting container filesystem", "path", containerFSPath)
 	if err := untar(ctx, containerFSPath, r); err != nil {
 		return fmt.Errorf("failed to extract tarball: %v", err)
 	}
```
bcrochet, line 166: So, instead of using the ReadCloser, we would just drain the reader here?

`_, err := io.Copy(io.Discard, r)`

That seems it would limit the scope of this change even further? 

I concur that the WaitGroup is extraneous here.

internal/engine/engine.go
```diff
@@ -426,6 +424,43 @@ func untar(ctx context.Context, dst string, r io.Reader) error {
...
 }
 
+// writePaddedTarball writes a tar archive to out with filename containing contents at the base path
+// with extra bytes written at the end of length extraBytes+1.
+// note: this should only be used as a helper function in tests
+func writePaddedTarball(out io.Writer, contents []byte, filename string, extraBytes uint) error {
```
bcrochet, line 430: This should be in the _test.go

bcrochet (approved):
/lgtm

internal/engine/engine_test.go
```diff
@@ -277,3 +356,40 @@ var _ = Describe("Check Name Queries", func() {
...
+	if err != nil {
+		return err
+	}
+
+	_, err = io.Copy(tw, bytes.NewReader(contents))
+	if err != nil {
+		return err
+	}
+
+	if extraBytes > 0 {
```
komish, line 385: This conditional seems out of place. If extraBytes `<=` 0, then we don't write any padding which defeats the purpose of this function, no?

internal/engine/engine_test.go
```diff
@@ -138,6 +148,75 @@ var _ = Describe("Execute Checks tests", func() {
...
+				layer := static.NewLayer(buf.Bytes(), types.MediaType("application/vnd.docker.image.rootfs.diff.tar"))
+				img, err := mutate.AppendLayers(empty.Image, layer)
+				Expect(err).ToNot(HaveOccurred())
+
+				src = fmt.Sprintf("%s/test/crane", u.Host)
+
+				err = crane.Push(img, src)
+				Expect(err).ToNot(HaveOccurred())
+			})
+			It("should just hang forever err and hash mean nothing", func() {
```
komish, line 167: Does this really hang forever?

internal/engine/engine_test.go
komish: I see that you're using this function to write a tarball that isn't padded as well. In that case, maybe just remove the addition of 1 extra byte, rename the function, update the comment, etc.

internal/engine/engine_test.go
acornett21: No it does not, I wrote this test against `main`, which would be true for `main` but not with the fix. I'll double check the asserts again so they are correct.

PR #897, opened 2023-02-21
adding a test to write an GNU tar layer with padded zeros at the end

acornett21:
This is just a test pr against main to show prove that our untar logic will hang forever if given an image with a padded tar layer


skattoju (approved):
/lgtm

acornett21:
Closing this PR, since this was just to show the test written was a valid test. And failed against `main`.

PR #898, opened 2023-02-22
updating the readme to add clarity to cert docs and developer testing

acornett21:
- Updating readme to point to official certification docs for clarity
- Updating the testing section so parters know these are not tests they run, but tests for developers.
- [JIRA](https://issues.redhat.com/browse/EET-3216)

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,11 @@
 This project is in active and rapid development! Many facets of this project are
 subject to change, and some features are not fully implemented.
 
+## Certification Workflow Guide
+
+For the complete container and operator bundle certification workflow instructions, please reference our 
+[official certification documentation](https://access.redhat.com/documentation/en-us/red_hat_software_certification/8.56/html/red_hat_software_certification_workflow_guide/index).
+
 ## Requirements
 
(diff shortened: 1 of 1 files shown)


acornett21:
@taylorjordanNC Please take a look at let us know if you have more suggestions. Thanks.

taylorjordanNC:
lgtm!

taylorjordanNC (approved):

README.md
```diff
@@ -13,6 +13,9 @@
 This project is in active and rapid development! Many facets of this project are
 subject to change, and some features are not fully implemented.
```
skattoju, line 15: May be add a section header ex: `## Certification Documentation` or something like that ?

skattoju (changes requested):
A section header might help it stand out ex: ## Certification Workflow Docs or something like that ?

README.md
acornett21: I added a header with the same name that is used on the website.

acornett21:
> This looks like it does more than just update for the official docs? Do we want to split this into two PRs?
@bcrochet Both of these asks were in the JIRA (which I added above). I've updated the title of this PR for clarity, since it only pulled the first half of the commit message.

bcrochet (approved):
/lgtm

PR #902, opened 2023-02-27
updating openshift api and go-client dependencies

acornett21:
- Dependabot was pulling old dependencies, since OpenShift does not release/tag versions of these two dependencies.


bcrochet (approved):
/lgtm

komish (approved):
/lgtm

PR #908, opened 2023-03-01
Switch go-rpmdb to use the upstream

bcrochet:
We would like to no longer have to carry any downstream commits, and have out own fork to maintain. Recent changes to upstream have allowed us to possibly drop our fork and just use the upstream repo.

Need to verify that this does NOT break s390x.


bcrochet:
s390x has been tested and is currently known to work

internal/rpm/rpm.go
```diff
@@ -8,6 +8,8 @@ import (
 	"path/filepath"
 
 	rpmdb "github.com/knqyf263/go-rpmdb/pkg"
+	// This pulls in the sqlite dependency
+	_ "github.com/glebarez/go-sqlite"
```
komish, line 12: Isn't it kind of weird that _we_ have to initialize go-sqlite, vs. the rpmdb lib doing so? Or is this one of those things where "we choose"?

komish:
Just waiting for rebase to LGTM.

internal/rpm/rpm.go
acornett21: Pretty sure this is a `we choose` type deal since stdlib only gives the interface for DB's and then the caller chooses the implementation.

acornett21 (approved):

Issue #781, opened 2022-09-15
Add Support to Preflight for Packaging of Container Results


jomkz:
## Is your feature request related to a problem? Please describe.

We need to add support for packaging up the results that get generated as part of the `check container` command. This will allow partners running in disconnected environments the ability to submit the results to Red Hat through some other means than the preflight submit flow, which requires an internet connection.

## Describe the solution you'd like.

We discussed adding a new `--offline` flag that will cause preflight to package the results in a way that can be easily sent outside of the normal submit flow.

## Describe alternatives you've considered.

N/A

## Additional context.

N/A

PR #912, opened 2023-03-02
add offline flag and necessary code to tar artifacts directory

acornett21:
This is the work for `--offline` flag to tar up the artifacts directory.

- Fixes: #781 
- Add new `--offline` flag for `check container` command
   - make this flag mutually exclusive with `--submit` flag's presence
- Add a method to `tar` the artifacts directory.

### DIFF
--- a/cmd/preflight/cmd/check_container.go
+++ b/cmd/preflight/cmd/check_container.go
@@ -1,8 +1,13 @@
 package cmd
 
 import (
+	"archive/tar"
+	"bytes"
 	"context"
 	"fmt"
+	"io"
+	"os"
+	"path/filepath"
 	rt "runtime"
 	"strings"
 
@@ -49,6 +54,14 @@ func checkContainerCmd(runpreflight runPreflight) *cobra.Command {
 
 	flags.Bool("insecure", false, "Use insecure protocol for the registry. Default is False. Cannot be used with submit.")
 	_ = viper.BindPFlag("insecure", flags.Lookup("insecure"))
+
+	flags.Bool("offline", false, "Intended to be used in disconnected environments and will tar artifacts used for submission,\n"+
+		"enabling other Red Hat managed tools/process to submit these artifacts at a later time.\n"+
+		"Cannot be used with submit.")
+	_ = viper.BindPFlag("offline", flags.Lookup("offline"))
+
+	// Make --submit mutually exclusive to --offline
+	checkContainerCmd.MarkFlagsMutuallyExclusive("submit", "offline")
 
 	// Make --submit mutually exclusive to --insecure
 	checkContainerCmd.MarkFlagsMutuallyExclusive("submit", "insecure")
@@ -116,7 +129,7 @@ func checkContainerRunE(cmd *cobra.Command, args []string, runpreflight runPrefl
 	// Run the  container check.
 	cmd.SilenceUsage = true
 
-	return runpreflight(
+	err = runpreflight(
 		ctx,
 		checkcontainer.Run,
 		cli.CheckConfig{
@@ -127,6 +140,46 @@ func checkContainerRunE(cmd *cobra.Command, args []string, runpreflight runPrefl
 		&runtime.ResultWriterFile{},
 		resultSubmitter,
 	)
+
+	if err != nil {
+		return err
+	}
+
+	// checking for offline flag, if present tar up the contents of the artifacts directory
+	if cfg.Offline {
+		src := artifactsWriter.Path()
+		var buf bytes.Buffer
+
+		// check to see if a tar file already exist to account for someone re-running
+		exists, err := artifactsWriter.Exists(check.DefaultArtifactsTarFileName)
+		if err != nil {
+			return fmt.Errorf("unable to check if tar already exists: %v", err)
+		}
+
+		// remove the tar file if it exists
+		if exists {
+			err = artifactsWriter.Remove(check.DefaultArtifactsTarFileName)
+			if err != nil {
+				return fmt.Errorf("unable to remove existing tar: %v", err)
+			}
+		}
+
+		// tar the directory
+		err = artifactsTar(ctx, src, &buf)
+		if err != nil {
+			return fmt.Errorf("unable to tar up artifacts directory: %v", err)
+		}
+
+		// writing the tar file to disk
+		_, err = artifactsWriter.WriteFile(check.DefaultArtifactsTarFileName, &buf)
+		if err != nil {
+			return fmt.Errorf("could not artifacts tar to artifacts dir: %w", err)
+		}
+
+		logger.Info("artifact tar written to disk", "filename", check.DefaultArtifactsTarFileName)
+	}
+
+	return nil
 }
 
 func checkContainerPositionalArgs(cmd *cobra.Command, args []string) error {
@@ -213,4 +266,76 @@ func generateContainerCheckOptions(cfg *runtime.Config) []container.Option {
 	}
 
 	return o
+}
+
+// artifactsTar takes a source path and a writer; a tar writer loops over the files in the source
+// directory, writes the appropriate header information and copies the file into the tar writer
+//
+//nolint:unparam // ctx is unused. Keep for future use.
+func artifactsTar(ctx context.Context, src string, w io.Writer) error {
+	// ensure the src actually exists before trying to tar it
+	if _, err := os.Stat(src); err != nil {
+		return fmt.Errorf("unable to tar files - %v", err.Error())
+	}
+
+	tw := tar.NewWriter(w)
+	defer tw.Close()
+
+	// getting a list of DirEntry's
+	files, err := os.ReadDir(src)
+	if err != nil {
+		return err
+	}
+
+	// iterating over the list
+	for _, file := range files {
+		// ignoring directories and only processing files
+		if file.IsDir() {
+			continue
+		}
+
+		// getting the FileInfo from the DirEntry, account for errors with this call ie ErrNotExist
+		fileInfo, err := file.Info()
+		if err != nil {
+			return err
+		}
+
+		// continue on non-regular files
+		if !fileInfo.Mode().IsRegular() {
+			continue
+		}
+
+		// create a new dir/file header
+		header, err := tar.FileInfoHeader(fileInfo, fileInfo.Name())
+		if err != nil {
+			return err
+		}
+
+		// write the header
+		if err := tw.WriteHeader(header); err != nil {
+			return err
+		}
+
+		err = func() error {
+			// open files to tar
+			f, err := os.Open(filepath.Join(src, fileInfo.Name()))
+			if err != nil {
+				return err
+			}
+			defer f.Close()
+
+			// copy file data into tar writer
+			if _, err := io.Copy(tw, f); err != nil {
+				return err
+			}
+
+			return nil
+		}()
+
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
 }
--- a/cmd/preflight/cmd/check_container_test.go
+++ b/cmd/preflight/cmd/check_container_test.go
@@ -7,7 +7,9 @@ import (
 	"os"
 	"path/filepath"
 
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/artifacts"
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification"
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/internal/check"
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/internal/cli"
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/internal/formatters"
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/internal/lib"
@@ -24,7 +26,7 @@ var _ = Describe("Check Container Command", func() {
 	Context("when running the check container subcommand", func() {
 		Context("With all of the required parameters", func() {
 			It("should reach the core logic, but throw an error because of the placeholder values for the container image", func() {
-				_, err := executeCommand(checkContainerCmd(mockRunPreflight), "example.com/example/image:mytag")
+				_, err := executeCommand(checkContainerCmd(mockRunPreflightReturnNil), "example.com/example/image:mytag")
 				Expect(err).To(HaveOccurred())
 			})
 		})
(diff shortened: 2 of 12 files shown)


internal/pyxis/submit_test.go
```diff
@@ -101,6 +107,40 @@ var _ = Describe("Pyxis Submit", func() {
...
+			})
+			It("should write a tar to disk", func() {
+				certResults, err := pyxisClient.SubmitResults(artifactsContext, &certInput)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(certResults).ToNot(BeNil())
+				Expect(certResults.CertProject).ToNot(BeNil())
+				Expect(certResults.CertImage).ToNot(BeNil())
+				Expect(certResults.TestResults).ToNot(BeNil())
+			})
+		})
```
bcrochet, line 143: Negative testing with a bad writer perhaps?

bcrochet (changes requested):
First round of comments.

internal/pyxis/submit.go
```diff
@@ -134,3 +172,68 @@ func normalizeDockerRegistry(registry string) string {
 
 	return registry
 }
+
+// artifactsTar takes a source path and a writer; a tar writer loops over the files in the source
+// directory, writes the appropriate header information and copies the file into the tar writer
+//
+//nolint:unparam // ctx is unused. Keep for future use.
+func artifactsTar(ctx context.Context, src string, w io.Writer) error {
```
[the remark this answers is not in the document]
acornett21: I think what you're asking is to use `afero.Fs` here, but I think that would make it more confusing of an abstraction. All of the places in the `engine` that write files use `path` and `src` or `dst` as a string, this follows that pattern.

If we didn't want to follow that pattern, maybe since the caller of this is a `FileSystemWriter` which containrs an afero.FS (though that wouldn't be used in the implementation, just the `dir string`) this method could take the `FileSystemWriter`?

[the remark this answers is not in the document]
bcrochet: That's fair. No need to hold this up. We can revisit another time.

internal/pyxis/submit.go
```diff
@@ -1,11 +1,22 @@
...
+	"os"
+	"path/filepath"
 
+	prfltartifacts "github.com/redhat-openshift-ecosystem/openshift-preflight/artifacts"
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/internal/check"
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/internal/log"
+
+	"github.com/go-logr/logr"
 	"github.com/google/go-containerregistry/pkg/name"
+	"github.com/spf13/viper"
```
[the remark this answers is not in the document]
acornett21: I went with _one_ approach to update the `ResolveSubmitter` to get the `offline` flag into the submit flow. Another approach would be to add the flag to the `cli` then the submit functions would have to change their params to take this `bool` as well.

I dislike both approaches, but I think what I have here might be more acceptable then changing a bunch of method signatures.

internal/pyxis/submit_test.go
```diff
@@ -101,6 +107,40 @@ var _ = Describe("Pyxis Submit", func() {
...
+				Expect(err).ToNot(HaveOccurred())
+				defer f2.Close()
+
+				aw, err := artifacts.NewFilesystemWriter(artifacts.WithDirectory(tmpDir))
+				Expect(err).ToNot(HaveOccurred())
+
+				artifactsContext = artifacts.ContextWithWriter(context.Background(), aw)
+			})
+			JustAfterEach(func() {
+				viper.Set("offline", false)
```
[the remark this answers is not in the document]
acornett21: Closing since this isn't applicable anymore.

internal/pyxis/submit_test.go
acornett21: I've added one negative test to not add an `ArtifactsWriter` to the context, which is the only error that can be tested in the main flow. 

I'll add `artifactsTar` method level testing in it's own `Context()` block, I don't think it makes sense at this location and should live on it's own.

acornett21: More tests have been added. I was not able to cover the `Irregular` file or `WriteHeader` errors, since whatever I passed in, either failed in the test setup, or failed before either of these method calls. 

If you feel covering those are important, let me know your ideas to cover them with proper 'fake' files and 'permissions' that fall through the other code, but fails in either case.

internal/pyxis/submit.go
```diff
@@ -134,3 +172,68 @@ func normalizeDockerRegistry(registry string) string {
...
+			}
+
+			// copy file data into tar writer
+			if _, err := io.Copy(tw, f); err != nil {
+				return err
+			}
+
+			// manually close here after each file operation; deferring would cause each file close
+			// to wait until all operations have completed.
+			f.Close()
```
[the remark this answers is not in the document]
acornett21: Let me know if this closer is correct, my concern is that `err` is already defined, so scoping of `err` may be an issue, but I'm not sure.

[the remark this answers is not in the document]
acornett21: closing this to continue conversation above.

bcrochet (approved):
/lgtm

cmd/preflight/cmd/check_container.go
```diff
@@ -42,6 +42,15 @@ func checkContainerCmd() *cobra.Command {
...
 	_ = viper.BindPFlag("insecure", flags.Lookup("insecure"))
 
+	flags.Bool("offline", false, "Intended to be used in disconnected environments and will tar artifacts used for submission,\n"+
+		"enabling other Red Hat managed tools/process to submit these artifacts at a later time.\n"+
+		"Note: this flag is only available with the submit flag")
+
+	// TODO: remove when feature is fully implemented
+	_ = flags.MarkHidden("offline")
+
+	_ = viper.BindPFlag("offline", flags.Lookup("offline"))
```
komish, line 52: NIT: remove the spaces here.

internal/pyxis/submit.go
```diff
@@ -1,10 +1,20 @@
...
+	"archive/tar"
+	"bytes"
 	"context"
 	"errors"
 	"fmt"
+	"io"
+	"os"
+	"path/filepath"
 
+	prfltartifacts "github.com/redhat-openshift-ecosystem/openshift-preflight/artifacts"
```
komish, line 13: NIT:
```suggestion
	pfltartifacts "github.com/redhat-openshift-ecosystem/openshift-preflight/artifacts"
```

komish (commented):
Few nits. I'll test the code shortly

skattoju4 (approved):

skattoju (approved):

cmd/preflight/cmd/check_container.go
```diff
@@ -49,6 +55,17 @@ func checkContainerCmd(runpreflight runPreflight) *cobra.Command {
 	flags.Bool("insecure", false, "Use insecure protocol for the registry. Default is False. Cannot be used with submit.")
 	_ = viper.BindPFlag("insecure", flags.Lookup("insecure"))
 
+	flags.Bool("offline", false, "Intended to be used in disconnected environments and will tar artifacts used for submission,\n"+
+		"enabling other Red Hat managed tools/process to submit these artifacts at a later time.\n"+
+		"Note: this flag is only available with the submit flag")
```
komish, line 60:
```suggestion
		"Cannot be used with submit.")
```

cmd/preflight/cmd/check_container.go
```diff
@@ -49,6 +55,17 @@ func checkContainerCmd(runpreflight runPreflight) *cobra.Command {
...
+
+	// TODO: remove MarkHidden when feature is fully implemented
+	_ = flags.MarkHidden("offline")
```
komish, line 63: What remains for this feature to be fully implemented?

cmd/preflight/cmd/check_container.go
```diff
@@ -126,6 +143,32 @@ func checkContainerRunE(cmd *cobra.Command, args []string, runpreflight runPrefl
...
+			return fmt.Errorf("unable to tar up artifacts directory: %v", err)
+		}
+
+		// writing the tar file to disk
+		_, err = artifactsWriter.WriteFile(check.DefaultArtifactsTarFileName, &buf)
+		if err != nil {
+			return fmt.Errorf("could not artifacts tar to artifacts dir: %w", err)
+		}
+
+		logger.V(log.DBG).Info("artifact tar written to disk", "filename", check.DefaultArtifactsTarFileName)
```
komish, line 168: We should make this log line a lower verbosity. This is an explicit flag provided by the user, so the outcome should be visible without needing extra verbosity.

cmd/preflight/cmd/check_container.go
```diff
@@ -211,3 +254,75 @@ func generateContainerCheckOptions(cfg *runtime.Config) []container.Option {
 
 	return o
 }
+
+// artifactsTar takes a source path and a writer; a tar writer loops over the files in the source
+// directory, writes the appropriate header information and copies the file into the tar writer
+//
+//nolint:unparam // ctx is unused. Keep for future use.
+func artifactsTar(ctx context.Context, src string, w io.Writer) error {
```
komish, line 262: I haven't tested this here, but I think you can remove the nolint directive by anonymous assigning `ctx`. E.g.

```suggestion
func artifactsTar(_ context.Context, src string, w io.Writer) error {
```

cmd/preflight/cmd/check_container_test.go
```diff
@@ -23,7 +24,7 @@ var _ = Describe("Check Container Command", func() {
 	Context("when running the check container subcommand", func() {
 		Context("With all of the required parameters", func() {
 			It("should reach the core logic, but throw an error because of the placeholder values for the container image", func() {
-				_, err := executeCommand(checkContainerCmd(mockRunPreflight), "example.com/example/image:mytag")
+				_, err := executeCommand(checkContainerCmd(mockRunPreflightReturnNil), "example.com/example/image:mytag")
```
komish, line 27: These test suite adjustments, they're from a different PR, right?

komish (changes requested):
A few changes and questions. Seems to function. Not sure I love the artifacts.tar being written to the artifacts directory, but alas.

cmd/preflight/cmd/check_container_test.go
acornett21: The changes are included from the original PR, I refactored them again to provide a positive and negative test. The original test never needed a negative test since it did not have an 
```
if err ! = nil
```
block of code. I hope this makes sense.

cmd/preflight/cmd/check_container.go
acornett21: That does pass a lint check, but I think I'd rather keep the `nolint` since in an IDE the `_` makes the code look like it's used, ie highlighted vs grey'd out. But if other's prefer `_` I'll switch to that.

cmd/preflight/cmd/check_container.go
acornett21: Good catch.

acornett21:
> A few changes and questions. Seems to function. Not sure I love the artifacts.tar being written to the artifacts directory, but alas.

I'd be open to other ideas, should we force that the user giving the `offline` flag _also_ has to give us the `artifacts` flag and provide a custom directory? I think that might be an option.

cmd/preflight/cmd/check_container.go
komish: I do prefer the `_` here, I think it's the appropriate application. Happy to have someone else weigh in.

komish:
> > A few changes and questions. Seems to function. Not sure I love the artifacts.tar being written to the artifacts directory, but alas.
> 
> I'd be open to other ideas, should we force that the user giving the `offline` flag _also_ has to give us the `artifacts` flag and provide a custom directory? I think that might be an option.

I don't think this buys us anything. We'd still write to /custompath/artifacts.tar, alongside the artifacts that would otherwise be included.

cmd/preflight/cmd/check_container.go
bcrochet: I'm on the fence here. If you go with the _, I'd still leave the //nolint:unparam (or just the comment) so it's clear to future someone why that's even there.

komish: With this in mind, up to you @acornett21 how you'd like to proceed.

acornett21:
@komish I added logic to remove a tar if it already exists, I think this makes the most sense to not run into the situation you described where someone wants to keep re-running `check container`

cmd/preflight/cmd/check_container.go
acornett21: For consistency with other functions that do not use context, I'll keep this the as is in this PR.

komish (changes requested):
Mostly LGTM. I think this needs to include preflight.log too, though.

internal/pyxis/submit.go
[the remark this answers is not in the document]
acornett21: This is no longer valid since these code moved out of submit.

internal/pyxis/submit_test.go
acornett21: More tests have been added since this code moved out of the submit flow.

acornett21:
> I think this needs to include preflight.log too, though.

@komish I've added logic to write the log to the same directory where the artifacts are written to.

bcrochet (approved):
/lgtm

komish:
This isn't quite working how I expect. I'm seeing a log format that doesn't match our standard formatter on first runs with the offline flag.

```
 ⤿ ./preflight check container --offline quay.io/opdev/simple-demo-operator:latest
INFO[0000] certification library version                 version="0.0.0 <commit: cd9998494513973fb4fd6e4e657f38f9890dc2ee>"
INFO[0005] check completed                               check=HasLicense result=PASSED
INFO[0005] check completed                               check=HasUniqueTag result=PASSED
INFO[0005] check completed                               check=LayerCountAcceptable result=PASSED
INFO[0005] check completed                               check=HasNoProhibitedPackages result=PASSED
INFO[0005] check completed                               check=HasRequiredLabel result=PASSED
INFO[0005] USER 65532:65532 specified that is non-root  
INFO[0005] check completed                               check=RunAsNonRoot result=PASSED
INFO[0006] check completed                               check=HasModifiedFiles result=PASSED
INFO[0007] check completed                               check=BasedOnUbi result=PASSED

... truncated ...
```

I get this log format only on initial runs, and I get our usual log format on reruns (e.g. when I already have an artifacts directory already in the working directory). 

It almost looks like we _aren't_ initializing our logger when we run with the offline flag on the first attempt, but then are if we're overwriting the preflight log.

---

I also don't _love_ that we aren't writing preflight.log where it was before. This will probably break folks that have automation oriented around preflight.log being in the current working directory.

---

Finally, if we set the logfile using `PFLT_LOGFILE` on a rerun, we get both files in the archive.

```
 ⤿ PFLT_LOGFILE=foo.log ./preflight check container --offline quay.io/opdev/simple-demo-operator:latest
... truncated ...


 ⤿ tar xvf artifacts/artifacts.tar 
cert-image.json
foo.log
preflight.log
results.json
rpm-manifest.json
```

I think that is probably passable for the moment. The easy workaround is "clean up your directory".

komish (changes requested):
See previous comment for requested changes

acornett21:
@komish I've addressed most of the issues from your comment [here](https://github.com/redhat-openshift-ecosystem/openshift-preflight/pull/912#issuecomment-1487473339). 

The only thing that really can't be addressed is the setting of `PFLT_LOGFILE` in one run and the unsetting in another run. The issue that you point out would exist even in _without_ the `--offline` flag.

Note: PR #939 has a change that is also in this PR, so I need that to merge first so I can rebase.

komish (approved):
Alright. I know you know this already, but I can't avoid restating that I think this workflow (creating a tarball of the artifacts directory and the logfile) could/should have lived outside of preflight. I know there are pending actions for the folks that requested this.. so we'll see where this goes.

With that

/lgtm

PR #918, opened 2023-03-10
adding DeferCleanup to check container tests to reset test data

acornett21:
- Adding DeferCleanup to tests to reset the certification project id. This ensures that new tests can be added and do not fail for the same reason that these tests were written to cover.


bcrochet (approved):
/lgtm

skattoju (approved):

PR #919, opened 2023-03-13
refactoring check container rune to use a closure

acornett21:
## Motivation
None of the `check_container` code had test coverage, and existing tests only cover the error scenario where the logger exits early. This isn't ideal, but not terrible with how we currently return `cli.RunPreflight` as the last step in this execution flow and the CLI is heavily tested. However, in the future there will be additional code after `cli.RunPreflight` so being able to test various pass/fail scenarios will be helpful.

## Additions
- executeCommandWithLogger func gives the ability to pass in a logger that is then added to a context, this enables getting to the core functionality.
- runPreflight func gives the ability to mock `cli.RunPreflight`

### DIFF
--- a/cmd/preflight/cmd/check.go
+++ b/cmd/preflight/cmd/check.go
@@ -1,6 +1,8 @@
 package cmd
 
 import (
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/internal/cli"
+
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 )
@@ -21,7 +23,7 @@ func checkCmd() *cobra.Command {
 	_ = viper.BindPFlag("artifacts", checkCmd.PersistentFlags().Lookup("artifacts"))
 
 	checkCmd.AddCommand(checkOperatorCmd())
-	checkCmd.AddCommand(checkContainerCmd())
+	checkCmd.AddCommand(checkContainerCmd(cli.RunPreflight))
 
 	return checkCmd
 }
--- a/cmd/preflight/cmd/check_container.go
+++ b/cmd/preflight/cmd/check_container.go
@@ -1,11 +1,13 @@
 package cmd
 
 import (
+	"context"
 	"fmt"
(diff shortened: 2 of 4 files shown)


bcrochet (approved):
/lgtm

cmd/preflight/cmd/check_container.go
```diff
@@ -22,7 +24,10 @@ import (
 
 var submit bool
 
-func checkContainerCmd() *cobra.Command {
+// runPreflight is introduced to make testing of this command possible, it has the same method signature as cli.RunPreflight.
```
skattoju, line 27: I see you have defined `mockRunPreflight` in the test code to have the same signature as `cli.RunPreflight` which makes sense. Do you need another `type runPreflight` here ? Can't you just do `checkContainerCmd(runprelight cli.RunPreflight)` (I might be confused 😅)

cmd/preflight/cmd/check.go
```diff
@@ -21,7 +23,7 @@ func checkCmd() *cobra.Command {
 	_ = viper.BindPFlag("artifacts", checkCmd.PersistentFlags().Lookup("artifacts"))
 
 	checkCmd.AddCommand(checkOperatorCmd())
```
skattoju, line 25: Is a similar abstraction needed for checkOperator ?

bcrochet: Yes. Separate PR coming.

skattoju (approved):

cmd/preflight/cmd/check_container.go
skattoju: The type definition could be placed in the cli package though right ?

acornett21: @skattoju That's what I was trying to do in the first place, but then you end up with the problem of not being about to 'inject' it into the CMD. Also, since the CLI package is internal, it shouldn't care about how/why a `caller` needs to mock it's function call to be testable.

PR #922, opened 2023-03-16
incorporating runPreflight func call into check_operator cmd

acornett21:
## Motivation
The `cmd` at this level had basically zero code coverage since we were always exiting early for not having a `logr` in the context. This PR builds upon the same format that was added for `check_container`. These tests should now cover all of the core logic that I think is possible. But if there is another `err` case that we can mock I'm happy to add it here, just let me know.

**Note: This PR has mockRunPreflightReturnErr method that is in another PR as well, whichever PR is merged first I'll rebase accordingly**

### DIFF
--- a/cmd/preflight/cmd/check_operator.go
+++ b/cmd/preflight/cmd/check_operator.go
@@ -19,7 +19,7 @@ import (
 	"github.com/spf13/viper"
 )
 
-func checkOperatorCmd() *cobra.Command {
+func checkOperatorCmd(runpreflight runPreflight) *cobra.Command {
 	checkOperatorCmd := &cobra.Command{
 		Use:   "operator",
 		Short: "Run checks for an Operator",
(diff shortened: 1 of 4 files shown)


acornett21:

cmd/preflight/cmd/check_operator.go
```diff
@@ -71,7 +73,7 @@ func ensureIndexImageConfigIsSet() error {
 }
 
 // checkOperatorRunE executes checkOperator using the user args to inform the execution.
-func checkOperatorRunE(cmd *cobra.Command, args []string) error {
```
komish, line 74: I think I'd like to see this, and potentially the equivalent for container checks, change function signatures.

By convention, checkOperatorRunE's signature matches the RunE signature definition. We're changing this to accept dependencies in the function signature, which to me, strays from the original intention behind the convention's use.

I'm thinking these should be `newCheckOperatorRunE` and return a RunE signature.

komish (commented):
Small nit. otherwise, LGTM.

cmd/preflight/cmd/check_operator.go
acornett21: The order got messed up in my refactor I can fix that, do you still feel they still need a new name?

komish: It's a small detail; I'd say it can't hurt. Your call.

skattoju (approved):

cmd/preflight/cmd/check_operator.go
bcrochet: Why return a RunE signature? I get renaming them (not a RunE technically anymore) but I don't understand the return part.

komish: It would let you define `RunE: NewRunEFunctionWithThese(deps deps)`. Either that, or you use the anon function. They accomplish the same thing.

acornett21: @komish You can't do what you are suggesting, if adding more args it has to be in a closure, meaning you can't call with params.

bcrochet: @acornett21 I see what he's getting at. If NewRunEFunctionWithThese() has a return type of `func (cmd *cobra.Command, args []string) error` then you can call it as he states. 

However, that would require re-working the function we have to account for that. IMHO how we have it is cleaner. As @komish said, it accomplishes the same thing.

komish (approved):
/lgtm

PR #929, opened 2023-03-22
Ensure the engine logs error strings

komish:
The engine is logging errors slightly differently than other places that might call our log helper, using WithValues to inject context instead of using logger.Error(...). 

WithValues doesn't seem to stringify error structs, so currently we seem to be getting empty structures instead of strings. This should resolve that.


acornett21 (approved):

bcrochet (approved):
/lgtm

PR #930, opened 2023-03-23
Fix HasModifiedFiles to check layer by layer

### LINKED ISSUE #846: Preflight does not detect the modification of the package and throw the warning or error in the results

bcrochet:
* (almost) Complete rewrite of the check
* Ignore any path that starts with "etc/" or "var/"
* Ignore any RPM file that has config, doc, or config(missingok) flags
* Check each layer against the last, rather than all the flattened layers at once

Fixes #846


internal/engine/engine.go
```diff
@@ -224,7 +224,7 @@ func (c *CraneEngine) ExecuteChecks(ctx context.Context) error {
 		checkElapsedTime := time.Since(checkStartTime)
 
 		if err != nil {
-			logger.WithValues("result", "ERROR", "err", err).Info("check completed", "check", check.Name())
+			logger.WithValues("result", "ERROR", "err", err.Error()).Info("check completed", "check", check.Name())
```
acornett21, line 227: Isn't this a change from `main`?

komish: Confirmed - Did something go sideways in a rebase?

komish (approved):
/lgtm

Still seeing only expected failures across my test 30 images or so.

internal/engine/engine.go
bcrochet: Actually, no. I think I had added that for something else. Won't hurt to leave it here for the moment. It will just get removed when we cherry-pick this back.

acornett21 (approved):

PR #939, opened 2023-03-28
setting up log formatter early to force a proper format

acornett21:
@komish pointed out in the `--offline` pr that the first run of preflight logs differently (see below). The root case to this is that if we can't do an `os.Open` and fall bace to logging to stderr the logrus textformatter just does what it wants.

This pr moves up the logic and forces us to disable color, which then gives us a consistent format to a console as we would to a file with timestamp format.

Example of the error
```
 ./preflight check container --offline quay.io/opdev/simple-demo-operator:latest
INFO[0000] certification library version                 version="0.0.0 <commit: cd9998494513973fb4fd6e4e657f38f9890dc2ee>"
INFO[0005] check completed                               check=HasLicense result=PASSED
INFO[0005] check completed                               check=HasUniqueTag result=PASSED
INFO[0005] check completed                               check=LayerCountAcceptable result=PASSED
INFO[0005] check completed                               check=HasNoProhibitedPackages result=PASSED
INFO[0005] check completed                               check=HasRequiredLabel result=PASSED
INFO[0005] USER 65532:65532 specified that is non-root  
INFO[0005] check completed                               check=RunAsNonRoot result=PASSED
INFO[0006] check completed                               check=HasModifiedFiles result=PASSED
INFO[0007] check completed                               check=BasedOnUbi result=PASSED

... truncated ...
```

### DIFF
--- a/cmd/preflight/cmd/root.go
+++ b/cmd/preflight/cmd/root.go
@@ -88,6 +88,8 @@ func initConfig() {
 func preRunConfig(cmd *cobra.Command, args []string) {
 	viper := viper.Instance()
 	l := logrus.New()
+	l.SetFormatter(&logrus.TextFormatter{DisableColors: true})
+
 	// set up logging
 	logname := viper.GetString("logfile")
 	logFile, err := os.OpenFile(logname, os.O_CREATE|os.O_WRONLY|os.O_TRUNC, 0o600)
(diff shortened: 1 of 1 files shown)


komish (approved):
/lgtm

FWIW, what's happening is that Logrus is internally changing the format if the output file is a terminal (e.g. stdout, stderr). If it isn't, it internally sets the output to look how preflight has seen in the past. If it is, it shows the `INFO[0000]` that we've seen. Disabling color is the key difference that seems to give us the format we want so doing this makes it consistent, regardless of the output file.

bcrochet (approved):
/lgtm

Issue #938, opened 2023-03-28
Emit the target image into the log at the INFO log level


komish:
This line only exists in our debug logs. It would be a good point of reference for debugging in cases where we don't have the results.json file itself. 

https://github.com/redhat-openshift-ecosystem/openshift-preflight/blob/main/internal/engine/engine.go#L80

Make this visible at INFO log level so that all logs will include it.

PR #940, opened 2023-03-29
remove verbosity from target image log line 

acornett21:
- Fixes: #938 
- Image registry/repo should always be logged, so logs that are uploaded to pyxis have all the necessary information for cert ops to better troubleshoot.


komish (approved):
/lgtm

PR #942, opened 2023-04-03
mac architectures build

### LINKED ISSUE #850: Preflight needs an Apple Silicon binary
As a user, I would like to download a binary made for MacOS on M1/M2 Silicon without having to build it myself.

itroyano:
Darwin OS builds for Preflight:  Intel Mac,  Apple Silicon Mac.

Pending - adjust the GH Actions.

Fixes #850

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -41,5 +41,8 @@ jobs:
         github-token: ${{ secrets.GITHUB_TOKEN }}
         file: coverage.out
 
-    - name: Build Multi-arch
-      run: make build-multi-arch
+    - name: Build Multi-arch-linux
+      run: make build-multi-arch-linux
+
+    - name: Build Multi-arch-mac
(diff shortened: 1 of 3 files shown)


redhatrises:
@acornett21 is this mergable or can it be merged soon? Would be handy to have.

acornett21:
@redhatrises I don't think this is ready for merge, and I don't think there was a consensus for the approach. Due to most of the team working on other higher priority issues, I'm not sure when this can be revisited.

itroyano:
I'm honestly surprised there are no conflicts.

acornett21:
If we're sure that this works, (which last I remember talking about this we were unsure), then I'm fine merging this.

jomkz:
Agree with @bcrochet here, let's decide if we want to merge this and unofficially support mac binaries going forward or close it out as not happening.

PR #948, opened 2023-04-04
Use mutate.Extract for all images

### LINKED ISSUE #928: preflight cannot process tests with whiteout files present in the bundle directory
## Bug Description
preflight is not able to show the errors or run tests on bundles that have whiteout files in the bundle directory.
Eg: registry.redhat.io/openshift-update-service/cincinnati-operator-bundle@sha256:2743f74cc102c059d7c1dc6698eb947cca75a3cc4230905f09703666ae7a4dbf

preflight version 0.0.0 <commit: 8973da0f12304b7b6f18b9d6687639be3d0ad9b9>

## Steps to Reproduce:
(How can we reproduce this?)
1) run preflight against registry.redhat.io/openshift-update-service/cincinnati-operator-bundle@sha256:2743f74cc102c059d7c1dc6698eb947cca75a3cc4230905f09703666ae7a4dbf
``` 
preflight check operator registry.redhat.io/openshift-update-service/cincinnati-operator-bundle@sha256:2743f74cc102c059d7c1dc6698eb947cca75a3cc4230905f09703666ae7a4dbf
```

## Expected Result
Show errors in debug log.

## Actual Result
After running preflight in debug mode:
```
result log: 2023/03/22 19:29:38 error loading objs in directory: bundle manifests dir has hidden file: /bundle/manifests/.wh..wh..opq
```

cc: @acornett21

bcrochet:
Crane's crane.Export function was special-casing when there is a single layer. It doesn't call mutate.Extract if there is a single layer. Problem is, if there are whiteout files in that first layer, they don't get handled.

So instead, we will just have our own 'export' function that just calls mutate.Extract no matter what.

Fixes #928

### DIFF
--- a/internal/engine/engine.go
+++ b/internal/engine/engine.go
@@ -41,6 +41,7 @@ import (
 	"github.com/google/go-containerregistry/pkg/name"
 	cranev1 "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/google/go-containerregistry/pkg/v1/cache"
+	"github.com/google/go-containerregistry/pkg/v1/mutate"
 	"github.com/google/go-containerregistry/pkg/v1/remote"
 )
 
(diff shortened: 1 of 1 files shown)


acornett21 (approved):

komish (approved):
/lgtm

Nice catch

PR #959, opened 2023-04-14
Fix tests writing to filesystem

### LINKED ISSUE #957: Tests writing to filesystem and not cleaning up
Not sure which tests are doing it, but something is writing artifacts and logs to random directories throughout the codebase. This is undesirable.

bcrochet:
The root_test.cmd was not the direct cause. But the offline tests were what was manifesting the problem. Basically, when the offline flag was used, a second file was opened in the artifacts directory, but no provision was made for cleaning up the first file that was opened (usually in the cwd). Because of this, the tests wouldn't clean up that first file. This moves the logic around to only open a single log file, either in the default location, a location specified by the user, or the artifacts directory in offline mode.

Fixes #957


komish (approved):
/lgtm

cmd/preflight/cmd/root.go
```diff
@@ -104,19 +114,7 @@ func preRunConfig(cmd *cobra.Command, args []string) {
...
-		artifacts := viper.GetString("artifacts")
-
-		// ignoring error since OpenFile will error and we'll still have the multiwriter from above
-		_ = os.Mkdir(artifacts, 0o777)
-
-		artifactsLogFile, err := os.OpenFile(filepath.Join(artifacts, logname), os.O_CREATE|os.O_WRONLY|os.O_TRUNC, 0o600)
-		if err == nil {
-			mw := io.MultiWriter(os.Stderr, logFile, artifactsLogFile)
-			l.SetOutput(mw)
-		}
```
acornett21, line 118: The intent of this is to have the files in the `tar` and at the default location. This is so the user can actually see the logs and so we get a copy of them.

acornett21 (approved):

komish (approved):
/lgtm

PR #960, opened 2023-04-17
removing dependency on ocp-olm-catalog-validator

acornett21:
## Motivation
With trying to debug an issue with an error in a PR for the hosted pipeline found [here](https://github.com/redhat-openshift-ecosystem/certified-operators/pull/2109). It was determined that the `ocp validator` was calling `AlphaDeprecatedAPIsValidator` from `operator-sdk`, and making decisions based on the outcome of that call. The problem is that it seems this library was only targeting deprecations in `4.9`, and was leading to unnecessary failures for later versions of OCP. Also, this meant that prelfight was calling `AlphaDeprecatedAPIsValidator` twice.

## Conculsion
After meeting with the business, we can remove this extra library call, since `AlphaDeprecatedAPIsValidator` gives us 4.9, 4.12, and 4.13 k8's api deprecations out of the box, and we need not update/maintain another library to do so.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -17,7 +17,6 @@ require (
 	github.com/operator-framework/api v0.17.4-0.20230223191600-0131a6301e42
 	github.com/operator-framework/operator-manifest-tools v0.2.3-0.20230227155221-caa8b9e1ab12
 	github.com/redhat-certification/chart-verifier v0.0.0-20230309162114-4bccec13d65a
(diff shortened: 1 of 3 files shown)


bcrochet (approved):
/lgtm

komish (approved):
/lgtm

komish (approved):
/lgtm

PR #967, opened 2023-04-27
Process links in layers to ignore the target

bcrochet:
When a lyaer contains the creation of the symbolic link, the target of the link is added to the layer as well. If the file being linked to is in the RPM db, it will cause a failure since the RPM version doesn't change, but the file is included. So, to work around this, we will track the link targets, then remove them from the list of layer files.


acornett21 (approved):

komish:

komish (approved):
/lgtm

PR #968, opened 2023-04-27
Keep preflight.log and artifacts from being written during tests

bcrochet:
It was missed that the root_test.go should have the BeforeEach that creates the artifacts dir in a temp location, and sets the preflight.log to there as well, then cleans up after.


skattoju (approved):

bcrochet:
Putting this on pause. I don't think it will be necessary (or even wanted) after the viper refactors.

PR #970, opened 2023-04-28
Ignore layer files that have security capability attrs

### LINKED ISSUE #969: Adding setcap to the Dockerfile causes the HasModifiedFiles check in Preflight 1.6.1 to fail
## Bug Description
Adding setcap to the Dockerfile causes the HasModifiedFiles check in Preflight 1.6.1 to fail.

1.6.1

## Steps to Reproduce:

1. Build an image quay.io/tkrishtop/setcap-example:v0.0.2 using the Dockerfile below:

```
FROM registry.access.redhat.com/ubi8/ubi:8.7-1112

RUN microdnf -y install iptables && \
    microdnf clean all

RUN setcap CAP_NET_ADMIN,CAP_NET_RAW+eip /usr/sbin/xtables-nft-multi

CMD [ "cat", "hello world"]
```

2. Run the Preflight check and observe the HasModifiedFiles check fail:

```
podman run -e PFLT_LOGLEVEL=trace quay.io/opdev/preflight:1.6.1 check container quay.io/tkrishtop/setcap-example:v0.0.2

-- snip --

time="2023-04-28T07:38:06Z" level=debug msg="found disallowed modification in layer" file=usr/sbin/xtables-nft-multi layer="sha256:0765ec7ff35e3e077fb047f7a8710cdcef9c2ad08877fce70f25c94492b8a213"
time="2023-04-28T07:38:06Z" level=info msg="check completed" check=HasModifiedFiles result=FAILED
```

3. Confirm that the issue is caused specifically by the setcap line: `setcap CAP_NET_ADMIN,CAP_NET_RAW+eip /usr/sbin/xtables-nft-multi` by building another image quay.io/tkrishtop/setcap-example:v0.0.3 without setcap:

```
FROM registry.access.redhat.com/ubi8/ubi:8.7-1112

RUN microdnf -y install iptables && \
    microdnf clean all

CMD [ "cat", "hello world"]
```

This image passes the HasModifiedFiles check

```
$ podman run -e PFLT_LOGLEVEL=trace quay.io/opdev/preflight:1.6.1 check container quay.io/tkrishtop/setcap-example:v0.0.3

-- snip --

time="2023-04-28T10:10:45Z" level=info msg="check completed" check=HasModifiedFiles result=PASSED
```

## Expected Result
Telco partners often require the use of [setcap](https://linux.die.net/man/7/capabilities): 

CAP_NET_ADMIN - Allow various network-related operations (e.g., setting privileged socket options, enabling multicasting, interface configuration, modifying routing tables).
CAP_NET_RAW - Permit use of RAW and PACKET sockets.

Containers using setcap do nothing wrong or non-standard, they should pass the certification checks. 

## Actual Result
check=HasModifiedFiles result=FAILED

## Additional Context
Since there is no solution on our side, this issue is becoming urgent and has the potential to impact many Telco partners.

cc: @bcrochet @acornett21

bcrochet:
If a file in a layer has had setcap used on it, it will show up in the layer, but be unchanged. However, there will be a PAX record that shows the extended attrs. For now, preflight will ignore these files, unless/until we find a situation where we should be more robust in this check and possibly fail it.

Fixes #969

### DIFF
--- a/internal/policy/container/has_modified_files.go
+++ b/internal/policy/container/has_modified_files.go
@@ -113,7 +113,7 @@ func (p *HasModifiedFilesCheck) gatherDataToValidate(ctx context.Context, imgRef
 
 		layerIDs = append(layerIDs, layerID)
 
-		files, err := generateChangesFor(layer)
+		files, err := generateChangesFor(ctx, layer)
 		if err != nil {
 			return nil, nil, "", err
 		}
@@ -375,7 +375,8 @@ func installedFileMapWithExclusions(ctx context.Context, pkglist []*rpmdb.Packag
 }
 
 // generateChangesFor will check layer for file changes, and will return a list of those.
-func generateChangesFor(layer v1.Layer) ([]string, error) {
+func generateChangesFor(ctx context.Context, layer v1.Layer) ([]string, error) {
+	logger := logr.FromContextOrDiscard(ctx)
 	layerReader, err := layer.Uncompressed()
 	if err != nil {
 		return nil, fmt.Errorf("reading layer contents: %w", err)
@@ -406,6 +407,13 @@ func generateChangesFor(layer v1.Layer) ([]string, error) {
 		if tombstone {
 			basename = basename[len(whiteoutPrefix):]
 		}
+
+		// If there is a capability entry, ignore the file
+		if _, found := header.PAXRecords["SCHILY.xattr.security.capability"]; found {
+			logger.V(log.TRC).Info("security capabilities found in layer tar, ignoring file", "file", header.Name)
+			continue
+		}
+
 		switch {
 		case (header.Typeflag == tar.TypeDir && tombstone) || header.Typeflag == tar.TypeReg:
 			filelist = append(filelist, strings.TrimPrefix(filepath.Join(dirname, basename), "/"))


internal/policy/container/has_modified_files.go
```diff
@@ -406,6 +407,13 @@ func generateChangesFor(layer v1.Layer) ([]string, error) {
 		if tombstone {
 			basename = basename[len(whiteoutPrefix):]
 		}
+
+		// If there is a capability entry, ignore the file
+		if _, found := header.PAXRecords["SCHILY.xattr.security.capability"]; found {
+			logger.V(log.TRC).Info("security capabilities found in layer tar, ignoring file", "file", header.Name)
+			continue
+		}
+
```
lslebodn, line 416: There are files in important rpm packages (shadow-utils) with capabilities and should not be skipped.
One could add extra capabilities to them. BTW `rpm --verify` nicely report changes to capabilities

```
[root@localhost ~]# getcap /usr/bin/new*
/usr/bin/newgidmap cap_setgid=ep
/usr/bin/newuidmap cap_setuid=ep
[root@localhost ~]# rpm -V shadow-utils && echo OK
OK

[root@localhost ~]# setcap CAP_SYS_ADMIN+eip /usr/bin/newgidmap
[root@localhost ~]# rpm -V shadow-utils
........P    /usr/bin/newgidmap
```

man rpm -> `VERIFY OPTIONS`
```
       The format of the output is a string of 9 characters, a possible attri‐
       bute marker:

              c %config configuration file.
              d %doc documentation file.
              g %ghost file (i.e. the file contents are not included in the package payload).
              l %license license file.
              r %readme readme file.

       from  the  package  header,  followed  by the file name.  Each of the 9
       characters denotes the result of a comparison of  attribute(s)  of  the
       file  to  the  value of those attribute(s) recorded in the database.  A
       single "." (period) means the test passed, while a single "?" (question
       mark)  indicates the test could not be performed (e.g. file permissions
       prevent reading).  Otherwise, the (mnemonically  emBoldened)  character
       denotes failure of the corresponding --verify test:

              S file Size differs
              M Mode differs (includes permissions and file type)
              5 digest (formerly MD5 sum) differs
              D Device major/minor number mismatch
              L readLink(2) path mismatch
              U User ownership differs
              G Group ownership differs
              T mTime differs
              P caPabilities differ
```

lslebodn: Or just changes to capabilities should be ignored 
```
[root@localhost ~]# setcap CAP_SYS_ADMIN+eip /usr/bin/newgidmap
[root@localhost ~]# rpm -V shadow-utils
........P    /usr/bin/newgidmap
[root@localhost ~]# rpm --verify --nocaps shadow-utils && echo OK
OK
```
but not other changes
```
[root@localhost ~]# > /usr/bin/newgidmap
[root@localhost ~]# rpm --verify --nocaps shadow-utils && echo OK
S.5....T.    /usr/bin/newgidmap
```

bcrochet: I'm about to send up a new version that will only ignore if there is no RPM db in the layer. If that's the case, then we are making the assumption that it's _just_ the capabilities that are changing, and AFAIK, this is allowed.

lslebodn: > if there is no RPM db in the layer.

Will it detect one of following ?
```
RUN curl -o /usr/bin/newgidmap http://someurl && chmod 755 && setcap CAP_SYS_ADMIN+eip /usr/bin/newgidmap
```

or 
```
RUN curl -o /usr/bin/whoami http://someurl/exploit && chmod 755 && setcap CAP_SYS_ADMIN+eip /usr/bin/whoami
```

> are making the assumption that it's just the capabilities 

Ideally there should be no assumptions and try to do the similar thing as rpm binary does in previous comment with the parameter `--nocaps`.  Keeping the content the same and changing just capabilities  might be OKish
```
[root@localhost ~]# setcap CAP_SYS_ADMIN+eip /usr/bin/newgidmap
[root@localhost ~]# rpm -V shadow-utils
........P    /usr/bin/newgidmap
[root@localhost ~]# rpm --verify --nocaps shadow-utils && echo OK
OK
```

Keeping the **content** of the binary the same and changing **just** capabilities  might be OK-ish.
Skipping file because it has some capabilities set and **not** checking `file size` or `digest` might be a problem.

bcrochet: > > if there is no RPM db in the layer.
> 
> Will it detect one of following ?
> 
> ```
> RUN curl -o /usr/bin/newgidmap http://someurl && chmod 755 && setcap CAP_SYS_ADMIN+eip /usr/bin/newgidmap
> ```
> 
> or
> 
> ```
> RUN curl -o /usr/bin/whoami http://someurl/exploit && chmod 755 && setcap CAP_SYS_ADMIN+eip /usr/bin/whoami
> ```
> 
> > are making the assumption that it's just the capabilities
> 
> Ideally there should be no assumptions and try to do the similar thing as rpm binary does in previous comment with the parameter `--nocaps`. Keeping the content the same and changing just capabilities might be OKish
> 
> ```
> [root@localhost ~]# setcap CAP_SYS_ADMIN+eip /usr/bin/newgidmap
> [root@localhost ~]# rpm -V shadow-utils
> ........P    /usr/bin/newgidmap
> [root@localhost ~]# rpm --verify --nocaps shadow-utils && echo OK
> OK
> ```
> 
> Keeping the **content** of the binary the same and changing **just** capabilities might be OK-ish. Skipping file because it has some capabilities set and **not** checking `file size` or `digest` might be a problem.

I see what you are driving at. I'll try to make this possible. 

However, keep in mind, preflight is explicitly *not* trying to catch bad actors. It's more meant to try to weed out bad practices in creating containers so that partners can be informed that they should not be doing whatever it is that preflight is rejecting.

In this case, your scenarios should fail because it's modifying an RPM file AND setting caps.

bcrochet: And I'm also inclined to push this out as is, and work on the more robust solution so that we can get some partners unstuck. (Specifically telco partners)

acornett21 (approved):

internal/policy/container/has_modified_files.go
acornett21: I'm inclined to agree with @bcrochet. We have partners blocked today, we need to unblock them when we know how, and then iterate on better solutions now that more edge-cases have been identified.

komish (approved):
/lgtm

internal/policy/container/has_modified_files.go
lslebodn: Unblocking partners is very important => +100.
@acornett21 would you mind to share a link to tracker for "better solution" ?

@bcrochet Did you manage to identify why telco partners need to modify rpm from ubi repositories.
I would expect they reported bug to the `iptables` package and rhel maintainer would fix that. (obviously that would still require workaround in preflight to temporarily unblock partners)

PR #971, opened 2023-05-01
Bump operator-sdk to 1.28.1 to unlock the latest Preflight for DCI

tkrishtop:
This PR is to include the replace-tags-with-digests fix: https://github.com/operator-framework/operator-sdk/commit/82b28e85707d2448e1cdd385ce0444ea16e5eb75

PR #972, opened 2023-05-01
updating to operator-sdk v1.28.1

acornett21:
- Updating to v1.28.1 of operator-sdk
- This has the change to use `shas` for scorecard images, so that disconnected envs can work properly again.


acornett21:
- Relates: #971 
Going to close this on, in favor of 971.

PR #958, opened 2023-04-13
Viper refactor

bcrochet:

PR #982, opened 2023-05-18
Resolve race condition in test execution preventing multiple tests from asserting config state

komish:
The ideal long-term state is probably going to be something like https://github.com/redhat-openshift-ecosystem/openshift-preflight/pull/958, but for now, this should resolve the issues in our testing where we occasionally flake.

The issue seems to be a race condition where specs that call `initConfig`, which internally modifies the application-wide `viper.Instance()` would conflict, causing the assertions being made on config items to fail.

This PR modifies initConfig to accept an instance of a viper configuration. In production, we'll always want to pass in `viper.Instance()`, but for testing, we can pass in the instance we want and then make our assertions on a test instance.

### DIFF
--- a/cmd/preflight/cmd/check_container_test.go
+++ b/cmd/preflight/cmd/check_container_test.go
@@ -93,7 +93,7 @@ certification_project_id: mycertid`
 				})
 				It("should still execute with no error", func() {
 					// Make sure that we've read the config file
-					initConfig()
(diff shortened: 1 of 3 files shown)


acornett21 (approved):

bcrochet (approved):
/lgtm

PR #983, opened 2023-05-18
Update golangci-lint to 1.52.2

komish:
Subject line covers it. I randomly ran into an issue with the old version on Golang 1.20 that would cause every ounce of my machine's CPU to get used up and crash the system. Nooo clue what happened.


bcrochet (approved):
/lgtm

acornett21 (approved):

PR #553, opened 2022-04-06
Move OsContentType to CertProjectContainer

bcrochet:
The OsContentType field is actually under the CertProjectContainer,
not the project itself.

Signed-off-by: Brad P. Crochet <brad@redhat.com>

PR #990, opened 2023-05-24
Add check for old-style scratch project

jfrancin:
Support Case: 03509612

Add check for old-style location of Scratch Image definition.

Old Pyxis projects (PIDs starting with 5xxx...) put the Scratch Image status in container.os_content_type = "Scratch Image" rather than in container.type.

It looks like part of the change was added (the older field was defined in types.go), but no code was added in lib/libs.go.


acornett21:
To add to what @bcrochet said, we already were checking for this before and were told to remove it and check what we are checking today. This field is intended to be used for `container` related things, not `project` related things.

- Relates: #553

jfrancin:
Very well; I withdraw the PR.

Issue #1011, opened 2023-06-13
Preflight does not distinguish between scratch and non scratch when trying to read the RPM manifest from disc


yashoza19:
## Bug Description
preflight does not distinguish between scratch and non-scratch when trying to read the RPM manifest from disc and send it to pyxis. 
For scratch images, rpm-manifest.json is not generated.

## Version and Command Invocation
```preflight version 0.0.0 <commit: 1f72aa2f5b111e1e201eadd89e11ad0ca4bc7310>```
```preflight check container quay.io/tigera/operator:v1.29.3```

## Steps to Reproduce:
(How can we reproduce this?)
1) Set Content type to Scratch

## Expected Result
Ignore rpm-manifest.json while submitting the result.

## Actual Result
```level=info msg="preparing results that will be submitted to Red Hat"
Error: could not open file for submission: rpm-manifest.json: open /home/lance/scratch/openshift/artifacts/rpm-manifest.json: no such file or directory
2023/06/12 20:03:47 could not open file for submission: rpm-manifest.json: open /home/lance/scratch/openshift/artifacts/rpm-manifest.json: no such file or directory```

PR #1012, opened 2023-06-13
conditionally calling pyxis.WithRPMManifest based on the policy value in ctx

acornett21:
- Adding `Policy` to `context`
- Retrieving `Policy` from `context` for all policies except `scratch` in the submission flow while reading the `rpm.Manifest`, since scratch policy will not write this file to disk.
- Fixes: #1011

### DIFF
--- a/internal/lib/types.go
+++ b/internal/lib/types.go
@@ -11,13 +11,14 @@ import (
 	"path/filepath"
 	"time"
 
-	"github.com/go-logr/logr"
-
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/artifacts"
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/internal/check"
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/internal/log"
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/internal/policy"
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/internal/pyxis"
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/internal/viper"
+
+	"github.com/go-logr/logr"
 )
 
 // ResultWriter defines methods associated with writing check results.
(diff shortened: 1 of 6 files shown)


internal/lib/types.go
```diff
@@ -161,18 +152,34 @@ func (s *ContainerCertificationSubmitter) Submit(ctx context.Context) error {
 	}
 	defer logfile.Close()
 
+	options := make([]pyxis.CertificationInputOption, 0, 3)
```
bcrochet, line 155: Initialize this in one statement. There is no reason to do the append like this.

internal/policy/policy.go
```diff
@@ -8,3 +10,23 @@ const (
...
 	PolicyRoot      Policy = "root"
 )
+
+// ContextWithPolicy adds Policy p to the context ctx.
+func ContextWithPolicy(ctx context.Context, p Policy) context.Context {
+	return context.WithValue(ctx, policyContextKey, p)
+}
+
+// PolicyFromContext returns the policy from the context, or empty string.
+func PolicyFromContext(ctx context.Context) Policy {
```
bcrochet, line 20: I think I'd prefer something like...

```
policy.NewContext(ctx, p)
```

and just 

```
policy.FromContext(ctx)
```


Otherwise, you're saying

```
policy.PolicyFromContext
```

It's redundant.

bcrochet: I also realize that artifacts should be like this too...

bcrochet (changes requested):
🛑 Couple of requests

internal/lib/types.go
acornett21: Fixed.

internal/policy/policy.go
acornett21: Makes, sense. I've updated this.

komish (approved):
/lgtm

internal/lib/types.go
```diff
@@ -161,18 +152,32 @@ func (s *ContainerCertificationSubmitter) Submit(ctx context.Context) error {
...
+			)
+		}
+		defer rpmManifest.Close()
+
+		// The certification engine writes the rpmManifest for images not based on scratch.
+		options = append(options, pyxis.WithRPMManifest(rpmManifest))
+	}
+
+	// prepare submission. We ignore the error because nil checks for the certProject
+	// are done earlier to prevent panics, and that's the only error case for this function.
```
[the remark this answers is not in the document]
acornett21: I don't see where we ever had `submission, _` must have been before we had the `finalizer` logic and such.

bcrochet (approved):
/lgtm

PR #1013, opened 2023-06-16
Add 'github-actions' config to dependabot

sebrandon1:
Automatically update github actions versions via dependabot.

https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot

Similar to: 
- https://github.com/test-network-function/cnf-certification-test/pull/954
- https://github.com/test-network-function/cnf-certification-test/pull/936

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -6,3 +6,7 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+  - package-ecosystem: github-actions
+    directory: /
+    schedule:
+      interval: weekly


.github/dependabot.yml
```diff
@@ -6,3 +6,11 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+  - package-ecosystem: docker
```
acornett21, line 9: I'm not sure if I like the idea of dependabot updating our image. I'll like @bcrochet and @komish chime in.

sebrandon1: For example, it'll make PRs like: https://github.com/test-network-function/cnf-certification-test/pull/1118

acornett21: Yeah, we used `latest` for UBI and our builder is golang `1.20` at this point. I don't want it updating the builder before we've updated the codebase, and I also don't want the bot updating to `minor` versions of the builder. Especially, since we had an issue last week with the version of `libc` being updated on the base `1.20` version.

sebrandon1: I removed the `Docker` section from the config.  The PR only does github actions updates now.

bcrochet (approved):
/lgtm

acornett21 (approved):

PR #1019, opened 2023-06-20
Make .cache exclusion broader

bcrochet:
Expand the .cache exlusion to be anywhere in usr/ instead of usr/lib64/.


acornett21 (approved):

komish (approved):
/lgtm

PR #1022, opened 2023-06-21
removing unused bundle validate code that relied on operator-sdk binary

acornett21:
- I noticed this code has been unused for awhile since we now rely on the `operator-framework/api` to validate the bundle, and no longer use the binary


komish (approved):
/lgtm

You've got other stuff to remove, though, because CI is failing with `unused` as the reason.

PR #1024, opened 2023-06-22
adding arch to packlist maps

acornett21:
This is incase two archs of the same rpm are in the same image layer.

### DIFF
--- a/internal/policy/container/has_modified_files.go
+++ b/internal/policy/container/has_modified_files.go
@@ -239,7 +239,9 @@ func (p *HasModifiedFilesCheck) validate(ctx context.Context, layerIDs []string,
 			previousOsRelease := strings.Contains(previousPackage.Release, packageDist)
 			currentOsRelease := strings.Contains(currentPackage.Release, packageDist)
 
-			if (previousOsRelease && !currentOsRelease) || (previousPackage.Arch != currentPackage.Arch) {
+			if (previousOsRelease && !currentOsRelease) ||
+				(previousPackage.Arch != currentPackage.Arch && previousPackageVersion == currentPackageVersion) {
+				// todo-adam we need to add a log statement for this failure
 				// If either of these differ, that's a fail
 				return false, nil
 			}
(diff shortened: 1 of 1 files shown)


internal/policy/container/has_modified_files.go
```diff
@@ -239,7 +239,9 @@ func (p *HasModifiedFilesCheck) validate(ctx context.Context, layerIDs []string,
 			previousOsRelease := strings.Contains(previousPackage.Release, packageDist)
 			currentOsRelease := strings.Contains(currentPackage.Release, packageDist)
 
-			if (previousOsRelease && !currentOsRelease) || (previousPackage.Arch != currentPackage.Arch) {
+			if (previousOsRelease && !currentOsRelease) ||
+				(previousPackage.Arch != currentPackage.Arch && previousPackageVersion == currentPackageVersion) {
+				// todo-adam we need to add a log statement for this failure
```
komish, line 244: Do we need to do this?

If we're being more specific with our keys in pkgNameList, then why are we also doing arch filtering here?

Something's not lining up for me.

I'd need to see what some of these values are to be more certain.

acornett21: This logic is outside of the `if` that checks `current == previous`, so this was my attempt the container with this issue to pass. I'm not sure if we need this, or need some other change, since I really don't fully understand all of the logic.

lslebodn: The unique identifier for rpm is `nevra`
* name 
* epoch 
* version
* release
* architecture

One can bump epoch as well. Even though it happens rarely and "nvr" can be still the same; just different epoch
But content of packages can be different
e.g. bind-utils-32:9.16.23-11.el9.x86_64 -> bind-utils-33:9.16.23-11.el9.x86_64

```
[root@5844766e3985 /]# rpm -q --info bind-utils
Name        : bind-utils
**Epoch       : 32**
Version     : 9.16.23
Release     : 11.el9
Architecture: x86_64
Install Date: Tue Jun 27 14:38:25 2023
Group       : Unspecified
Size        : 657326
License     : MPLv2.0
Signature   : RSA/SHA256, Mon Feb 27 21:39:38 2023, Key ID 199e2f91fd431d51
Source RPM  : bind-9.16.23-11.el9.src.rpm
Build Date  : Mon Feb 27 14:33:22 2023
Build Host  : x86-64-02.build.eng.rdu2.redhat.com
Packager    : Red Hat, Inc. <http://bugzilla.redhat.com/bugzilla>
Vendor      : Red Hat, Inc.
URL         : https://www.isc.org/downloads/bind/
Summary     : Utilities for querying DNS name servers
Description :
Bind-utils contains a collection of utilities for querying DNS (Domain
Name System) name servers to find out information about Internet
hosts. These tools will provide you with the IP addresses for given
host names, as well as other information about registered domains and
network addresses.

You should install bind-utils if you need to get information from DNS name
servers.
[root@5844766e3985 /]#
```

PR #1027, opened 2023-06-28
Fix Image signing

komish:
Our release images were not being signed using `cosign` because we were matching the incorrect values in the ReleaseEvent object.

According to the documentation: https://docs.github.com/en/webhooks-and-events/events/github-event-types#releaseevent - the `release` object is populated as a part of the `github.event` object on release events. The `action` key is what contains the state of `published`.

These changes combine the existence of the `release` key along with action being `published` and the calling workflow requesting signing in order to derive whether an image should be signed.

### DIFF
--- a/.github/workflows/build-multiarch.yml
+++ b/.github/workflows/build-multiarch.yml
@@ -48,7 +48,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Install cosign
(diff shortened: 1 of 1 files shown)


komish:
I forgot to WIP this to prevent CI run before approval. My bad

acornett21 (approved):

bcrochet (approved):
/lgtm

PR #1029, opened 2023-06-29
Check policy directly in submit

bcrochet:
There is no need to stuff the policy in the context. In the places that we need it, we can either make another API call, or, as the case with Submit is, we already have what we need in the certProject.

### DIFF
--- a/container/check_container.go
+++ b/container/check_container.go
@@ -58,9 +58,6 @@ func (c *containerCheck) Run(ctx context.Context) (certification.Results, error)
 			return certification.Results{}, fmt.Errorf("%w: %s", preflighterr.ErrCannotResolvePolicyException, err)
 		}
 
-		// adding policy to context to be retrieved later in the submit flow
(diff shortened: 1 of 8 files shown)


komish:
@bcrochet if we're going to walk back stuffing the policy into the context, should we also revert any other changes made in the same PR? E.g. the context helpers in the policy package?

bcrochet:
> @bcrochet if we're going to walk back stuffing the policy into the context, should we also revert any other changes made in the same PR? E.g. the context helpers in the policy package?

I'm ambivalent. You say remove, I'll remove.

bcrochet:
> @bcrochet if we're going to walk back stuffing the policy into the context, should we also revert any other changes made in the same PR? E.g. the context helpers in the policy package?

Removed.

acornett21 (approved):

komish (approved):
/lgtm

PR #1032, opened 2023-07-05
Ignore ghost and artifact RPM files

bcrochet:
If a file is marked as a ghost or artifact, it should be ignored in the context of HasModifiedFiles.


komish (approved):
/lgtm

acornett21 (approved):

PR #1038, opened 2023-07-18
adding rpm architecture of to our maps, for 32 bit and 64 bit rpms

acornett21:
- Adding `pkg.Arch` as a key value to our maps so we are comparing files on an `arch` by `arch` basis, and so we are not overwriting any map value, if multiple arch's of a given rpm exists in a given layer.
- Updating `installedFileMapWithExclusions` method to filter out `directories`
- Updating `installedFileMapWithExclusions` to filter out files from the same rpm if only `architecture` differs
- Updating/Adding docker files to test more success/failures of the HasModifiedFiles check

### DIFF
--- a/internal/policy/container/has_modified_files.go
+++ b/internal/policy/container/has_modified_files.go
@@ -231,17 +231,27 @@ func (p *HasModifiedFilesCheck) validate(ctx context.Context, layerIDs []string,
 				}
 
 				// Nope, nope, nope. File was modified without using RPM
-				logger.V(log.DBG).Info("found disallowed modification in layer", "file", modifiedFile)
+				logger.Info("found disallowed modification in layer", "file", modifiedFile)
 				disallowedModifications = true
 				continue
 			}
-			// Check that release contains the same arch and OS release
+
+			// Check that release contains the same arch, this is to ensure that a package did not get rebuilt with
+			// a different architecture
 			previousOsRelease := strings.Contains(previousPackage.Release, packageDist)
 			currentOsRelease := strings.Contains(currentPackage.Release, packageDist)
 
-			if (previousOsRelease && !currentOsRelease) || (previousPackage.Arch != currentPackage.Arch) {
-				// If either of these differ, that's a fail
-				return false, nil
+			if previousOsRelease && !currentOsRelease {
+				logger.Info("mismatch in OS release", "file", modifiedFile)
+				disallowedModifications = true
+				continue
+			}
+
+			// Check that the architectures for previous version and current version of a given package match
+			if previousPackage.Arch != currentPackage.Arch {
+				logger.Info("mismatch in package architecture", "file", modifiedFile)
+				disallowedModifications = true
+				continue
 			}
 
 			// This appears like an update. This is allowed.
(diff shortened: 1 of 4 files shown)


internal/policy/container/has_modified_files.go
```diff
@@ -403,7 +421,7 @@ func installedFileMapWithExclusions(ctx context.Context, pkglist []*rpmdb.Packag
 				// It is either an explicitly excluded path or directory. Skip it.
 				continue
 			}
-			m[normalized] = fmt.Sprintf("%s-%s-%s", pkg.Name, pkg.Version, pkg.Release)
+			m[normalized] = fmt.Sprintf("%s-%s-%s-%s", pkg.Name, pkg.Version, pkg.Release, pkg.Arch)
```
bcrochet, line 424: Same here.

bcrochet (changes requested):
Couple of changes.

internal/policy/container/has_modified_files.go
```diff
@@ -208,6 +208,22 @@ func (p *HasModifiedFilesCheck) validate(ctx context.Context, layerIDs []string,
 			previousPackage := packageFiles[layerIDs[idx-1]].LayerPackages[previousPackageVersion]
 			currentPackage := ref.LayerPackages[currentPackageVersion]
 
+			var architecture string
+			switch currentPackage.Arch {
```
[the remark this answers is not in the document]
acornett21: or wrap the below code in an `if architecutre == ""` and don't execute that block? I'm not sure really what we are going for here, or all permutations we need to cover.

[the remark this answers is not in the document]
acornett21: I went with a different direction for this, and just wrapped the comparison logic in an `if ok` block since we are looking up in a map for the opposite arch in the map.

internal/policy/container/has_modified_files.go
```diff
@@ -208,6 +208,26 @@ func (p *HasModifiedFilesCheck) validate(ctx context.Context, layerIDs []string,
...
+			// this is needed since the RPM DB, will have an entry for each arch, and potentially the same file listed as `owned` by both arches
+			var architecture string
+			switch currentPackage.Arch {
+			case "i686":
+				architecture = "x86_64"
+			case "x86_64":
+				architecture = "i686"
+			}
+
+			// checking to see if the opposite architectures RPM exists in the same layer, if so we need to ensure that NVRA of both architectures match,
```
komish, line 221:
```suggestion
			// checking to see if the opposite architectures RPM exists in the same layer, if so we need to ensure that NVR of both architectures match,
```

internal/policy/container/has_modified_files.go
```diff
@@ -241,7 +261,9 @@ func (p *HasModifiedFilesCheck) validate(ctx context.Context, layerIDs []string,
 
 			if (previousOsRelease && !currentOsRelease) || (previousPackage.Arch != currentPackage.Arch) {
 				// If either of these differ, that's a fail
-				return false, nil
+				logger.V(log.DBG).Info("mismatch in package architecture", "file", modifiedFile)
```
komish, line 264: Is this phrasing correct? This log line seems to imply only the arch could be wrong here. Should we ever hit this case for x86_64 || i686?

komish (approved):
Seems fine

/lgtm

internal/policy/container/has_modified_files.go
acornett21: We will not hit this case coded for the above code. This was the code block that was originally causing a failure and were never logging this and it caused confusion for cert-ops. I'd like to make this line better, but I need to re-read @bcrochet comments about what this if block covers and comment accordingly, that should yield a better log message.

bcrochet: Let's make them more explicit. It makes sense to split these two so that we get a more granular message.
This is still valid. This is the case that someone replaced a Red Hat package with one of a different arch. OR, they did a rebuild, and the release changed. 

EIther way, it shouldn't pass. But making it separate will make the code more readable.

internal/policy/container/has_modified_files.go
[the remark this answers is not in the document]
acornett21: Going a different direction, and trying to exclude these file(s) before we get to the validation logic.

internal/policy/container/has_modified_files.go
acornett21: I split this into two different `if` conditions and added a unit test to cover both cases.

bcrochet (approved):
/lgtm

komish (approved):
/lgtm

PR #1044, opened 2023-08-04
Refactor options code to its own package

bcrochet:
The Crane options code is now being used in multiple places, so it was time to refactor it out into its own package.

This touches a few places, but remains "backward-compatible".

Any breakage of current functionality or current public API should be considered a bug.


skattoju (approved):

komish (approved):
/lgtm

PR #1074, opened 2023-10-09
Update doc url

aksaswadkar:
Added the correct document URL in the certified_image.go for the certifiedImages check.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ subject to change, and some features are not fully implemented.
 ## Certification Workflow Guide
 
 For the complete container and operator bundle certification workflow instructions, please reference our 
-[official certification documentation](https://access.redhat.com/documentation/en-us/red_hat_software_certification/8.68/html/red_hat_software_certification_workflow_guide/index).
+[official certification documentation](https://access.redhat.com/documentation/en-us/red_hat_software_certification/8.69/html/red_hat_software_certification_workflow_guide/index).
 
 ## Requirements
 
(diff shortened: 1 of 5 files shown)


acornett21:
HI @aksaswadkar It looks like you raised two pr's with overlapping info I assume you will close the other pr? Can you also search for `red_hat_software_certification` in the entire project and update all links to `8.68`? Thanks!

aksaswadkar:
Hi @acornett21 I closed the other PR and sure, I will find and update the URL to the latest version.

acornett21 (approved):

aksaswadkar:
Hi @acornett21 Updated the URLs to 8.69, please review.

README.md
```diff
@@ -16,7 +16,7 @@ subject to change, and some features are not fully implemented.
 ## Certification Workflow Guide
 
 For the complete container and operator bundle certification workflow instructions, please reference our 
-[official certification documentation](https://access.redhat.com/documentation/en-us/red_hat_software_certification/8.56/html/red_hat_software_certification_workflow_guide/index).
+[official certification documentation](https://access.redhat.com/documentation/en-us/red_hat_software_certification/8.69/html/red_hat_software_certification_workflow_guide/index).
```
bcrochet, line 19: I won't keep this from going in, but is there instead a 'latest' URL that we can use in preflight, so that it does not require a code change and a release in order for these to be updated?

bcrochet (commented):
Have a question, but otherwise lgtm.

README.md
aksaswadkar: Hi @bcrochet there is no 'latest' URL, each URL has got the new version(8.6x). It'd be great to not have the version in the URL then we won't have to update it when there is a new version which is every 3-4 weeks.

bcrochet: We might want to figure out a redirector then that is external to preflight.

bcrochet (approved):
/lgtm

PR #1100, opened 2023-12-01
Library: add ability to get a test list

edcdavid:
Extracts the check listing component from the operatorCheck and containerCheck Run functions and exposes a "List" function. This code using the library to list the checks (including full metadata) that are executed by the Run function without actually running the checks. 
A new Err field is added to the Result struct to report the detail of errors (generated by  ExecuteChecks, see https://github.com/redhat-openshift-ecosystem/openshift-preflight/blob/31f7e815872037c910b57e16a50489cb1e566174/internal/engine/engine.go#L234) via the library. The same field is also used construct a Results struct in case the test could not be ran at all. 
see: https://github.com/test-network-function/cnf-certification-test/pull/1684

### DIFF
--- a/certification/results.go
+++ b/certification/results.go
@@ -12,7 +12,10 @@ type openshiftClusterVersion = runtime.OpenshiftClusterVersion
 type Result struct {
 	check.Check
 	ElapsedTime time.Duration
-	Err         error
+	// Err contains the error a check itself throws if it failed to run.
+	// If populated, the expectation is that this Result is in the
+	// Results{}.Errors slice.
+	Err error
 }
 
 type Results struct {
--- a/certification/results.go
+++ b/certification/results.go
@@ -15,7 +15,7 @@ type Result struct {
 	// Err contains the error a check itself throws if it failed to run.
 	// If populated, the expectation is that this Result is in the
 	// Results{}.Errors slice.
-	Err error
+	err error
 }
 
 type Results struct {
@@ -27,4 +27,13 @@ type Results struct {
 	Failed            []Result
 	Errors            []Result
 	Warned            []Result
+}
+
+func (r Result) Error() error {
+	return r.err
+}
+
+func (r *Result) WithError(err error) *Result {
+	r.err = err
+	return r
 }
--- a/certification/results.go
+++ b/certification/results.go
@@ -12,6 +12,7 @@ type openshiftClusterVersion = runtime.OpenshiftClusterVersion
 type Result struct {
 	check.Check
 	ElapsedTime time.Duration
+	Err         error
 }
 
 type Results struct {
--- a/container/check_container.go
+++ b/container/check_container.go
@@ -72,7 +72,7 @@ func (c *containerCheck) Run(ctx context.Context) (certification.Results, error)
 // pyxis information is provided.
 func (c *containerCheck) List(ctx context.Context) (policy.Policy, []check.Check, error) {
 	if c.image == "" {
-		return policy.PolicyNone, []check.Check{}, preflighterr.ErrImageEmpty
+		return "", []check.Check{}, preflighterr.ErrImageEmpty
 	}
 
 	pol := policy.PolicyContainer
@@ -88,7 +88,7 @@ func (c *containerCheck) List(ctx context.Context) (policy.Policy, []check.Check
 
 		override, err := lib.GetContainerPolicyExceptions(ctx, p)
 		if err != nil {
-			return policy.PolicyNone, []check.Check{}, fmt.Errorf("%w: %s", preflighterr.ErrCannotResolvePolicyException, err)
+			return "", []check.Check{}, fmt.Errorf("%w: %s", preflighterr.ErrCannotResolvePolicyException, err)
 		}
 
 		pol = override
--- a/container/check_container.go
+++ b/container/check_container.go
@@ -97,6 +97,7 @@ func (c *containerCheck) resolve(ctx context.Context) error {
 		if err != nil {
 			return fmt.Errorf("%w: %s", preflighterr.ErrCannotInitializeChecks, err)
 		}
+		c.resolved = true
 	}
 	return nil
 }
--- a/container/check_container.go
+++ b/container/check_container.go
@@ -91,8 +91,8 @@ func (c *containerCheck) resolve(ctx context.Context) error {
 
 		c.policy = override
 	}
-	var err error
-	c.checks, err = engine.InitializeContainerChecks(ctx, c.policy, engine.ContainerCheckConfig{
+
+	newChecks, err := engine.InitializeContainerChecks(ctx, c.policy, engine.ContainerCheckConfig{
 		DockerConfig:           c.dockerconfigjson,
 		PyxisAPIToken:          c.pyxisToken,
 		CertificationProjectID: c.certificationProjectID,
@@ -100,6 +100,7 @@ func (c *containerCheck) resolve(ctx context.Context) error {
 	if err != nil {
 		return fmt.Errorf("%w: %s", preflighterr.ErrCannotInitializeChecks, err)
 	}
+	c.checks = newChecks
 	c.resolved = true
 
 	return nil
@@ -108,8 +109,7 @@ func (c *containerCheck) resolve(ctx context.Context) error {
 // List the available container checks. Policy exceptions will be resolved if the proper
 // pyxis information is provided.
 func (c *containerCheck) List(ctx context.Context) (policy.Policy, []check.Check, error) {
-	err := c.resolve(ctx)
-	return c.policy, c.checks, err
+	return c.policy, c.checks, c.resolve(ctx)
 }
 
 // hasPyxisData returns true of the values necessary to make a pyxis
--- a/container/check_container.go
+++ b/container/check_container.go
@@ -65,40 +65,43 @@ func (c *containerCheck) Run(ctx context.Context) (certification.Results, error)
 }
 
 func (c *containerCheck) resolve(ctx context.Context) error {
-	if !c.resolved {
-		if c.image == "" {
-			return preflighterr.ErrImageEmpty
-		}
-
-		c.policy = policy.PolicyContainer
-
-		// If we have enough Pyxis information, resolve the policy.
-		if c.hasPyxisData() {
-			p := pyxis.NewPyxisClient(
-				c.pyxisHost,
-				c.pyxisToken,
-				c.certificationProjectID,
-				&http.Client{Timeout: 60 * time.Second},
-			)
-
-			override, err := lib.GetContainerPolicyExceptions(ctx, p)
-			if err != nil {
-				return fmt.Errorf("%w: %s", preflighterr.ErrCannotResolvePolicyException, err)
-			}
-
-			c.policy = override
-		}
-		var err error
-		c.checks, err = engine.InitializeContainerChecks(ctx, c.policy, engine.ContainerCheckConfig{
-			DockerConfig:           c.dockerconfigjson,
-			PyxisAPIToken:          c.pyxisToken,
-			CertificationProjectID: c.certificationProjectID,
-		})
-		if err != nil {
-			return fmt.Errorf("%w: %s", preflighterr.ErrCannotInitializeChecks, err)
-		}
-		c.resolved = true
+	if c.resolved {
+		return nil
 	}
+
+	if c.image == "" {
+		return preflighterr.ErrImageEmpty
+	}
+
+	c.policy = policy.PolicyContainer
+
+	// If we have enough Pyxis information, resolve the policy.
+	if c.hasPyxisData() {
+		p := pyxis.NewPyxisClient(
+			c.pyxisHost,
+			c.pyxisToken,
+			c.certificationProjectID,
+			&http.Client{Timeout: 60 * time.Second},
+		)
+
+		override, err := lib.GetContainerPolicyExceptions(ctx, p)
+		if err != nil {
+			return fmt.Errorf("%w: %s", preflighterr.ErrCannotResolvePolicyException, err)
+		}
+
+		c.policy = override
+	}
+	var err error
+	c.checks, err = engine.InitializeContainerChecks(ctx, c.policy, engine.ContainerCheckConfig{
+		DockerConfig:           c.dockerconfigjson,
+		PyxisAPIToken:          c.pyxisToken,
+		CertificationProjectID: c.certificationProjectID,
+	})
+	if err != nil {
+		return fmt.Errorf("%w: %s", preflighterr.ErrCannotInitializeChecks, err)
+	}
+	c.resolved = true
+
 	return nil
 }
 
--- a/container/check_container.go
+++ b/container/check_container.go
@@ -40,7 +40,7 @@ func NewCheck(image string, opts ...Option) *containerCheck {
 func (c *containerCheck) Run(ctx context.Context) (certification.Results, error) {
 	pol, checks, err := c.List(ctx)
 	if err != nil {
-		return certification.Results{}, fmt.Errorf("%w: %s", preflighterr.ErrCannotInitializeChecks, err)
+		return certification.Results{}, fmt.Errorf("%w: %s", preflighterr.ErrCannotListChecks, err)
 	}
 
 	cfg := runtime.Config{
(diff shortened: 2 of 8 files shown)


edcdavid:
> Hi @edcdavid preflight already has the function to list the checks via `preflight list-checks`, is this what you are looking for?



> Hi @edcdavid preflight already has the function to list the checks via `preflight list-checks`, is this what you are looking for?

Hi @acornett21 thanks! I believe that would work. I missed it before . I will close the PR

edcdavid (commented):
@acornett21, Can I use InitializeOperatorChecks directly to get the other metadata fields for the check?

container/check_container.go
```diff
@@ -38,34 +38,7 @@ func NewCheck(image string, opts ...Option) *containerCheck {
...
-	}
-
-	checks, err := engine.InitializeContainerChecks(ctx, pol, engine.ContainerCheckConfig{
-		DockerConfig:           c.dockerconfigjson,
-		PyxisAPIToken:          c.pyxisToken,
-		CertificationProjectID: c.certificationProjectID,
-	})
+	pol, checks, err := c.List(ctx)
 	if err != nil {
 		return certification.Results{}, fmt.Errorf("%w: %s", preflighterr.ErrCannotInitializeChecks, err)
```
acornett21, line 43: This code needs to go into the new `List` function and add a new if condition here to return err

container/check_container.go
```diff
@@ -38,8 +38,41 @@ func NewCheck(image string, opts ...Option) *containerCheck {
 // pyxis information is provided. Calls should add a relevant ArtifactWriter to the context if they
 // wish to work with artifact files written by checks.
 func (c *containerCheck) Run(ctx context.Context) (certification.Results, error) {
+	pol, checks, err := c.List(ctx)
+	if err != nil {
+		return certification.Results{}, fmt.Errorf("%w: %s", preflighterr.ErrCannotListChecks, err)
```
acornett21, line 43: This will fix the failing tests

```suggestion
		return certification.Results{}, err
```

operator/check_operator.go
```diff
@@ -35,29 +36,9 @@ func NewCheck(image, indeximage string, kubeconfig []byte, opts ...Option) *oper
...
-		ScorecardServiceAccount: c.scorecardServiceAccount,
-		IndexImage:              c.indeximage,
-		DockerConfig:            c.dockerConfigFilePath,
-		Channel:                 c.operatorChannel,
-		Kubeconfig:              c.kubeconfig,
-	})
+	_, checks, err := c.List(ctx)
 	if err != nil {
-		return certification.Results{}, fmt.Errorf("%w: %s", preflighterr.ErrCannotInitializeChecks, err)
+		return certification.Results{}, fmt.Errorf("%w: %s", preflighterr.ErrCannotListChecks, err)
```
acornett21, line 41: This will fix the failing tests
```suggestion
		return certification.Results{}, err
```

errors/errors.go
```diff
@@ -9,4 +9,5 @@ var (
 	ErrImageEmpty                   = errors.New("image is empty")
 	ErrCannotResolvePolicyException = errors.New("cannot resolve policy exception")
 	ErrCannotInitializeChecks       = errors.New("unable to initialize checks")
+	ErrCannotListChecks             = errors.New("unable to list checks")
```
acornett21, line 12: since we don't wrap errors from `Run` this can be removed.

container/check_container.go
edcdavid: fixed

operator/check_operator.go
edcdavid: fixed

errors/errors.go
edcdavid: fixed

Apetree100122 (approved):
/approved

container/check_container.go
```diff
@@ -55,7 +88,7 @@ func (c *containerCheck) Run(ctx context.Context) (certification.Results, error)
 
 		override, err := lib.GetContainerPolicyExceptions(ctx, p)
 		if err != nil {
-			return certification.Results{}, fmt.Errorf("%w: %s", preflighterr.ErrCannotResolvePolicyException, err)
+			return policy.PolicyNone, []check.Check{}, fmt.Errorf("%w: %s", preflighterr.ErrCannotResolvePolicyException, err)
```
bcrochet, line 91: Since Policy is just a type alias, and not a typedef, you can just return "" here without the addition of
"PolicyNone".

I would prefer that.

internal/policy/policy.go
```diff
@@ -7,4 +7,5 @@ const (
 	PolicyContainer Policy = "container"
 	PolicyScratch   Policy = "scratch"
 	PolicyRoot      Policy = "root"
+	PolicyNone      Policy = "none"
```
bcrochet, line 10: See comment about this being a type alias.

operator/check_operator.go
```diff
@@ -91,6 +72,35 @@ func (c operatorCheck) Run(ctx context.Context) (certification.Results, error) {
 	return eng.Results(ctx), nil
 }
 
+// List the available operator checks.
+func (c operatorCheck) List(ctx context.Context) (policy.Policy, []check.Check, error) {
+	switch {
+	case c.image == "":
+		return policy.PolicyNone, []check.Check{}, preflighterr.ErrImageEmpty
```
bcrochet, line 79: See comment about type alias. Return "" for policy on error. It shouldn't _look_ like a valid policy. And if err != nil, none of the other values returned should be expected to be valid.

operator/check_operator.go
```diff
@@ -35,29 +36,9 @@ func NewCheck(image, indeximage string, kubeconfig []byte, opts ...Option) *oper
...
-		ScorecardImage:          c.scorecardImage,
-		ScorecardWaitTime:       c.scorecardWaitTime,
-		ScorecardNamespace:      c.scorecardNamespace,
-		ScorecardServiceAccount: c.scorecardServiceAccount,
-		IndexImage:              c.indeximage,
-		DockerConfig:            c.dockerConfigFilePath,
-		Channel:                 c.operatorChannel,
-		Kubeconfig:              c.kubeconfig,
-	})
+	_, checks, err := c.List(ctx)
```
bcrochet, line 39: This right here drives the point home about returning "". In case there was a question. :)

bcrochet (changes requested):
Comments about PolicyNone. I'm open to discussion on it.

container/check_container.go
edcdavid: fixed

internal/policy/policy.go
edcdavid: fixed

operator/check_operator.go
edcdavid: fixed

komish (changes requested):
Hey, thanks for your work on this @edcdavid!

I have some challenges with the List method as it stands here, but it's a bit tricky to get in-lined, so I'm posting it here instead.

Effectively, the `List` and `Run` methods both require the resolution of the options provided by the user to the (operator/container)Check instantiators. You've moved that logic from the Run method (which was the primary function of interest) into the List method.

Now that have more-than-1 methods of interest on these structs, I think it makes sense to offload early "initialization" logic into its own method on each of these classes, and then make sure it's executed by all other methods (e.g. `List`, `Run`). 

In an ideal world, we'd do this in `NewCheck`, but since it doesn't return an error today, and adding an error to the return signature breaks the API, we can just do this a different way to preserve compatibility.

I'd say it makes sense to make an unexported method (say, `resolve`) that does all of the early logic, resolving checks based on the user-provided Options.

I'd also like to see this run only once, no matter how many times it's called. Maybe just a struct key on (operator/containerCheck) `resolved` with true/false, and have `resolve` only run if it's set to false, regardless of where it's called.

It may not be efficient, but it should preserve backwards compatibility, and I don't think we run any risk of resolving without all of the details we need because I don't think users can change options of the check once instantiated.

certification/results.go
```diff
@@ -12,6 +12,7 @@ type openshiftClusterVersion = runtime.OpenshiftClusterVersion
 type Result struct {
 	check.Check
 	ElapsedTime time.Duration
+	Err         error
```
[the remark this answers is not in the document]
edcdavid: fixed

container/check_container.go
```diff
@@ -38,8 +38,41 @@ func NewCheck(image string, opts ...Option) *containerCheck {
...
+		ManifestListDigest: c.manifestListDigest,
+	}
+	eng, err := engine.New(ctx, checks, nil, cfg)
+	if err != nil {
+		return certification.Results{}, err
+	}
+
+	if err != nil {
+		return certification.Results{}, err
+	}
```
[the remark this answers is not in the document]
edcdavid: fixed

edcdavid:
> Hey, thanks for your work on this @edcdavid!
> 
> I have some challenges with the List method as it stands here, but it's a bit tricky to get in-lined, so I'm posting it here instead.
> 
> Effectively, the `List` and `Run` methods both require the resolution of the options provided by the user to the (operator/container)Check instantiators. You've moved that logic from the Run method (which was the primary function of interest) into the List method.
> 
> Now that have more-than-1 methods of interest on these structs, I think it makes sense to offload early "initialization" logic into its own method on each of these structs, and then make sure it's executed by all other methods (e.g. `List`, `Run`).
> 
> In an ideal world, we'd do this in `NewCheck`, but since it doesn't return an error today, and adding an error to the return signature breaks the API, we can just do this a different way to preserve compatibility.
> 
> I'd say it makes sense to make an unexported method (say, `resolve`) that does all of the early logic, resolving checks based on the user-provided Options[1].
> 
> I'd also like to see this run only once, no matter how many times it's called. Maybe just a struct key on (operator/containerCheck) `resolved` with true/false, and have `resolve` only run if it's set to false, regardless of where it's called.
> 
> It may not be efficient, but it should preserve backwards compatibility, and I don't think we run any risk of resolving without all of the details we need because I don't think users can change options of the check once instantiated.
> 
> Let me know if I can clarify anything.
> 
> [1] E.g.
> 
> ```go
> func (c *containerCheck) resolve(...) error {
>     if !c.resolved {
>         // check the options
>         // try to resolve the checks based on the options
>         // store relevant things internally
>         // etc.
>     }
>     return nil // happy path!
> ```

Thanks for the comments, I pushed a change to follow your suggestion

komish:
@edcdavid Looking good, thank you!

Can you extend the **container/check_container_test.go** and **operator/check_operator_test.go** suites to account for these changes? Specifically:

- Making sure the `resolve` calls populate the internal struct keys that are expected. I don't think you have to test error cases as those should be covered by the existing tests against Run.

- Adding a test case for `List`, to make sure it produces what you want.

edcdavid:
@komish Thanks, I added the missing unit tests

komish (approved):
/lgtm

operator/check_operator.go
```diff
@@ -84,11 +65,44 @@ func (c operatorCheck) Run(ctx context.Context) (certification.Results, error) {
 		return certification.Results{}, err
 	}
 
-	if err != nil {
-		return certification.Results{}, err
+	return eng.Results(ctx), nil
+}
+
+func (c *operatorCheck) resolve(ctx context.Context) error {
+	if !c.resolved {
```
bcrochet, line 72: Can we switch this around to keep all of this code out of an indented block?

container/check_container.go
```diff
@@ -95,6 +64,51 @@ func (c *containerCheck) Run(ctx context.Context) (certification.Results, error)
 	return eng.Results(ctx), nil
 }
 
+func (c *containerCheck) resolve(ctx context.Context) error {
+	if !c.resolved {
```
bcrochet, line 68: Same as other block. Please swap this around to keep the indentation to a minimum.

operator/check_operator.go
edcdavid: fixed

container/check_container.go
edcdavid: fixed

container/check_container.go
```diff
@@ -55,44 +86,30 @@ func (c *containerCheck) Run(ctx context.Context) (certification.Results, error)
...
+			return fmt.Errorf("%w: %s", preflighterr.ErrCannotResolvePolicyException, err)
 		}
 
-		pol = override
+		c.policy = override
 	}
-
-	checks, err := engine.InitializeContainerChecks(ctx, pol, engine.ContainerCheckConfig{
+	var err error
+	c.checks, err = engine.InitializeContainerChecks(ctx, c.policy, engine.ContainerCheckConfig{
```
bcrochet, line 95: I'm not a fan of updating fields directly like this, especially when there is a possible error involved. The field should only be updated one it has been determined that there is no error.

container/check_container.go
```diff
@@ -55,44 +86,30 @@ func (c *containerCheck) Run(ctx context.Context) (certification.Results, error)
...
-	}
+	return nil
+}
 
-	return eng.Results(ctx), nil
+// List the available container checks. Policy exceptions will be resolved if the proper
+// pyxis information is provided.
+func (c *containerCheck) List(ctx context.Context) (policy.Policy, []check.Check, error) {
+	err := c.resolve(ctx)
+	return c.policy, c.checks, err
```
bcrochet, line 112: If you aren't going to check the error here, you might as well just inline it all into the return statement.

operator/check_operator.go
```diff
@@ -84,11 +65,47 @@ func (c operatorCheck) Run(ctx context.Context) (certification.Results, error) {
...
+		return preflighterr.ErrImageEmpty
+	case c.kubeconfig == nil:
+		return preflighterr.ErrKubeconfigEmpty
+	case c.indeximage == "":
+		return preflighterr.ErrIndexImageEmpty
+	}
+
+	c.policy = policy.PolicyOperator
+	var err error
+	c.checks, err = engine.InitializeOperatorChecks(ctx, c.policy, engine.OperatorCheckConfig{
```
bcrochet, line 87: Same comment as before.  Don't set fields directly if there is an error possibility.

container/check_container.go
edcdavid: fixed

container/check_container.go
edcdavid: fixed

komish (approved):
/lgt

certification/results.go
```diff
@@ -12,6 +12,10 @@ type openshiftClusterVersion = runtime.OpenshiftClusterVersion
 type Result struct {
 	check.Check
 	ElapsedTime time.Duration
+	// Err contains the error a check itself throws if it failed to run.
+	// If populated, the expectation is that this Result is in the
+	// Results{}.Errors slice.
+	Err error
```
bcrochet, line 18: I'd like to propose that instead of exposing a raw struct field, we extend this via a new method and a private field.

```
type Result struct {
	check.Check
	ElapsedTime time.Duration
	err error
}

func (r Result) Error() error {
	return r.err
}

func (r *Result) WithError(err error) *Result {
	r.err = err
	return r
}
```

internal/engine/engine.go
```diff
@@ -231,7 +231,7 @@ func (c *craneEngine) ExecuteChecks(ctx context.Context) error {
 
 		if err != nil {
 			logger.WithValues("result", "ERROR", "err", err.Error()).Info("check completed", "check", executedCheck.Name())
-			c.results.Errors = appendUnlessOptional(c.results.Errors, certification.Result{Check: executedCheck, ElapsedTime: checkElapsedTime})
+			c.results.Errors = appendUnlessOptional(c.results.Errors, certification.Result{Check: executedCheck, ElapsedTime: checkElapsedTime, Err: err})
```
bcrochet, line 234: And to round it out:

```
			result := certification.Result{Check: executedCheck, ElapsedTime: checkElapsedTime}
			c.results.Errors = appendUnlessOptional(c.results.Errors, *result.WithError(err))
```

bcrochet (changes requested):
I bit about moving the field to be private. LMKWYT.

certification/results.go
edcdavid: ok for me, fixed

internal/engine/engine.go
edcdavid: fixed

bcrochet (approved):
/lgtm

PR #1102, opened 2023-12-06
remove chart-verifier (helm) experimental subcommand

acornett21:
## Motivation
Currently, it is impossible to keep `preflight's` dependencies up to date, since there is a divergence in indirect dependencies that are used in `helm` vs `k8s` examples are:
- opentelemetry
- oras 
- others?

Removal will allow preflight to stay current on dependencies and up to date on CVE's.

## Changes
- Removed `helm` cmd.
- Removed `expermental` cmd as a separate commit incase we have use case in the future.


komish (approved):
/lgtm

PR #1103, opened 2023-12-07
Notify users if there is a new version of preflight

### LINKED ISSUE #1043: Add Version Lookup to Notify Users of `latest` Release
When trying to troubleshoot with partners, usually our first questions is `Are you on the latest version?`, which adds some unnecessary back and forth (and delay in support) before we can get to the actual issue they are experiencing. It would be nice if we did a lookup when a command is executed, `check container` for example, and let the user know if they are on the latest version of prelfight or not.

Below is the rough outline of how I see this going:
- Call GH `/release/latest` endpoint as a non-blocking call (non blocking means, if it errors/timeouts we still continue)
- Compare `tag_name` from the above payload with the `version` built into the code
   -  It would also be nice to check the commit, but I don't see this in the `/release` endpoint
- If the binary version `!=` to the version from the GH payload, display the GH url to download the latest version

N/A

## Additional context.
Maybe it makes sense that this lives at the `root` cmd level and is done no matter what cmd is executed? Though I'm not sure that is best. For now, I think we keep this informative and see how things go, and potentially make this required for `--submit` at least in the `check container` workflow.

skattoju:
fixes #1043 

- Currently this is only done when running check container.
- There is an option for specifying a Github auth token in case rate limiting becomes an issue. As of now its 60 request per hour unauthenticated which goes up to 5000 when using an auth token. In any case remaining requests are printed in the logs
- The timeout for requests from the Github client to Github is set to one second to avoid waiting too long in case Github is slow to respond.


cmd/preflight/cmd/check_container.go
```diff
@@ -198,7 +204,7 @@ func checkContainerRunE(cmd *cobra.Command, args []string, runpreflight runPrefl
 
 func checkContainerPositionalArgs(cmd *cobra.Command, args []string) error {
 	if len(args) != 1 {
-		return fmt.Errorf("a container image positional argument is required")
+		return fmt.Errorf("a container image positional argument is required, found %v", args)
```
acornett21, line 207: Is this updated needed for this task? It seems like an enhancement that might need it's own PR.

cmd/preflight/cmd/check_container.go
```diff
@@ -257,6 +270,38 @@ func validateCertificationProjectID(cmd *cobra.Command, args []string) error {
 	return nil
 }
 
+// checkForNewerReleaseVersion checks if there is a newer release available
+func checkForNewerReleaseVersion(cmd *cobra.Command) {
+	logger, err := logr.FromContext(cmd.Context())
```
acornett21, line 275: I think using `logr.FromContextOrDiscard` might be a better choice, that way we don't have to worry about errors. But I'll let other's chime in as well.

version/version.go
```diff
@@ -25,3 +38,35 @@ type VersionContext struct {
 func (vc *VersionContext) String() string {
 	return fmt.Sprintf("%s <commit: %s>", vc.Version, vc.Commit)
 }
+
+func (vc *VersionContext) LatestReleasedVersion(cmd *cobra.Command, svc VersionClient) (*github.RepositoryRelease, error) {
+	ctx := cmd.Context()
+	logger, err := logr.FromContext(ctx)
```
acornett21, line 44: same `logr.FromContextOrDiscard` here

version/version.go
```diff
@@ -2,11 +2,20 @@
...
+	"github.com/go-logr/logr"
+	"github.com/google/go-github/v57/github"
+	semvc "github.com/hashicorp/go-version"
+	"github.com/spf13/cobra"
+)
 
 var (
 	projectName = "github.com/redhat-openshift-ecosystem/openshift-preflight"
-	version     = "unknown"
+	version     = "0.0.1"
```
acornett21, line 18: Was this just changed so that other tests became valid?

acornett21 (changes requested):

cmd/preflight/cmd/check_container.go
skattoju: i figured no harm in keeping it 🤷‍♂️

version/version.go
skattoju: yes this was for tests

cmd/preflight/cmd/check_container.go
```diff
@@ -257,6 +270,38 @@ func validateCertificationProjectID(cmd *cobra.Command, args []string) error {
...
+		client = github.NewClient(&http.Client{
+			// Timeout in 1s in case Github is slow to respond
+			Timeout: time.Second * 1,
+		}).WithAuthToken(ghToken)
+	} else {
+		client = github.NewClient(&http.Client{
+			// timeout in 1s in case Github is slow to respond
+			Timeout: time.Second * 1,
+		})
+	}
```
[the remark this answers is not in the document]
skattoju: I wanted to keep the time out short in order not to interfere with CI when github has issues

version/version.go
skattoju: Its not implicit ? Ok will add a comment 👍

acornett21: I don't think it's implicit, since it's been unknown or `0.0.0` locally forever.

version/version.go
```diff
@@ -2,11 +2,20 @@
...
 
-import "fmt"
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/go-logr/logr"
+	"github.com/google/go-github/v57/github"
+	semvc "github.com/hashicorp/go-version"
```
komish, line 12: Is there any particular version why we're using this, and not https://pkg.go.dev/golang.org/x/mod/semver, or https://github.com/Masterminds/semver

version/version.go
```diff
@@ -25,3 +38,35 @@ type VersionContext struct {
 func (vc *VersionContext) String() string {
 	return fmt.Sprintf("%s <commit: %s>", vc.Version, vc.Commit)
 }
+
+func (vc *VersionContext) LatestReleasedVersion(cmd *cobra.Command, svc VersionClient) (*github.RepositoryRelease, error) {
```
komish, line 42: This shouldn't need  `*cobra.Command` to function. If you need a context, accept a context.

With that said, I would say we probably don't want this to log in any way. It likely makes more sense to emit an error here, and have the caller figure out what to do with it.

komish: Also, does this need to be a mod of the VersionContext struct for any reason? I think there's certainly value here in hard-coding some of the relevant values for our project ( e.g. the org/project values), but we're not accomplishing that by adding this as a method to the struct.

Make this an library function that accepts the parameters it needs to execute, then call it with the proper parameters provided at your call point(s).

If we make this just a generic library call that accepts the client you need to make to resolve the version, it may also simplify testing of this logic.

version/version.go
komish: I would rather this remain its original value "unknown". We inject this at build time, and we only truly care about whether someone is using an outdated version in cases where they've used our release binaries, which have this value populated.

Which test needs this to change?

komish (changes requested):
Couple of changes/questions.

Overall, I'm wondering if we might be better off updating a cookie file somewhere on each release than querying the GitHub releases page.

I don't particularly love having to pass a GitHub API token into the tool for behavior that isn't core to Preflight's execution. That seems distracting to the end user, and I'm fairly confident it would be hit frequently since the only real client information GitHub would have is IP space. I think. 

At any rate, I would much prefer making like an HTTP call that's hard-code to pull a value. Maybe it's a gh-pages thing. Not sure what I would want to see.

Some food for thought. cc @acornett21 @bcrochet

acornett21:
> I don't particularly love having to pass a GitHub API token into the tool for behavior that isn't core to Preflight's execution.
@komish The API key should be optional and not required. It would only be if a user experiences rate limiting from GH. Say they are using preflight in their CI system (which already calls GH a bunch).

version/version.go
acornett21: Is one of these already used in a project we manage? If so I'd vote for picking that one.

cmd/preflight/cmd/check_container_test.go
```diff
@@ -33,6 +25,14 @@ import (
...
 	. "github.com/onsi/gomega"
 	"github.com/onsi/gomega/types"
+
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/artifacts"
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/certification"
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/internal/check"
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/internal/cli"
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/internal/formatters"
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/internal/lib"
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/internal/viper"
```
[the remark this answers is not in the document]
skattoju: looks like just the order is changing

version/version.go
skattoju: changed to Masterminds 👍

version/version.go
skattoju: I have tests comparing semver style versions thats why I changed it from unknown to 0.0.1. May be 0.0.0 is better ?

cmd/preflight/cmd/check_container.go
```diff
@@ -6,10 +6,20 @@ import (
...
 	"fmt"
 	"io"
+	"net/http"
 	"os"
 	"path/filepath"
 	rt "runtime"
 	"strings"
+	"time"
+
+	"github.com/go-logr/logr"
```
bcrochet, line 16: This should not move like it is. Is the IDE moving this block up? Or is the linter suggesting that this be moved ahead of "local" imports?

The order should be:
stdlib

project

3rd party

cmd/preflight/cmd/check_container_test.go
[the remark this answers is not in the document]
bcrochet: Well, it's not correct. The project-local imports should be after stdlib, not last. Please figure out if it's your IDE doing that, and update its settings accordingly.

version/version.go
bcrochet: Agreed on the context here. cobra.Command should not have made it out of cmd package.

cmd/preflight/cmd/check_container.go
```diff
@@ -237,9 +243,16 @@ func checkContainerPositionalArgs(cmd *cobra.Command, args []string) error {
 	return nil
 }
 
+// validateConditions run all pre-run functions
+func validateConditions(cmd *cobra.Command, args []string) error {
+	err := validateCertificationProjectID()
+	checkForNewerReleaseVersion(cmd)
+	return err
+}
+
```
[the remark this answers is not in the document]
bcrochet: I concur here. It seems odd here to call another function between setting err and returning it. Not sure about the use of MatchAll. But it wouldn't hurt to try it. Otherwise, at the very least I would suggest something like

```suggestion
// validateConditions run all pre-run functions
func validateConditions(cmd *cobra.Command, args []string) error {
	checkForNewerReleaseVersion(cmd)
	return validateCertificationProjectID()
}

```

cmd/preflight/cmd/check_container.go
```diff
@@ -257,6 +270,34 @@ func validateCertificationProjectID(cmd *cobra.Command, args []string) error {
...
+			Timeout: time.Second * 1,
+		}).WithAuthToken(ghToken)
+	} else {
+		client = github.NewClient(&http.Client{
+			// timeout in 1s in case Github is slow to respond
+			Timeout: time.Second * 1,
+		})
+	}
+	// check if a newer release is available
+	latestRelease, err := version.Version.LatestReleasedVersion(cmd, client.Repositories)
```
bcrochet, line 292: This is awkward. LatestReleasedVersion, as @komish stated elsewhere, really doesn't need to be a part of Version.
```suggestion
	latestRelease, err := version.LatestRelease(cmd.Context(), client.Repositories)
```

cmd/preflight/cmd/check_container.go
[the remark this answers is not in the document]
bcrochet: I really don't think we should add the entire GitHub client solely to use the token. Do we really foresee people running into the GitHub rate limit?

We should probably just be hitting it with net/http without bringing in the github client library. If we get reports that people are hitting it, we can address it then.

bcrochet (changes requested):
A few concurrences and some additional questions.

cmd/preflight/cmd/check_container.go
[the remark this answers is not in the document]
skattoju: I think we said earlier that hitting the rate limit would happen quite often to people running the tool in CI behind a VPN and that was the reason to use GH client and have the option to pass a token.

skattoju:
Closing this since version enforcement will be implemented in the backend (pyxis) ref: CLOUDWF-10041

PR #1108, opened 2023-12-18
Ensure all disconnected annotations are accounted for when reviewing guidelines

komish:
We recently added support for new annotations for various infrastructure features, including disconnected/restricted network support, but our informational check that reviews the CSV's implementation of the guidelines for disconnected support was not accounting for the new annotation.

As a result, we'd emit logs indicating that "this operator didn't support deployment into environments with restricted networks" even if the new annotation was present, which was misleading. 

This PR properly checks both annotations, and in the presence of either with a value indicating support for deployment into these environments, executes the logic to see how closely the maintainer adheres to the guidelines.

This also adds relevant tests, and relabels/documents the older annotation as "legacy".

### DIFF
--- a/internal/csv/csv.go
+++ b/internal/csv/csv.go
@@ -24,7 +24,7 @@ const (
 	CSIAnnotation                    = "features.operators.openshift.io/csi"
 )
 
-// SupportsDisconnected accepts a stringified list of supported features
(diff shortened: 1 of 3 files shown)


bcrochet (approved):
/lgtm

skattoju (approved):

PR #1116, opened 2024-01-12
updating operator framework api dependency

acornett21:
- Updating `operator-framework/api` and migrating to new validations, since the original is depricated

**Note:** OperatorHubV2Validator, StandardCapabilitiesValidator and StandardCategoriesValidator provide the same functionality as the deprecated OperatorHubValidator validator


internal/bundle/bundle.go
```diff
@@ -44,7 +44,11 @@ func Validate(ctx context.Context, imagePath string) (*Report, error) {
 	}
 	validators := validation.DefaultBundleValidators.WithValidators(
 		validation.AlphaDeprecatedAPIsValidator,
-		validation.OperatorHubValidator,
+		// OperatorHubV2Validator, StandardCapabilitiesValidator and StandardCategoriesValidator provide the same functionality
+		// as the deprecated OperatorHubValidator validator
```
komish, line 48: Nit: We probably don't need these comments here. It was useful for my review, but I don't think it needs to live in the source.

komish (approved):
Left a nit. Non-blocking.

/lgtm

internal/bundle/bundle.go
bcrochet: Agreed. Should just live in the commit message.

bcrochet (commented):
Agreement on the comment.

komish (approved):
/lgtm

Issue #1083, opened 2023-10-16
Docs URLs have a hard-coded version


bcrochet:
The hard-coded version makes updates to the docs (which occur every 3-4 weeks) require a rebuild and a release. 

A couple of options:
1. Somehow get a 'latest' url that always points to the latest docs
2. Have an external redirector take care of that mapping, and have preflight point to it.

Other options?

acornett21:
I wish the cert docs had a `latest` URI value like other Red Hat docs do, but whatever platform they have, only has hard-coded version values. It seems that those change anytime there is an update anywhere in the documentation, which makes it really tricky for us to track.

The last issue, they also changed where the pages existed. I'm not really sure how we overcome this, but just giving some backstory to think about as well.

acornett21:
This would be a good first issue for someone to work with the docs team to see if they can create a `latest` redirect URL that we could use in our code.

Garv978:
Hi! I'd like to work on this issue. I'll investigate the current hard-coded documentation URLs and whether a stable latest redirect is available, then update the relevant references and add/update tests as appropriate. Thanks!

PR #1120, opened 2024-01-29
updating documentation URLs

aksaswadkar:
Partner certification documentation has been updated to use the current year as the version number.  This simplifies and creates annual static links.

I believe this also solves the https://github.com/redhat-openshift-ecosystem/openshift-preflight/issues/1083

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ subject to change, and some features are not fully implemented.
 ## Certification Workflow Guide
 
 For the complete container and operator bundle certification workflow instructions, please reference our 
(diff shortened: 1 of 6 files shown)


acornett21:
@aksaswadkar Are you saying that the docs folks have changed their URL scheme, and it will change based on the year now? Do you know if they'll ever offer a `/latest` route? Also, Can you rebase off of main? That will fix the test failure. Thanks for the PR.

aksaswadkar:
Hi @acornett21 Yes, the docs URLs are based on the year now. Not sure about the `/latest` though.

acornett21 (approved):

Issue #1104, opened 2023-12-11
Update `RequiredAnnotations` Check to be Enforced


acornett21:
## Describe the solution you'd like.
The `RequiredAnnotations` check is first going to go into production as a `warning` (level: warn) and then at a laster date it will need to `enforced` (level: best) and becoming blocking. Moving the check to `level: best` will make the check enforced, and fail if a partner does not have all of the proper annotations/values.

## Additional info/changes.
When we move this check to be enforced, we should also update the `RestrictedNetworkAware` check's logic to use the new disconnected value of `features.operators.openshift.io/disconnected: "true"`.  This check needs to move to this new annotation, since the old one will become deprecated in the Openshift Console.

If other maintainers feel this should be updated separately, that is fine as well, the main thing for now is that this gets documented.

PR #1129, opened 2024-02-22
moving RequiredAnnotations check to LevelBest so that it is enforced.

acornett21:
- Fixes: #1104 
- Moving `RequiredAnnotations` check to LevelBest so that it will be enforced.


bcrochet (approved):
/lgtm

komish (approved):
/lgtm

Issue #1125, opened 2024-02-06
Preflight dies if Pyxis rejects a submission due to unsupported architecture


jfrancin:
## Bug Description

If you attempt to run Preflight on a multi-image manifest digest that includes an unsupported architecture (such as 32-bit arm), Preflight will exit out with an error when Pyxis rejects the submission. 

## Version and Command Invocation
preflight version 1.8.1 <commit: 4c9ea7cbd33fb7deaf0521690a6ab42c8879c55a>

## Steps to Reproduce:
(How can we reproduce this?)
1) preflight check container  quay.io/calico/node-driver-registrar:v3.24.6 --pyxis-api-token <redacted> --certification-project-id 64c01702093679e0f47fa153 --submit 
2)
3)

This is a multi-architecture image manifest.  The repo is open for read: you should be able to copy the image manifest to a more suitable registry for debugging.

## Expected Result
Preflight would carry on with the scanning of the other architectures in the image manifest
-OR- would skip the unsupported architecture entirely (perhaps outputting a suitable message about the skip)

## Actual Result
Preflight exited with an error on the arm architecture, the first one (alphabetically) in the manifest digest:

```
ryzen1> preflight check container  quay.io/calico/node-driver-registrar:v3.24.6 --pyxis-api-token <redacted> --certification-project-id 64c01702093679e0f47fa153 --submit 
time="2024-02-06T11:41:52-05:00" level=info msg="certification library version" version="1.8.1 <commit: 4c9ea7cbd33fb7deaf0521690a6ab42c8879c55a>"
time="2024-02-06T11:41:53-05:00" level=info msg="running checks for quay.io/calico/node-driver-registrar:v3.24.6 for platform arm"
time="2024-02-06T11:41:55-05:00" level=info msg="target image" image="quay.io/calico/node-driver-registrar:v3.24.6"
time="2024-02-06T11:41:56-05:00" level=info msg="check completed" check=HasLicense result=FAILED
time="2024-02-06T11:41:56-05:00" level=info msg="check completed" check=HasUniqueTag result=PASSED
time="2024-02-06T11:41:56-05:00" level=info msg="check completed" check=LayerCountAcceptable result=PASSED
time="2024-02-06T11:41:56-05:00" level=info msg="check completed" check=HasRequiredLabel result=FAILED
time="2024-02-06T11:41:56-05:00" level=info msg="detected empty USER. Presumed to be running as root"
[...]

PR #1131, opened 2024-03-01
add allowedArchitectures function to filter out unsupported arches

acornett21:
- Fixes: #1125 
- Adding a new allowedArchitectures function to filter out unsupported architectures from manifestlist submissions.
- This will unblock users that have `arm` or other unsupported arches to continue with the submission process of supported architectures.

### DIFF
--- a/cmd/preflight/cmd/check_container.go
+++ b/cmd/preflight/cmd/check_container.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"path/filepath"
 	rt "runtime"
+	"slices"
 	"strings"
 
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/artifacts"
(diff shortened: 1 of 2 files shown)


bcrochet (approved):
/lgtm

cmd/preflight/cmd/check_container.go
```diff
@@ -435,6 +437,10 @@ func platformsToBeProcessed(cmd *cobra.Command, cfg *runtime.Config) ([]string,
 				// This must be an attestation manifest. Skip it.
 				continue
 			}
+			if !slices.Contains(allowedArchitectures(), img.Platform.Architecture) {
+				// The user has a architecture type in the manifest list that we do not support.
```
skattoju, line 441: should we log that an unsupported architecture was encountered ?

skattoju (approved):

cmd/preflight/cmd/check_container.go
acornett21: I feel like logging here is more noise then anything, we don't log for any of the other conditions in this code. Also, since preflight only ships for the 4 arches, and OCP does as well, I think partners know we won't publish anything other then what we support.

PR #1134, opened 2024-03-12
updating dependencies away from retracted module versions

acornett21:
- Updating OpenShift dependencies to `main` and away from retracted module versions


bcrochet (approved):
/lgtm

PR #1141, opened 2024-04-10
handling pyxis version check errors to display preflight release url

acornett21:
- Relates: ##1135

- Handling the two possible 400 series error messages.
- Displaying the latest preflight release URL to the user.

Sample Pyxis Responses:
Screenshot from 2024-03-12 12-39-13
Screenshot from 2024-03-12 12-38-52

### DIFF
--- a/internal/pyxis/pyxis.go
+++ b/internal/pyxis/pyxis.go
@@ -4,14 +4,17 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
+	"strings"
 
 	"github.com/go-logr/logr"
 	"github.com/shurcooL/graphql"
 
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/internal/log"
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/version"
 )
 
 const (
@@ -409,6 +412,7 @@ func (p *pyxisClient) updateProject(ctx context.Context, certProject *CertProjec
 }
 
 func (p *pyxisClient) createTestResults(ctx context.Context, testResults *TestResults) (*TestResults, error) {
+	logger := logr.FromContextOrDiscard(ctx)
 	b, err := json.Marshal(testResults)
 	if err != nil {
 		return nil, fmt.Errorf("could not marshal test results: %w", err)
@@ -431,6 +435,14 @@ func (p *pyxisClient) createTestResults(ctx context.Context, testResults *TestRe
 	}
 
 	if ok := checkStatus(resp.StatusCode); !ok {
+		// checking to see if the users is using an unsupported versions
+		// if so display preflights latest download url to the user
+		if resp.StatusCode == http.StatusBadRequest &&
+			(strings.Contains(string(body), "not recognized") || strings.Contains(string(body), "not supported")) {
+			logger.Error(errors.New("invalid preflight version, please download the latest version and re-submit"),
+				"release-url", fmt.Sprintf("https://%s/releases/latest", version.Version.Name))
+		}
+
 		return nil, fmt.Errorf(
 			"status code: %d: body: %s",
 			resp.StatusCode,
--- a/internal/pyxis/pyxis_suite_test.go
+++ b/internal/pyxis/pyxis_suite_test.go
@@ -157,6 +157,12 @@ func pyxisTestResultsHandler(ctx context.Context) http.HandlerFunc {
 		switch {
 		case request.Header["X-Api-Key"][0] == "my-bad-testresults-api-token":
 			response.WriteHeader(http.StatusUnauthorized)
+		case request.Header["X-Api-Key"][0] == "my-unrecognized-preflight-version":
+			response.WriteHeader(http.StatusBadRequest)
+			mustWrite(response, `{"detail":"not recognized"}`)
+		case request.Header["X-Api-Key"][0] == "my-unsupported-preflight-version":
+			response.WriteHeader(http.StatusBadRequest)
+			mustWrite(response, `{"detail":"not supported"}`)
 		case request.Method == http.MethodPatch && request.Header["X-Api-Key"][0] == "my-bad-results-patch-api-token":
 			response.WriteHeader(http.StatusInternalServerError)
 		default:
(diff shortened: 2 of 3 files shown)


internal/pyxis/pyxis.go
```diff
@@ -430,16 +434,25 @@ func (p *pyxisClient) createTestResults(ctx context.Context, testResults *TestRe
 		return nil, fmt.Errorf("could not read body: %w", err)
 	}
 
+	bodyAsString := string(body)
```
bcrochet, line 437: Nit: Not a big fan of including the type in the var name. Just call it `body`, rename the above to just `b`, and be done with it.

internal/pyxis/pyxis_suite_test.go
```diff
@@ -157,6 +157,12 @@ func pyxisTestResultsHandler(ctx context.Context) http.HandlerFunc {
 		switch {
 		case request.Header["X-Api-Key"][0] == "my-bad-testresults-api-token":
 			response.WriteHeader(http.StatusUnauthorized)
+		case request.Header["X-Api-Key"][0] == "my-unrecognized-perflight-version":
```
bcrochet, line 160: Are you misspelling `preflight` on purpose?

internal/pyxis/pyxis_suite_test.go
```diff
@@ -157,6 +157,12 @@ func pyxisTestResultsHandler(ctx context.Context) http.HandlerFunc {
...
+			response.WriteHeader(http.StatusBadRequest)
+			mustWrite(response, `{"detail":"not recognized"}`)
+		case request.Header["X-Api-Key"][0] == "my-unsupported-perflight-version":
```
bcrochet, line 163: Ditto.

internal/pyxis/submit_test.go
```diff
@@ -293,6 +293,38 @@ var _ = Describe("Pyxis Submit", func() {
 		})
 	})
 
+	Context("createTestResults 400 version not supported", func() {
+		BeforeEach(func() {
+			pyxisClient.APIToken = "my-unsupported-perflight-version"
```
bcrochet, line 298: `preflight`

internal/pyxis/submit_test.go
```diff
@@ -293,6 +293,38 @@ var _ = Describe("Pyxis Submit", func() {
...
+					Expect(err).To(HaveOccurred())
+					Expect(certResults).To(BeNil())
+				})
+			})
+		})
+	})
+
+	Context("createTestResults 400 version not recognized", func() {
+		BeforeEach(func() {
+			pyxisClient.APIToken = "my-unrecognized-perflight-version"
```
bcrochet, line 314: Ditto

internal/pyxis/pyxis.go
bcrochet, line 437:
```suggestion
	body := string(b)
```

bcrochet (changes requested):
Questions...

internal/pyxis/pyxis_suite_test.go
acornett21: Good catch, this was _not_ on purpose.

internal/pyxis/pyxis.go
acornett21: I updated this to `respBody`, since we already have `b` and `resp` and `body` elsewhere in this method. LMK what you think.

internal/pyxis/pyxis.go
```diff
@@ -425,21 +429,30 @@ func (p *pyxisClient) createTestResults(ctx context.Context, testResults *TestRe
...
+		// if so display preflights latest download url to the user
+		if resp.StatusCode == http.StatusBadRequest &&
+			(strings.Contains(body, "not recognized") || strings.Contains(body, "not supported")) {
+			logger.Error(errors.New("invalid preflight version, please please download the latest version and re-submit"),
+				"release-url", fmt.Sprintf("https://%s/releases/latest", version.Version.Name))
+		}
+
 		return nil, fmt.Errorf(
-			"status code: %d: body: %s",
+			"status code: %d: respBody: %s",
```
bcrochet, line 448: This should probably read `body`

internal/pyxis/pyxis.go
```diff
@@ -425,21 +429,30 @@ func (p *pyxisClient) createTestResults(ctx context.Context, testResults *TestRe
...
 			resp.StatusCode,
-			string(body))
+			body)
 	}
 
 	newTestResults := TestResults{}
-	if err := json.Unmarshal(body, &newTestResults); err != nil {
-		return nil, fmt.Errorf("could not unmarshal body: %s: %w", string(body), err)
+	if err := json.Unmarshal(respBody, &newTestResults); err != nil {
+		return nil, fmt.Errorf("could not unmarshal body: %s: %w", body, err)
```
bcrochet, line 455: I don't think we should be mixing these here... Use `body` for both or `respBody` for both.

acornett21: This is why just having `string(body)` everywhere is more clear then trying to have another variable IMO.

acornett21:
@komish @bcrochet I'm not really sure what to do with this PR. IMO using `string(body)` everywhere is the clearest path forward, since the intent of the type being used is in the code, trying to add another variable intrudes confusion of value/type of any given line of code that needs both the `byte` value and the `string` value.

internal/pyxis/pyxis.go
```diff
@@ -431,6 +435,14 @@ func (p *pyxisClient) createTestResults(ctx context.Context, testResults *TestRe
 	}
 
 	if ok := checkStatus(resp.StatusCode); !ok {
+		// checking to see if the users is using an unsupported versions
+		// if so display preflights latest download url to the user
+		if resp.StatusCode == http.StatusBadRequest &&
+			(strings.Contains(string(body), "not recognized") || strings.Contains(string(body), "not supported")) {
+			logger.Error(errors.New("invalid preflight version, please please download the latest version and re-submit"),
```
bcrochet, line 442: Are we super begging them to do it?

bcrochet (changes requested):
Typo?

internal/pyxis/pyxis.go
acornett21: Yes, we are being extra nice! I'll remove the dup! Good catch!

bcrochet (approved):
/lgtm

komish (approved):
/lgtm

PR #1142, opened 2024-04-15
adding new release action to send release information to pyxis

acornett21:
## Info
- Adding new workflow to send release information to pyxis.
   - This workflow does not get triggered on `pre-releases` (ie the checkbox on the release page). This will allow us to release a `alpha/beta` release that we can test in lower environments, but is not enabled/enforced by pyxis.

## Tests
- Success of a [release](https://github.com/acornett21/openshift-preflight/actions/runs/8653395800/job/23729060653)
- Prerelease getting [ignored](https://github.com/acornett21/openshift-preflight/actions/runs/8690084938)


.github/workflows/add-release-info-to-pyxis.yml
```diff
@@ -0,0 +1,42 @@
...
+
+jobs:
+  call-pyxis-endpoint:
+    name: Add Release Info
+    runs-on: ubuntu-latest
+    steps:
+      - name: Save Certificate and Key
+        run: |
+          echo -en "${{ secrets.certificate }}" > /tmp/preflight-release.crt
+          echo -en "${{ secrets.key }}" > /tmp/preflight-release.key
```
mrhillsman, line 31: These echos will not expose the keys right? Also is there a need to delete these even though they are in tmp is there a legitimate concern they could be exposed if not manually removed?

komish: @acornett21 it does not hurt to remove it.

Create a final step for this job that `rm -f <files> || true` and make sure its if expression is set to `always()` so it runs regardless of any failure in this workflow.

.github/workflows/build-release.yml
```diff
@@ -13,16 +13,19 @@ env:
...
-    strategy: 
+    strategy:
       matrix:
         architecture: [amd64,ppc64le,arm64,s390x]
         platform: [linux]
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
-    - name: Set Env Tags
+    - name: Set Release Tag
```
komish, line 23: Why is this changing?

.github/workflows/build-release.yml
```diff
@@ -13,16 +13,19 @@ env:
...
       run: echo RELEASE_TAG=$(echo $GITHUB_REF | cut -d '/' -f 3) >> $GITHUB_ENV
 
+    - name: Set COMMIT_SHA/Version
+      run: echo COMMIT_SHA=$(git rev-parse HEAD) >> $GITHUB_ENV
```
komish, line 27: Any reason why we're running `git` instead of relying on the environment for the event that triggered this workflow (e.g. the same as we do in the previous step, which uses `$GITHUB_REF`, we could be using `$GITHUB_SHA`

komish: Further, I think we could probably skip this step entirely if `$GITHUB_SHA` is in the right format already. That is, we do the previous step for `$RELEASE_TAG` because the format of that value is not what we want, and we have to manipulate it to get the value we want. I'm guessing `$GITHUB_SHA` can probably go straight into your workflow_call, no?

komish (changes requested):
Mostly LGTM, couple of small things.

.github/workflows/build-release.yml
acornett21: We are using `git` in other workflows so I thought we'd want to use it here as well, also since there are multiple SHA's in the GH ENV, it's confusing which one is actually `git rev-parse HEAD` value, and I didn't want there to be a mismatch. Are we sure we want `GITHUB_SHA` and not `COMMIT_SHA` or even `GITHUB_WORKFLOW_SHA`?

.github/workflows/build-release.yml
acornett21: This is changing for clarity, since we are setting multiple ENVs, if the step for the commit goes away, this can be reverted.

.github/workflows/build-release.yml
komish: COMMIT_SHA is what you've set, right? WORKFLOW_SHA is documented as the sha of the workflow file itself. GITHUB_SHA is what you want.

If you want to leave the `git` call, I'll leave it up to you. In practice, what you're doing is letting actions/checkout dictate what sha you have, and that defaults to GITHUB_SHA from what I can tell.

Should be fine. Just something to note.

.github/workflows/build-release.yml
komish: Sorry, I misread which was the incoming change and which was outgoing. This looks fine to me.

.github/workflows/build-release.yml
acornett21: I'm fine for reducing code that we need to maintain, if we're sure `GITHUB_SHA` is what we want I'll change to that. I'll double check that this is the same as have we have in our make file that is executing `git re-parse HEAD` that's present on/in the binary.

acornett21: I verified that the sha's produced in our make file matches `GITHUB_SHA`

### Info In Binary
```
$ wget https://github.com/acornett21/openshift-preflight/releases/download/0.0.32/preflight-linux-amd64
$ ./preflight-linux-amd64 --version
preflight version 0.0.32 <commit: 124521d9840b619a404d16e2bf5eaa8d83335925>
```

### Info In `curl` in GHA
```
curl -s \
  --cert '/tmp/preflight-release.crt' \
  --key '/tmp/preflight-release.key' \
  --pass '***' \
  -H 'Content-Type: application/json' \
  -d '{"commit":"124521d9840b619a404d16e2bf5eaa8d8335925","enabled_for_testing":true,"name":"github.com/redhat-openshift-ecosystem/openshift-preflight","version":"0.0.32"}' \
  -X POST \
  'https://pyxis.preprod.api.redhat.com/v1/tools' | jq
```

.github/workflows/add-release-info-to-pyxis.yml
acornett21: @komish I added a cleanup, let me know if this is correct.

### Info In GHA
```
Run rm -f /tmp/preflight-release.crt /tmp/preflight-release.key || true
  rm -f /tmp/preflight-release.crt /tmp/preflight-release.key || true
  shell: /usr/bin/bash -e {0}
  env:
    CERT_FILE_LOCATION: /tmp/preflight-release.crt
    KEY_FILE_LOCATION: /tmp/preflight-release.key
```

.github/workflows/add-release-info-to-pyxis.yml
```diff
@@ -0,0 +1,51 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Save Certificate and Key
+        run: |
+          echo -en "${{ secrets.certificate }}" > ${{ env.CERT_FILE_LOCATION }}
+          echo -en "${{ secrets.key }}" > ${{ env.KEY_FILE_LOCATION }}
+
+      - name: Call Pyxis API
+        run: |
+          curl -s \
```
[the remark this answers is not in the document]
acornett21: @lslebodn Are you saying to add `--fail-with-body` so that we can see the HTTP 500 series errors?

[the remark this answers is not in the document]
lslebodn: curl return 0 in shell with any status code.
```
sh-5.2$ curl -I https://httpbin.org/status/503
HTTP/2 503 
date: Wed, 17 Apr 2024 15:44:36 GMT
content-type: text/html; charset=utf-8
content-length: 0
server: gunicorn/19.9.0
access-control-allow-origin: *
access-control-allow-credentials: true

sh-5.2$ curl https://httpbin.org/status/503
sh-5.2$ echo $?
0
```
and you would not notice it with with `curl -s`  
so IMHO it is better to fail with curl for unexpected status code.
```
sh-5.2$ curl --fail https://httpbin.org/status/503
curl: (22) The requested URL returned error: 503
sh-5.2$ echo $?
22
sh-5.2$ curl --fail https://httpbin.org/status/404
curl: (22) The requested URL returned error: 404
sh-5.2$ echo $?
22
```

And it is simple one-liner change. But feel free to keep current version.
just my 2 cents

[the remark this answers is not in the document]
komish: +1, good catch @lslebodn

komish (approved):
/lgtm

skattoju (approved):
/lgtm

PR #1147, opened 2024-04-20
fix: close rpmdb

testwill:


bcrochet:
We cannot accept completely anonymous PRs. Closing.

PR #1151, opened 2024-04-29
reordering submit flow to create test results as first action

acornett21:
- Reordering submit flow to create test results early, then update them in the case where we have successfully created an image to link to the results.
- We need not worry about re-try, or re-run (of preflight like in other API calls) or any of the other things we worry about for other pyxis routes, since pyxis/connect will just ignore (not show?) any results not linked to any images.
- `createTestResults` was modified to account for both `PATCH/POST`, since there would be less duplicate code, if others are happier with two distinct methods, I'm happy to refactor that way as well.
- JIRA: EET-3965

### DIFF
--- a/internal/pyxis/pyxis.go
+++ b/internal/pyxis/pyxis.go
@@ -445,6 +445,44 @@ func (p *pyxisClient) createTestResults(ctx context.Context, testResults *TestRe
 	return &newTestResults, nil
 }
 
+func (p *pyxisClient) updateTestResults(ctx context.Context, testResults *TestResults) (*TestResults, error) {
+	b, err := json.Marshal(testResults)
+	if err != nil {
+		return nil, fmt.Errorf("could not marshal test results: %w", err)
+	}
+
+	req, err := p.newRequestWithAPIToken(ctx, http.MethodPatch, p.getPyxisURL(fmt.Sprintf("projects/certification/test-results/id/%s", testResults.ID)), bytes.NewReader(b))
+	if err != nil {
+		return nil, fmt.Errorf("could not create new request: %w", err)
+	}
+
+	resp, err := p.Client.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("could not update test results in pyxis: %w", err)
+	}
+
+	defer resp.Body.Close()
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, fmt.Errorf("could not read body: %w", err)
+	}
+
+	if ok := checkStatus(resp.StatusCode); !ok {
+		return nil, fmt.Errorf(
+			"status code: %d: body: %s",
+			resp.StatusCode,
+			string(body))
+	}
+
+	newTestResults := TestResults{}
+	if err := json.Unmarshal(body, &newTestResults); err != nil {
+		return nil, fmt.Errorf("could not unmarshal body: %s: %w", string(body), err)
+	}
+
+	return &newTestResults, nil
+}
+
 func (p *pyxisClient) createArtifact(ctx context.Context, artifact *Artifact) (*Artifact, error) {
 	logger := logr.FromContextOrDiscard(ctx)
 
--- a/internal/pyxis/submit.go
+++ b/internal/pyxis/submit.go
@@ -15,8 +15,20 @@ var defaultRegistryAlias = "docker.io"
 func (p *pyxisClient) SubmitResults(ctx context.Context, certInput *CertificationInput) (*CertificationResults, error) {
 	var err error
 
-	certProject := certInput.CertProject
 	certImage := certInput.CertImage
+	// You must have an existing repository.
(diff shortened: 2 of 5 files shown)


internal/pyxis/pyxis.go
```diff
@@ -408,12 +408,21 @@ func (p *pyxisClient) updateProject(ctx context.Context, certProject *CertProjec
...
 	}
-	req, err := p.newRequestWithAPIToken(ctx, http.MethodPost, p.getPyxisURL(fmt.Sprintf("projects/certification/id/%s/test-results", p.ProjectID)), bytes.NewReader(b))
+
+	var pyxisURL string
+	switch httpMethod {
+	case http.MethodPatch:
+		pyxisURL = fmt.Sprintf("projects/certification/test-results/id/%s", testResults.ID)
+	default:
+		pyxisURL = fmt.Sprintf("projects/certification/id/%s/test-results", p.ProjectID)
+	}
```
komish, line 423: NIT: The function signature indicates it accepts _any_ HTTP method but then only actually executes MethodPatch or MethodPost. If a user passes in http.MethodGet, wouldn't this behave unexpectedly (falling back to http.MethodPost)?

Considered a nit because this is an internal function.

internal/pyxis/submit.go
```diff
@@ -15,6 +16,13 @@ var defaultRegistryAlias = "docker.io"
 func (p *pyxisClient) SubmitResults(ctx context.Context, certInput *CertificationInput) (*CertificationResults, error) {
 	var err error
 
+	// Create the test results, so we can fail fast if version check throws error.
+	testResults := certInput.TestResults
+	testResults, err = p.createOrUpdateTestResults(ctx, testResults, http.MethodPost)
+	if err != nil {
+		return nil, fmt.Errorf("could not create test results: %v", err)
+	}
+
```
komish, line 25: I understand this is here to fail fast, but it would seem that if there are other failures (e.g. if `certImage.Repositories == 0` a few lines down), wouldn't we have created a TestResult for this account before we've asserted whether their content is valid?

internal/pyxis/submit.go
```diff
@@ -108,19 +116,22 @@ func (p *pyxisClient) SubmitResults(ctx context.Context, certInput *Certificatio
...
 
-	// Create the test results.
-	testResults := certInput.TestResults
-	testResults.ImageID = certImage.ID
-	testResults, err = p.createTestResults(ctx, testResults)
+	// Update the test results with the certification image id to link the results to the image.
+	updatedTestResults := &TestResults{
+		ID:      testResults.ID,
+		ImageID: certImage.ID,
+	}
```
komish, line 123: Why create a net-new instance here?

internal/pyxis/pyxis.go
acornett21: It's a good call out, I was trying to avoid 2 separate functions, that are almost identical. Though based on your other comment about a net-new instance, maybe it should be 2 functions, and the `updateTestResults` struct is encapsulated in that method, like we have in other places. Though 2 functions for this will have alot of duplicated code, I'm fine either way though.

internal/pyxis/submit.go
acornett21: Good catch, that should probably be first, since it was lost in the comments, I'll move it to be first thing.

internal/pyxis/submit.go
acornett21: This is net-new, since for a PATCH, we do not want/need to send the entire payload on the wire, just the bits we care about.

internal/pyxis/pyxis.go
komish: The function returns an error, so we could potentially just case out the methods we want and return an error if the method is invalid.

Allda (approved):
The new logic looks good to me.

komish (approved):
/lgtm

internal/pyxis/pyxis.go
```diff
@@ -408,12 +408,23 @@ func (p *pyxisClient) updateProject(ctx context.Context, certProject *CertProjec
 	return &newCertProject, nil
 }
 
-func (p *pyxisClient) createTestResults(ctx context.Context, testResults *TestResults) (*TestResults, error) {
+func (p *pyxisClient) createOrUpdateTestResults(ctx context.Context, testResults *TestResults, httpMethod string) (*TestResults, error) {
```
bcrochet, line 411: This seems... clunky. Why not have a createTestResults, and an updateTestResults? httpMethod seems to me to expose far to much about the implementation.

bcrochet (changes requested):
Question about the methods.

internal/pyxis/pyxis.go
acornett21: I've switch to adding a new `updateTestResults` method, which should resolve this.

komish (approved):
/lgtm

bcrochet (approved):
/lgtm

PR #1158, opened 2024-05-07
fix test results update to send entire payload

acornett21:
- It seems pyxis requires the entire payload, so sending it for the PATCH call as well.
- Relates: JIRA-ISVISSUE-605


skattoju (approved):
/lgtm

komish (approved):
/lgtm

Issue #1156, opened 2024-05-07
Investigate log.SetLogger for controller-runtime


bcrochet:
Witnessed an error during DeployableByOlm:

```
[controller-runtime] log.SetLogger(...) was never called; logs will not be displayed.
```

bcrochet:
This could be that we just call it with a "discard" logger.

PR #1160, opened 2024-05-07
adding a call to controller-runtimes SetLogger and io.Discard

acornett21:
- This avoids noise from controller runtime
- Fixes: #1156

### DIFF
--- a/cmd/preflight/cmd/root.go
+++ b/cmd/preflight/cmd/root.go
@@ -16,6 +16,7 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	spfviper "github.com/spf13/viper"
+	ctrl "sigs.k8s.io/controller-runtime"
 )
 
 var configFileUsed bool
@@ -133,5 +134,12 @@ func preRunConfig(cmd *cobra.Command, args []string) {
 
 	logger := logrusr.New(l)
 	ctx := logr.NewContext(cmd.Context(), logger)
+
+	// Setting the controller-runtime logger to a no-op logger by default,
+	// unless debug mode is enabled. This is because the controller-runtime
+	// logger is *very* verbose even at info level. This is not really needed,
+	// but otherwise we get a warning from the controller-runtime.
+	ctrl.SetLogger(logr.Discard())
+
 	cmd.SetContext(ctx)
 }


cmd/preflight/cmd/root.go
```diff
@@ -133,5 +135,12 @@ func preRunConfig(cmd *cobra.Command, args []string) {
 
 	logger := logrusr.New(l)
 	ctx := logr.NewContext(cmd.Context(), logger)
+
+	// Setting the controller-runtime logger to a no-op logger by default,
+	// unless debug mode is enabled. This is because the controller-runtime
+	// logger is *very* verbose even at info level. This is not really needed,
+	// but otherwise we get a warning from the controller-runtime.
+	ctrl.SetLogger(zap.New(zap.WriteTo(io.Discard)))
```
bcrochet, line 143: Does it need to be zap? Can't we use logr for this?

bcrochet (commented):
Quick question.

komish:
1. Is this a result of a controller-runtime version bump, or have these log lines been there for a while? 
2. If this is new, should this not be a bug filed against controller-runtime?
3. Should this be run closer to where the problem is? E.g. DeployableByOLM-related code?

acornett21:
> 1. Is this a result of a controller-runtime version bump, or have these log lines been there for a while?
> 2. If this is new, should this not be a bug filed against controller-runtime?
> 3. Should this be run closer to where the problem is? E.g. DeployableByOLM-related code?

1. This has been in controller-runtime since `0.15.0`, and I've noticed it off/on since then, but since I'm usually running in `trace` I never understood why it would sometimes happen, and others not.
2. No, this isn't a bug in controller runtime. It's just a warning to remind users of the library that 'hey you haven't called the logger'.
3. I believe this is where this needs to be, since this is part of the [init](https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/log/deleg.go#L100) chain and only have [30seconds](https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/log/log.go#L58) before the warning is thrown, at least that's how I read the code. Most other projects have this is `root` or [main](https://github.com/crossplane/crossplane/pull/4403/files)

komish (approved):
/lgtm

bcrochet (approved):
/lgtm

Issue #1164, opened 2024-05-10
Update Connect URL's to Point to New Locations


acornett21:
## Bug Description
Connect updated their endpoints, so preflight needs to update to matach

## Version and Command Invocation
Latest

## Steps to Reproduce:
(How can we reproduce this?)
1) `preflight check container` with `--submit`
2)
3)

## Expected Result
Expect that the overview URL's take a user to an actual page.

## Actual Result
They 404

## Additional Context
Below are the new URL's from the connect team.

```
https://connect.redhat.com/component/view/${projectId}/images  (This would replace the overview url)
https://connect.redhat.com/component/view/${projectId}/certification/test-results/${testId} (This would replace the scan-results url if you want the users to land to specific test result)
https://connect.redhat.com/component/view/${projectId}/certification (This would be the url if u want the users to simply see all the test results)
```

PR #1166, opened 2024-05-13
updating connect url's to point to new routes

acornett21:
- Fixes: #1164 
- Updating Connect URL's to point to new routes, that we display to the user after running preflight. This does not affect internal API calls to backend services.
- Since Connect layout is different, also adding a new route for Vulnerabilities.
   - This is due to tests/overview/etc not being default landing pages, and pages having nested sidebars.


komish (approved):
/lgtm

bcrochet (approved):
/lgtm

PR #1167, opened 2024-05-15
adding logic to deployablebyolm to validate packagename

acornett21:
- Adding logic to `DeployableByOlm` to validate that packagename complies with DNS-1035 labeling constraints.
   - `App` name was changed for `OperatorGroup`, `CatalogSource` and `Subscription` to address cross referencing.
- Updating test/mock data.

- Relates: redhat-openshift-ecosystem/community-operators-prod#4150

Testing
- Stage ProjectID: 664646cee2c5eef0d5b58e40
- Files with prefixed name uploaded successfully
Screenshot from 2024-05-16 13-51-35

### DIFF
--- a/internal/policy/operator/deployable_by_olm.go
+++ b/internal/policy/operator/deployable_by_olm.go
@@ -28,6 +28,7 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	apiruntime "k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/validation"
 	ctrl "sigs.k8s.io/controller-runtime"
 	crclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
@@ -214,6 +215,8 @@ func checkImageSource(ctx context.Context, operatorImages []string) bool {
 }
 
 func (p *DeployableByOlmCheck) operatorMetadata(ctx context.Context, bundleRef image.ImageReference) (*operatorData, error) {
+	logger := logr.FromContextOrDiscard(ctx)
+
 	// retrieve the operator metadata from bundle image
 	annotationsFileName := filepath.Join(bundleRef.ImageFSPath, "metadata", "annotations.yaml")
 	annotationsFile, err := os.Open(annotationsFileName)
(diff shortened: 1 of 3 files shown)


internal/policy/operator/deployable_by_olm.go
```diff
@@ -248,11 +251,19 @@ func (p *DeployableByOlmCheck) operatorMetadata(ctx context.Context, bundleRef i
 		installModes[val.Type] = val
 	}
 
+	// validating that package name complies with DNS-1035 labeling constraints
+	// ensuring that CatalogSources can be created/referenced in cluster
+	appName := packageName
+	if msgs := validation.IsDNS1035Label(packageName); len(msgs) != 0 {
+		logger.V(log.DBG).Info(fmt.Sprintf("package name %s does not comply with DNS-1035, prefixing to avoid errors", packageName))
+		appName = "preflight-" + packageName
```
komish, line 259: Why are we doing this instead of just failing this check here?

acornett21: We need the operator to still be able to be installed to avoid this error being thrown by k8s further along in the process.

bcrochet: I'd go for a shorter prefix... `pflt-` or even `p-`. You don't want to possibly hit a long one and end up failing because it goes beyond 63.

komish: Either:
a) Make this prefix shorter and validate that it fixes the problem before use (by re-running the isDNS!035Label func again on the new value), or
b) Generate the name used to create the catalog source in a way that's reliably formed.

internal/policy/operator/deployable_by_olm.go
```diff
@@ -248,13 +251,22 @@ func (p *DeployableByOlmCheck) operatorMetadata(ctx context.Context, bundleRef i
 		installModes[val.Type] = val
 	}
 
+	// validating that package name complies with DNS-1035 labeling constraints
+	// ensuring that CatalogSources can be created/referenced in cluster and reconciles properly
+	// if there is an error we need to prefix the name so the install/reconciles can continue
+	appName := packageName
+	if msgs := validation.IsDNS1035Label(packageName); len(msgs) != 0 {
+		logger.V(log.DBG).Info(fmt.Sprintf("package name %s does not comply with DNS-1035, prefixing to avoid errors", packageName))
+		appName = "p-" + packageName
```
komish, line 260: This still needs to be re-run through the DNS label validation. Prefixing the value just fixes one possible invalid scenario.

acornett21: @komish Are you sure? In my testing, and our tests, this fixes both issues that are thrown. Also, if we re-validate, and it fails, then what do we do?

komish: What if the package is called `1app_package`

komish (approved):
/lgtm

internal/policy/operator/deployable_by_olm.go
komish: I suppose this is non-blocking. You're addressing a specific failure that exists, and what I'm asking is a potential (but non-existent) future state. We already have an example in the wild that needs this workaround, but we haven't had an example of what I'm asking you to do yet, and it's unlikely we would given certification would fail if the CatalogSource doesn't come up.

LGTM given

acornett21: I think if this is the case, we'd want this to fail, since that name probably can't be used in other processes. Right now, we really are just trying to solve for the one case for `3scale`.

skattoju (approved):

bcrochet (approved):
/lgtm

Issue #1169, opened 2024-05-23
Update `CatalogSource` Logic to Support OCP 4.16 `PodSecurity` 


acornett21:
## Is your feature request related to a problem? Please describe.
In OCP 4.16 `PodSecurity` is `restricted`, so preflight fails to create a `CatalogSource` successfully.

## Describe the solution you'd like.
preflight should take into consideration the version of OCP under test when creating the `CatalogSource`, if the version is OCP 4.16 or greater the below can be added to the `CatalogSource` struct that is created in the code [here](https://github.com/redhat-openshift-ecosystem/openshift-preflight/blob/main/internal/openshift/openshift.go#L244-L255).

Code Snippet:
```
GrpcPodConfig: &operatorsv1alpha1.GrpcPodConfig{
				SecurityContextConfig: operatorsv1alpha1.Restricted,
},
``` 

## Describe alternatives you've considered.
Optionally we might want to change how we create the namespace to change the labels for pod security, though I'm not sure the ease of that. But this might be better in case an operator author has an incorrect config in their CSV. Alternatively, we might want to catch this in preflight, before it ends up in a real catalog and affects users/customers.

We had a POC of this approach awhile back that is linked [here](https://github.com/komish/openshift-preflight/commit/a9d0b88b9d09c4e580ab2c5859a28124da533b9c)

## Additional context.
Both options seam possible, but I'm leaning towards updating the `CatalogSource`, but it's totally open for discussion.

Error Observed on an OCP 4.16 cluster
```
apiVersion: operators.coreos.com/v1alpha1
kind: CatalogSource
metadata:
  creationTimestamp: "2024-05-23T19:42:51Z"
  generation: 1
  name: simple-demo-operator
  namespace: simple-demo-operator
  resourceVersion: "35590"
  uid: 24f962cf-16bb-4c0d-b019-0e801e5d839d
spec:
  displayName: simple-demo-operator
  icon:
    base64data: ""
    mediatype: ""
  image: quay.io/opdev/simple-demo-operator-catalog:latest
  secrets:
  - registry-auth-keys
  sourceType: grpc
status:
  message: 'couldn''t ensure registry server - error ensuring pod: : error creating
    new pod: simple-demo-operator-: pods "simple-demo-operator-fdr6g" is forbidden:
    violates PodSecurity "restricted:v1.24": allowPrivilegeEscalation != false (container
[...]

PR #1175, opened 2024-06-06
adding 'restricted' to SecurityContextConfig when creating a Catalog

acornett21:
- Fixes: #1169 
- Updating `CatalogSource` creation to include `SecurityContextCOnfig` type of `restricted`.
- Applying this globally with no conditional logic since this is works as expected for the versions of OCP that are currently supported (4.12-4.15), and solves the problem on 4.16.


bcrochet (approved):
/lgtm

komish (approved):
/lgtm

PR #1196, opened 2024-09-12
ensure per-check logger carries the check name throughout execution

komish:
When executing checks, the engine emits check names at the start and end of a check's execution, as well as with the result. We run synchronously, so for the most part, this isn't a problem, but I noticed while troubleshooting something else that logs emitted from the check itself (after pulling the logger from context) don't include the check's name, and it's a small amount of code to pass that context into the check and make the logs a bit more clear.

This PR derives a logger with the check's name as a value, stuffs that into a new context to pass to check functions, and then strips all the other locations that were manually adding the "check" key to the logger.

### DIFF
--- a/internal/engine/engine.go
+++ b/internal/engine/engine.go
@@ -217,9 +217,11 @@ func (c *craneEngine) ExecuteChecks(ctx context.Context) error {
 	// execute checks
 	logger.V(log.DBG).Info("executing checks")
 	for _, executedCheck := range c.checks {
+		logger := logger.WithValues("check", executedCheck.Name())
+		ctx := logr.NewContext(ctx, logger)
 		c.results.TestedImage = c.image
 
-		logger.V(log.DBG).Info("running check", "check", executedCheck.Name())
+		logger.V(log.DBG).Info("running check")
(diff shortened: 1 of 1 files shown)


acornett21 (approved):

internal/engine/engine.go
```diff
@@ -217,20 +217,22 @@ func (c *craneEngine) ExecuteChecks(ctx context.Context) error {
 	// execute checks
 	logger.V(log.DBG).Info("executing checks")
 	for _, executedCheck := range c.checks {
+		chkLogger := logger.WithValues("check", executedCheck.Name())
```
bcrochet, line 220: Since this is inside a loop, I think this can be a two line change. Just shadow ctx and logger. It will not effect the outer values, and will get thrown away at the end of the iteration. I don't see any instances where we use the old values.

bcrochet (changes requested):
See comment

internal/engine/engine.go
komish: Yep, initially thought it was less clear. I'll swap it out.

bcrochet (approved):
/lgtm

PR #1197, opened 2024-09-12
Return a failure instead of an error if no pinned images are found

komish:
If no pinned images were found, we would throw an error for the CertifiedImages check, which isn't technically valid (because we didn't fail to run). This PR converts that into a failure instead. We would get the error in trying to query Pyxis for images with the provided digests. We're returning the error (not Pyxis), but we still want to catch it when returned and make sure it translates to a failure result and not a check execution error.

It also seems like this check had an older style "optional" configuration, which was to return true regardless of things that would cause this check to fail.  I've now made this check's return value accurate to its execution, and allow for the "optional" tag enforce that it's non-blocking.

Finally, I've updated tests which asserted that the things would pass according to the older style of "optional". They would now fail, so the checks needed to be updated accordingly.

This check remains non-blocking for certification.

### DIFF
--- a/internal/policy/operator/certified_images.go
+++ b/internal/policy/operator/certified_images.go
@@ -76,6 +76,11 @@ func (p *certifiedImagesCheck) dataToValidate(ctx context.Context, imagePath str
 func (p *certifiedImagesCheck) validate(ctx context.Context, imageDigests []string) (bool, error) {
 	logger := logr.FromContextOrDiscard(ctx)
 
+	if len(imageDigests) == 0 {
+		logger.Info("warning: pinned images are expected but none were discovered")
+		return false, nil
+	}
+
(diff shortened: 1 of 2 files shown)


acornett21 (approved):

acornett21:
Just linking the logs below to show that this did not change our messaging or results

- Logs: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/redhat-openshift-ecosystem_openshift-preflight/1197/pull-ci-redhat-openshift-ecosystem-openshift-preflight-main-4.17-e2e/1834594772988203008/artifacts/e2e/e2e/build-log.txt

bcrochet (approved):
/lgtm

Issue #1119, opened 2024-01-23
Expose `csvTimeout` and possibly `subscriptionTimeout` as Configurable `ENV's`


acornett21:
## Is your feature request related to a problem? Please describe.
Recently there was an issue where a partners operator took 5-10mins on average to become healthy, but sometimes it took 2mins, and one time it took 2mins it passed the `marketplace` pipeline, but the certification pipeline it would always fail. This means that their operator was released in one catalog and not the other, and the CI pipeline had to be ran on their behalf to resolve the issue.

## Describe the solution you'd like.
It would make sense to expose these timeouts as ENV's so that if we run into this again, we can override this in `preflight-trigger` and possibly the hosted pipeline as well.

## Describe alternatives you've considered.
I think flags would be overkill for this, but that could be another option as well, since flags give us ENV's for free.

## Additional context.
Here is more info in an [ISV-ISSUE](https://issues.redhat.com/browse/ISVISSUE-521)

PR #1198, opened 2024-09-16
adding a flag for csv timeout for check operator cmd

acornett21:
- Adding a flag to set the `csv timeout` during DeployableByOLM check
- Adding new defaults file to the `runtime` package to avoid `import cycle` dependency issue(s) 
- Fixes: #1119

### DIFF
--- a/cmd/preflight/cmd/check_operator.go
+++ b/cmd/preflight/cmd/check_operator.go
@@ -52,6 +52,12 @@ func checkOperatorCmd(runpreflight runPreflight) *cobra.Command {
 		"If empty, the default operator channel in bundle's annotations file is used.. (env: PFLT_CHANNEL)")
 	_ = viper.BindPFlag("channel", checkOperatorCmd.Flags().Lookup("channel"))
 
+	checkOperatorCmd.Flags().Duration("csv-timeout", 0, "The Duration of time to wait for the ClusterServiceVersion to become healthy.\n"+
+		"If empty the default of 180s will be used. (env: PFLT_CSV_TIMEOUT)")
+	_ = viper.BindPFlag("csv_timeout", checkOperatorCmd.Flags().Lookup("csv-timeout"))
+
+	_ = checkOperatorCmd.Flags().MarkHidden("csv-timeout")
+
 	return checkOperatorCmd
 }
 
@@ -167,6 +173,10 @@ func generateOperatorCheckOptions(cfg *runtime.Config) []operator.Option {
 
 	if cfg.Insecure {
 		opts = append(opts, operator.WithInsecureConnection())
+	}
+
+	if cfg.CSVTimeout != 0 {
+		opts = append(opts, operator.WithCSVTimeout(cfg.CSVTimeout))
 	}
 
 	return opts
--- a/internal/engine/engine.go
+++ b/internal/engine/engine.go
@@ -722,6 +722,7 @@ type OperatorCheckConfig struct {
 	ScorecardImage, ScorecardWaitTime, ScorecardNamespace, ScorecardServiceAccount string
 	IndexImage, DockerConfig, Channel                                              string
 	Kubeconfig                                                                     []byte
+	CSVTimeout                                                                     time.Duration
 }
 
 // InitializeOperatorChecks returns opeartor checks for policy p give cfg.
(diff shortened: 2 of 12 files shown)


internal/engine/engine.go
```diff
@@ -729,7 +730,7 @@ func InitializeOperatorChecks(ctx context.Context, p policy.Policy, cfg Operator
 		return []check.Check{
 			operatorpol.NewScorecardBasicSpecCheck(operatorsdk.New(cfg.ScorecardImage, exec.Command), cfg.ScorecardNamespace, cfg.ScorecardServiceAccount, cfg.Kubeconfig, cfg.ScorecardWaitTime),
 			operatorpol.NewScorecardOlmSuiteCheck(operatorsdk.New(cfg.ScorecardImage, exec.Command), cfg.ScorecardNamespace, cfg.ScorecardServiceAccount, cfg.Kubeconfig, cfg.ScorecardWaitTime),
-			operatorpol.NewDeployableByOlmCheck(cfg.IndexImage, cfg.DockerConfig, cfg.Channel),
+			operatorpol.NewDeployableByOlmCheck(cfg.IndexImage, cfg.DockerConfig, cfg.Channel, cfg.CSVTimeout),
```
bcrochet, line 733: This should probably become a functional Option, especially if we foresee any other things needing to be exposed.

bcrochet (commented):
Feel free to ignore, but you added a functional option in one place, and pass it bare in another. I'd keep it a functional option everywhere.

internal/engine/engine.go
acornett21: This is good feedback, add `opts` for this instead.

operator/check_operator.go
```diff
@@ -92,6 +93,7 @@ func (c *operatorCheck) resolve(ctx context.Context) error {
 		DockerConfig:            c.dockerConfigFilePath,
 		Channel:                 c.operatorChannel,
 		Kubeconfig:              c.kubeconfig,
+		CSVTimeout:              c.csvTimeout,
```
komish, line 96: Unless I'm misreading, this value is 0 if the caller didn't include the option. We guard the user from this same problem by setting a default for the scorecard wait time when the check is instantiated, and probably need to do that for this option as well.

komish (changes requested):
Two changes we might be able to make to protect users from setting their timeout to 0 seconds.

cmd/preflight/cmd/check_operator.go
```diff
@@ -52,6 +52,10 @@ func checkOperatorCmd(runpreflight runPreflight) *cobra.Command {
 		"If empty, the default operator channel in bundle's annotations file is used.. (env: PFLT_CHANNEL)")
 	_ = viper.BindPFlag("channel", checkOperatorCmd.Flags().Lookup("channel"))
 
+	checkOperatorCmd.Flags().Int("csv-timeout", 180, "The amount of time to wait (in seconds) for the ClusterServiceVersion to become healthy.\n"+
```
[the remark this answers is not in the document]
bcrochet: A bare int with no context. Agreed. Make it a constant somewhere.

internal/runtime/config_test.go
```diff
@@ -54,6 +54,8 @@ var _ = Describe("Viper to Runtime Config", func() {
 		expectedRuntimeCfg.Channel = "mychannel"
 		baseViperCfg.Set("indeximage", "myindeximage")
 		expectedRuntimeCfg.IndexImage = "myindeximage"
+		baseViperCfg.Set("csv_timeout", 180)
+		expectedRuntimeCfg.CSVTimeout = 180
```
bcrochet, line 58: These should all reference the aforementioned constant.

acornett21: Added new defaults file for this.

cmd/preflight/cmd/check_operator.go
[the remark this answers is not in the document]
acornett21: Changed this to use `Duration` since I didn't know this was an option for `Flags` and moved to a constant.

operator/check_operator.go
acornett21: Default file was created for this and setting default in root cmd like other defaults.

cmd/preflight/cmd/check_operator.go
```diff
@@ -52,6 +52,12 @@ func checkOperatorCmd(runpreflight runPreflight) *cobra.Command {
 		"If empty, the default operator channel in bundle's annotations file is used.. (env: PFLT_CHANNEL)")
 	_ = viper.BindPFlag("channel", checkOperatorCmd.Flags().Lookup("channel"))
 
+	checkOperatorCmd.Flags().Duration("csv-timeout", 0, "The Duration of time to wait for the ClusterServiceVersion to become healthy.\n"+
```
skattoju, line 55: nit: " .. time in seconds .." but its implied by the 180s default value mentioned 👍

skattoju (approved):

cmd/preflight/cmd/check_operator.go
acornett21: I took out the mention of the unit here, since I switched to Duration, this could be any unit, seconds, minutes, hours, etc.

skattoju: Ah so the unit suffix is needed if not ‘—csv-timeout=300’ would mean nanoseconds

acornett21: No the suffix is required, otherwise cobra throws an error. `—csv-timeout=300` is not valid.

bcrochet (approved):
/lgtm

PR #1200, opened 2024-09-20
Refactor commands to take a viper instance

bcrochet:
Instead of relying on a global instance, the commands should instead have viper injected. This will make testing easier.


bcrochet:
Replaces #958

cmd/preflight/cmd/check_operator.go
```diff
@@ -11,28 +11,29 @@ import (
...
-func checkOperatorCmd(runpreflight runPreflight) *cobra.Command {
+func checkOperatorCmd(runpreflight runPreflight, viper *viper.Viper) *cobra.Command {
 	checkOperatorCmd := &cobra.Command{
 		Use:   "operator",
 		Short: "Run checks for an Operator",
 		Long:  `This command will run the Certification checks for an Operator bundle image. `,
-		Args:  checkOperatorPositionalArgs,
+		Args: func(cmd *cobra.Command, args []string) error {
+			return checkOperatorPositionalArgs(args, viper)
+		},
```
komish, line 29: Should we extract this closure into its own function? As in, to reduce in-line function definitions where not needed?

cmd/preflight/cmd/check_container.go
```diff
@@ -238,24 +239,34 @@ func checkContainerPositionalArgs(cmd *cobra.Command, args []string) error {
...
+	validatedCertificationProjectID, err := validateCertificationProjectID(certificationProjectID)
+	if err != nil {
+		return err
+	}
+	if validatedCertificationProjectID != certificationProjectID {
+		viper.Set("certification_project_id", validatedCertificationProjectID)
+	}
+	return nil
+}
+
```
komish, line 253: Extraneous?

komish (changes requested):
Looks fine overall. Mostly nits, except the `ResultSubmitter{}.Submit` interface definition change.

internal/lib/types.go
```diff
@@ -29,7 +28,7 @@ type ResultWriter interface {
 
 // ResultSubmitter defines methods associated with submitting results to Red HAt.
 type ResultSubmitter interface {
-	Submit(context.Context) error
+	Submit(context.Context, string) error
```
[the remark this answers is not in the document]
bcrochet: It could be a type alias to make it clearer. TBH, we shouldn't even be exporting interfaces in the first place.

cmd/preflight/cmd/check_operator.go
bcrochet: Either way. Since it's a single line I'm not super worried about this one.

cmd/preflight/cmd/check_container.go
bcrochet: Might be vestigial from an earlier revision, given the age of this patch.

I'll check it out.

internal/lib/types.go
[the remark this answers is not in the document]
komish: A type alias isn't strictly necessary. Something like this should be plenty:

```suggestion
	Submit(ctx context.Context, pyxisEnv string) error
```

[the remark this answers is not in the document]
acornett21: I was confused by this as well, it took me till the end of the PR to understand that this was `pyxisEnv`.

cmd/preflight/cmd/check_container_test.go
```diff
@@ -145,24 +152,24 @@ var _ = Describe("Check Container Command", func() {
 		Entry("index manifest, invalid platform", "index", "none", HaveOccurred(), true),
 	)
 
-	Context("When validating check container arguments and flags", func() {
+	When("validating check container arguments and flags", func() {
```
acornett21, line 155: Non-blocking: Why did you switch from `context` to `when`, is there some benefit? I'm fine with the change, but just want to understand the driving factor.

bcrochet: It does the same thing, but "reads" better. In both the code and the output when a test fails.

bcrochet:
This seems to have outlived its usefulness. Closing.

PR #1208, opened 2024-10-15
refactor operator newcheck constructor defaults

acornett21:
- Moving scorecard default timeout to runtime like csv and subscription timeout
- Update WithCSVTimeout method description for consistency


komish (approved):
/lgtm

Issue #1203, opened 2024-10-02
preflight fails on "operator-sdk scorecard: unexpected end of JSON input"


jbattiato:
## Bug Description

The `preflight check operator` command fails to read the `scorecard` output with the following error:

```
time="2024-10-02T15:57:04+02:00" level=info msg="running scorecard with the following invocation" args="[\"operator-sdk\",\"scorecard\",\"--output\",\"json\",\"--selector=test=basic-check-spec-test\",\"--kubeconfig\",\"/tmp/2804899030\",\"--wait-time\",\"240s\",\"--namespace\",\"default\",\"--service-account\",\"default\",\"--config\",\"/tmp/scorecard-test-config-798981960.yaml\",\"--verbose\",\"/tmp/preflight-3318550770/fs\"]"
time="2024-10-02T15:57:04+02:00" level=info msg="check completed" check=ScorecardBasicSpecCheck err="failed to run operator-sdk scorecard: unexpected end of JSON input" result=ERROR
```

## Version and Command Invocation

1.10.0

```
ime="2024-10-02T12:24:50Z" level=info msg="certification library version" version="1.10.0 <commit: c9048da99aae76ddee5a708edcc94e14c034cd1d>"
```

## Steps to Reproduce:
```
PFLT_INDEXIMAGE=<index-image> PFLT_ARTIFACTS=preflight_results  KUBECONFIG=${HOME}/.kube/config \
  ./preflight check operator <bundle-image>  --docker-config ${HOME}/.docker/config.json
```

I'm sorry, I'm working on a closed fork of [cloudnative-pg](https://github.com/cloudnative-pg/cloudnative-pg), and I can't share the associated workflow output.

## Expected Result
preflight should be able to read the scorecard JSON output since the latter is well formed when running it standalone:

```
./operator-sdk scorecard <bundle-image> --wait-time 60s --output json > scorecard.json
jq -e . >/dev/null 2>&1 <<<$(cat scorecard.json)
echo $?
0
```

## Actual Result

These errors from `ScorecardBasicSpecCheck` and `ScorecardOlmSuiteCheck`:

```
level=info msg="check completed" check=ScorecardBasicSpecCheck err="failed to run operator-sdk scorecard: unexpected end of JSON input" result=ERROR
```
```
level=info msg="check completed" check=ScorecardOlmSuiteCheck err="failed to run operator-sdk scorecard: unexpected end of JSON input" result=ERROR
```

## Additional Context

[...]

PR #1209, opened 2024-10-16
add exec.LookPath to check operator-sdk is in $PATH

acornett21:
## Motivation 
The ux when a user does not have `operator-sdk` in their path leads to a confusing error that is hard to track down.

Current Error Message:
```
time="2024-10-02T15:57:04+02:00" level=info msg="check completed" check=ScorecardBasicSpecCheck err="failed to run operator-sdk scorecard: unexpected end of JSON input" result=ERROR
```

This error is not clear at all what is going on, but we do not have direct control of this once we shell out to `operator-sdk`

## Changes
Add an explicit check to make sure that `operator-sdk` binary is in the users `PATH`. This provides an error message that is more clearer.

New Error Message:
```
time="2024-10-16T13:05:53-07:00" level=info msg="check completed" check=ScorecardBasicSpecCheck err="exec: \"operator-sdk\": executable file not found in $PATH" result=ERROR
```

## Note
Since this is in the `Scorecard` method, it does not matter which check `Basic` or `OlmSuite` gets called, this error will appear.

- Relates: #1203

### DIFF
--- a/internal/operatorsdk/operatorsdk.go
+++ b/internal/operatorsdk/operatorsdk.go
@@ -34,6 +34,12 @@ type execContext = func(name string, arg ...string) *exec.Cmd
 func (o operatorSdk) Scorecard(ctx context.Context, image string, opts OperatorSdkScorecardOptions) (*OperatorSdkScorecardReport, error) {
 	logger := logr.FromContextOrDiscard(ctx)
 
+	// checking to make sure operator-sdk is in the $PATH
+	_, err := exec.LookPath("operator-sdk")
+	if err != nil {
+		return nil, err
+	}
+
 	cmdArgs := []string{"scorecard"}
 	if opts.OutputFormat == "" {
 		opts.OutputFormat = "json"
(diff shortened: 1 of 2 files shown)


internal/operatorsdk/operatorsdk.go
```diff
@@ -34,6 +34,12 @@ type execContext = func(name string, arg ...string) *exec.Cmd
 func (o operatorSdk) Scorecard(ctx context.Context, image string, opts OperatorSdkScorecardOptions) (*OperatorSdkScorecardReport, error) {
 	logger := logr.FromContextOrDiscard(ctx)
 
+	// checking to make sure operator-sdk is in the $PATH
+	_, err := exec.LookPath("operator-sdk")
+	if err != nil {
+		return nil, fmt.Errorf("%v", err)
```
bcrochet, line 40: Just return err. This adds no context.

bcrochet (changes requested):
Small change

internal/operatorsdk/operatorsdk.go
acornett21: done

bcrochet (approved):
/lgtm

PR #1212, opened 2024-11-08
adding PyxisHost value to ContainerCheckConfig for ubi check

acornett21:
## Motivation
To be able to lookup pre-prod ubi images in pyxis for testing.

## Fix
Add `PyxisHost` as to the `ContainerCheckConfig`, and use the value of  of `c.pyxisHost` that we have from the `checkContainer` struct.


bcrochet (approved):
/lgtm

PR #1215, opened 2024-11-20
Introduce ability to get profile data

bcrochet:
This exposes two environment variables, PFLT_MEMPROFILE and PFLT_CPUPROFILE, that should be set to the path and name of a file to write the profile data to. These can then be processed with `go tool pprof` for further analysis.

### DIFF
--- a/cmd/preflight/cmd/check_container.go
+++ b/cmd/preflight/cmd/check_container.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"path/filepath"
 	rt "runtime"
+	"runtime/pprof"
 	"slices"
 	"strings"
 
@@ -89,6 +90,9 @@ func checkContainerCmd(runpreflight runPreflight) *cobra.Command {
 	flags.String("platform", rt.GOARCH, "Architecture of image to pull. Defaults to runtime platform.")
 	_ = viper.BindPFlag("platform", flags.Lookup("platform"))
 
+	_ = viper.BindEnv("cpuprofile")
+	_ = viper.BindEnv("memprofile")
+
 	return checkContainerCmd
 }
 
@@ -100,6 +104,21 @@ func checkContainerRunE(cmd *cobra.Command, args []string, runpreflight runPrefl
 		return fmt.Errorf("invalid logging configuration")
 	}
 	logger.Info("certification library version", "version", version.Version.String())
+
+	if viper.Instance().IsSet("cpuprofile") {
+		f, err := os.Create(viper.Instance().GetString("cpuprofile"))
+		if err != nil {
+			logger.Error(err, "could not create CPU profile")
+			return err
+		}
+		defer f.Close()
+
+		if err := pprof.StartCPUProfile(f); err != nil {
+			logger.Error(err, "could not start CPU profile")
+			return err
+		}
+		defer pprof.StopCPUProfile()
+	}
 
 	containerImage := args[0]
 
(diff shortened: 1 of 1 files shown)


skattoju (approved):

cmd/preflight/cmd/check_container.go
```diff
@@ -89,6 +90,9 @@ func checkContainerCmd(runpreflight runPreflight) *cobra.Command {
 	flags.String("platform", rt.GOARCH, "Architecture of image to pull. Defaults to runtime platform.")
 	_ = viper.BindPFlag("platform", flags.Lookup("platform"))
 
+	_ = viper.BindEnv("cpuprofile")
```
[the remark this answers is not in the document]
bcrochet: That is what it does... The path is from the env var.

cmd/preflight/cmd/check_container.go
```diff
@@ -101,6 +105,20 @@ func checkContainerRunE(cmd *cobra.Command, args []string, runpreflight runPrefl
...
+	if viper.Instance().IsSet("cpuprofile") {
+		f, err := os.Create(viper.Instance().GetString("cpuprofile"))
+		if err != nil {
+			logger.Error(err, "could not create CPU profile")
+			return err
+		}
+		defer f.Close()
+
+		if err := pprof.StartCPUProfile(f); err != nil {
+			logger.Error(err, "could not start CPU profile")
```
komish, line 117: should this error be returned?

cmd/preflight/cmd/check_container.go
```diff
@@ -194,6 +212,18 @@ func checkContainerRunE(cmd *cobra.Command, args []string, runpreflight runPrefl
 		}
 	}
 
+	if viper.Instance().IsSet("memprofile") {
+		f, err := os.Create(viper.Instance().GetString("memprofile"))
+		if err != nil {
+			logger.Error(err, "could not create memory profile")
```
komish, line 218: Should this error be returned?

cmd/preflight/cmd/check_container.go
```diff
@@ -194,6 +212,18 @@ func checkContainerRunE(cmd *cobra.Command, args []string, runpreflight runPrefl
...
+		}
+		defer f.Close()
+
+		rt.GC() // get up-to-date statistics
+		if err := pprof.Lookup("allocs").WriteTo(f, 0); err != nil {
+			logger.Error(err, "could not start memory profile")
```
komish, line 224: Should this error be returned?

cmd/preflight/cmd/check_container.go
komish, line 93: No flags? Just environment vars?

komish (commented):
Any particular reason why this is only being added to CheckContainer?

cmd/preflight/cmd/check_container.go
bcrochet: Sure. I'll add that.

cmd/preflight/cmd/check_container.go
bcrochet: Ditto.

cmd/preflight/cmd/check_container.go
bcrochet: Correct. It's not really something the end user is going to use. This is a dev tool IMHO.

bcrochet:
> Any particular reason why this is only being added to CheckContainer?

It was where I knew we were looking for a data from. If we decide we need it elsewhere in the future, we can cross that bridge then.

cmd/preflight/cmd/check_container.go
acornett21: I don't feel strongly one way or the other on this, but if others disagree and we need a flag, we could always go the hidden flag route like we have with some other flags.

komish (approved):
/lgtm

PR #1216, opened 2024-11-22
Remove extraneous code from CraneEngine

bcrochet:
Turns out that we were making an extra call to io.Copy that is completely unnecessary. We can use the io.Reader directly from crane's mutate.Extract call.

### DIFF
--- a/internal/engine/engine.go
+++ b/internal/engine/engine.go
@@ -38,7 +38,6 @@ import (
 
 	"github.com/google/go-containerregistry/pkg/crane"
 	"github.com/google/go-containerregistry/pkg/name"
-	cranev1 "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/google/go-containerregistry/pkg/v1/cache"
 	"github.com/google/go-containerregistry/pkg/v1/mutate"
 )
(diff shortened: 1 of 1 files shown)


internal/engine/engine.go
```diff
@@ -154,24 +147,16 @@ func (c *craneEngine) ExecuteChecks(ctx context.Context) error {
...
-		logger.V(log.DBG).Info("writing container filesystem", "outputDirectory", containerFSPath)
-
-		// Close the writer with any errors encountered during
-		// extraction. These errors will be returned by the reader end
-		// on subsequent reads. If err == nil, the reader will return
-		// EOF.
-		w.CloseWithError(export(img, w))
-	}()
+	fs := mutate.Extract(img)
+	defer fs.Close()
```
komish, line 151: If I'm reading this correctly, this close doesn't complete until the end of preflight's entire run. Should we close this explicitly sooner in success cases, or otherwise put in a closure to close it  once the untar has completed?

komish:
I'm a bit confused why we're losing code but also losing coverage here.

internal/engine/engine.go
bcrochet: You are correct. Would be good to put in the closure to force the closure. :)

komish (approved):
/lgtm

acornett21 (approved):

PR #1219, opened 2024-12-02
adding method to see if a given string violates Red Hat Trademark(s)

acornett21:
## Motivation
We need to start enforcing our trademark rights for `Red Hat` verbiage in various checks within the `container` policy. This PR is the starting point for that work.

## Changes
- Added a global method that contains the same validation as `pyxis` (this might become un-exported once full requirements are presented by the business) 
- Updated `HasRequiredLabels` check to check for trademark violations of various labels.
- Added `maintainer` to required label field in `HasRequiredLabels` check
- Added new `HasProhibitedContainerName` check

### DIFF
--- a/internal/policy/container/has_prohibited_container_name.go
+++ b/internal/policy/container/has_prohibited_container_name.go
@@ -1,0 +1,57 @@
+package container
+
+import (
+	"context"
+	"strings"
+
+	"github.com/go-logr/logr"
+
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/internal/check"
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/internal/image"
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/internal/log"
+)
+
+var _ check.Check = &HasProhibitedContainerName{}
+
+type HasProhibitedContainerName struct{}
+
+func (p HasProhibitedContainerName) Validate(ctx context.Context, imageReference image.ImageReference) (result bool, err error) {
+	return p.validate(ctx, p.getDataForValidate(imageReference.ImageRepository))
+}
+
+func (p HasProhibitedContainerName) getDataForValidate(imageRepository string) string {
+	// splitting on '/' to get container name, at this point we know that
+	// crane's ParseReference has set ImageReference.imageRepository in a valid format
+	return strings.Split(imageRepository, "/")[1]
+}
+
+func (p HasProhibitedContainerName) validate(ctx context.Context, containerName string) (bool, error) {
+	logger := logr.FromContextOrDiscard(ctx)
+
+	if violatesRedHatTrademark(containerName) {
+		logger.V(log.DBG).Info("container name violate Red Hat trademark", "container-name", containerName)
+		return false, nil
+	}
+
+	return true, nil
+}
+
+func (p HasProhibitedContainerName) Name() string {
+	return "HasProhibitedContainerName"
+}
+
+func (p HasProhibitedContainerName) Metadata() check.Metadata {
+	return check.Metadata{
+		Description:      "Checking if the container-name violates Red Hat trademark.",
+		Level:            "good",
+		KnowledgeBaseURL: certDocumentationURL,
+		CheckURL:         certDocumentationURL,
+	}
+}
+
+func (p HasProhibitedContainerName) Help() check.HelpText {
+	return check.HelpText{
+		Message:    "Check HasProhibitedContainerName encountered an error. Please review the preflight.log file for more information.",
+		Suggestion: "Update container-name ie (quay.io/repo-name/container-name) to not violate Red Hat trademark.",
+	}
+}
--- a/internal/policy/container/has_prohibited_container_name_test.go
+++ b/internal/policy/container/has_prohibited_container_name_test.go
@@ -1,0 +1,42 @@
+package container
+
+import (
+	"context"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/internal/image"
+)
+
+var _ = Describe("HasProhibitedContainerName", func() {
+	var (
+		hasProhibitedContainerName HasProhibitedContainerName
+		imageRef                   image.ImageReference
+	)
+
+	Describe("Checking for trademark violations", func() {
+		Context("When a container name does not violate trademark", func() {
+			BeforeEach(func() {
+				imageRef.ImageRepository = "opdev/simple-demo-operator"
+			})
+			It("should pass Validate", func() {
+				ok, err := hasProhibitedContainerName.Validate(context.TODO(), imageRef)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(ok).To(BeTrue())
+			})
+		})
+		Context("When a container name violates trademark", func() {
+			BeforeEach(func() {
+				imageRef.ImageRepository = "opdev/red-hat-container"
+			})
+			It("should not pass Validate", func() {
+				ok, err := hasProhibitedContainerName.Validate(context.TODO(), imageRef)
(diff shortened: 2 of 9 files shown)


internal/policy/container/trademark_validator.go
```diff
@@ -0,0 +1,20 @@
...
+	// string starts with Red Hat variant
+	startingWithRedHat := regexp.MustCompile("^[^a-z0-9]*red[^a-z0-9]*hat").MatchString(strings.ToLower(nameLabel))
+
+	// string contain Red Hat variant (not starting with)
+	containsRedHat := regexp.MustCompile("red[^a-z0-9]*hat").MatchString(strings.ToLower(nameLabel))
+
+	// string contains "for Red Hat" variant
+	containsForRedHat := regexp.MustCompile("for[^a-z0-9]*red[^a-z0-9]*hat").MatchString(strings.ToLower(nameLabel))
+
+	return startingWithRedHat && containsRedHat && !containsForRedHat
```
komish, line 19: Should `containsRedHat && !containsForRedHat)` be grouped? I'm understanding `containsForRedHat` to mean that you *may* use "for Red Hat" in your naming, but if you don't, doesn't MatchString return "false", making the return condition return "true" for containsRedHat?

internal/policy/container/trademark_validator.go
```diff
@@ -0,0 +1,20 @@
...
+	"strings"
+)
+
+// ViolatesRedHatTrademark validates if a string meets specific "Red Hat" naming criteria
+func ViolatesRedHatTrademark(nameLabel string) bool {
+	// string starts with Red Hat variant
+	startingWithRedHat := regexp.MustCompile("^[^a-z0-9]*red[^a-z0-9]*hat").MatchString(strings.ToLower(nameLabel))
+
+	// string contain Red Hat variant (not starting with)
+	containsRedHat := regexp.MustCompile("red[^a-z0-9]*hat").MatchString(strings.ToLower(nameLabel))
```
komish, line 14: Do we care about `redhat` (all one word)? I realize it's probably not a specific trademark violation, but can't hurt to ask.

internal/policy/container/trademark_validator.go
acornett21: I think you mean (containsRedHat && !containsForRedHat) correct? This is what I initially had but I wasn't sure.

internal/policy/container/trademark_validator.go
```diff
@@ -0,0 +1,20 @@
+package container
+
+import (
+	"regexp"
+	"strings"
+)
+
+// ViolatesRedHatTrademark validates if a string meets specific "Red Hat" naming criteria
+func ViolatesRedHatTrademark(nameLabel string) bool {
```
komish, line 9: NIT: the input is an arbitrary string. nameLabel is a reference to where it's called, but there's technically reason to carry this over as the variable name.

```suggestion
func ViolatesRedHatTrademark(s string) bool {
```

internal/policy/container/has_required_labels_test.go
```diff
@@ -11,18 +11,22 @@ import (
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/internal/image"
 )
 
-func getLabels(bad bool) map[string]string {
+func getLabels(override string) map[string]string {
 	labels := map[string]string{
-		"name":        "name",
+		"name":        "Dynatrace for Red Hat OpenShift",
```
komish, line 16: Can probably remove explicit references to company names.

internal/policy/container/has_prohibited_container_name_test.go
```diff
@@ -0,0 +1,42 @@
...
+)
+
+var _ = Describe("HasProhibitedContainerName", func() {
+	var (
+		hasProhibitedContainerName HasProhibitedContainerName
+		imageRef                   image.ImageReference
+	)
+
+	Describe("Checking for trademark violations", func() {
+		Context("When container-name does not violate trademark", func() {
```
komish, line 19:
```suggestion
		Context("When a container name does not violate trademark", func() {
```

internal/policy/container/has_prohibited_container_name.go
```diff
@@ -0,0 +1,57 @@
...
+	"github.com/go-logr/logr"
+
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/internal/check"
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/internal/image"
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/internal/log"
+)
+
+var _ check.Check = &HasProhibitedContainerName{}
+
+type HasProhibitedContainerName struct{}
```
komish, line 16: Should this have been something like `HasNoTrademarkViolations`?

internal/policy/container/has_prohibited_container_name.go
```diff
@@ -0,0 +1,57 @@
...
+	// splitting on '/' to get container name, at this point we know that
+	// crane's ParseReference has set ImageReference.imageRepository in a valid format
+	return strings.Split(imageRepository, "/")[1]
+}
+
+func (p HasProhibitedContainerName) validate(ctx context.Context, containerName string) (bool, error) {
+	logger := logr.FromContextOrDiscard(ctx)
+
+	if ViolatesRedHatTrademark(containerName) {
+		logger.V(log.DBG).Info("container-name violate Red Hat trademark", "container-name", containerName)
```
komish, line 32:
```suggestion
		logger.V(log.DBG).Info("container name violate Red Hat trademark", "container-name", containerName)
```

internal/policy/container/trademark_validator.go
acornett21: I think we do, and a test would fail with this string.

internal/policy/container/has_prohibited_container_name.go
acornett21: This only covers the container name, so a specific test for that is needed. Maybe if we have more tests that they want lumped together, we can combine.

internal/policy/container/trademark_validator.go
acornett21: I added a test for `redhat`, let me know what you think.

internal/policy/container/trademark_validator.go
komish: Yeah, I mean - I also think that might be incomplete.

I'd guess it's possible for a name to be something like "ProductName By Red Hat for Red Hat". Technically, still a violation, and I think the existing logic would not catch it, even with the grouping.

I'm guessing you want a count of matches, and a count of "for Red Hats". Subtract the two, that's the number of violations for the last two cases.

acornett21: I updated this.

komish:
@ramperher Is this still passing?

komish (approved):
/lgtm

acornett21:
> @ramperher Is this still passing?

The sha seems to have reverted back to the old one.

internal/policy/container/has_required_labels.go
```diff
@@ -67,7 +81,8 @@ func (p *HasRequiredLabelsCheck) Metadata() check.Metadata {
 
 func (p *HasRequiredLabelsCheck) Help() check.HelpText {
 	return check.HelpText{
-		Message:    "Check Check HasRequiredLabel encountered an error. Please review the preflight.log file for more information.",
-		Suggestion: "Add the following labels to your Dockerfile or Containerfile: name, vendor, version, release, summary, description",
+		Message: "Check HasRequiredLabel encountered an error. Please review the preflight.log file for more information.",
+		Suggestion: "Add the following labels to your Dockerfile or Containerfile: name, vendor, version, release, summary, description" +
```
skattoju, line 85: `maintainer` is also a required label now if i am reading correctly ?

acornett21: Good catch this also needs to be updated.

skattoju (approved):

PR #1222, opened 2024-12-09
Add certification-component-id flag

### LINKED ISSUE #1185: Add `--certification-component-id` as synonym to `--certification-project-id`
As related in Jira [PEC-648](https://issues.redhat.com/browse/PEC-648), the overall Partner Connect certification process and documentation now calls projects by the new term **components**.

Preflight still refers to these entities as **projects** rather than **components**.

Preflight should add a command-line option `--certification-component-id` in addition to the existing `--certification-project-id`.
In addition, Preflight should add an environment variable, `PFLT_CERTIFICATION_COMPONENT_ID` for the same purpose.

To keep existing automation working properly, the existing `--certification-project-id` option and `PFLT_CERTIFICATION_PROJECT_ID` environment variable should continue to be accepted if specified. However, only `--certification-component-id` should be shown in the help flags list. If that is not possible, then change the description to indicate that `--certification-project-id` is also accepted.

bcrochet:
This flag deprecates the certification-project-id flag. The field has been changed on the server side, and in certification documentation.

Fixes #1185

### DIFF
--- a/cmd/preflight/cmd/check_container.go
+++ b/cmd/preflight/cmd/check_container.go
@@ -33,7 +33,10 @@ import (
 	"github.com/spf13/pflag"
 )
 
-var submit bool
+var (
+	submit      bool
+	componentID string
+)
 
 // runPreflight is introduced to make testing of this command possible, it has the same method signature as cli.RunPreflight.
 type runPreflight func(context.Context, func(ctx context.Context) (certification.Results, error), cli.CheckConfig, formatters.ResponseFormatter, lib.ResultWriter, lib.ResultSubmitter) error
@@ -82,9 +85,21 @@ func checkContainerCmd(runpreflight runPreflight) *cobra.Command {
 	flags.String("pyxis-env", check.DefaultPyxisEnv, "Env to use for Pyxis submissions.")
 	_ = viper.BindPFlag("pyxis_env", flags.Lookup("pyxis-env"))
 
-	flags.String("certification-project-id", "", fmt.Sprintf("Certification Project ID from connect.redhat.com/projects/{certification-project-id}/overview\n"+
+	flags.String("certification-project-id", "", fmt.Sprintf("Certification project ID from connect.redhat.com/projects/{certification-project-id}/overview\n"+
 		"URL paramater. This value may differ from the PID on the overview page. (env: PFLT_CERTIFICATION_PROJECT_ID)"))
 	_ = viper.BindPFlag("certification_project_id", flags.Lookup("certification-project-id"))
+	_ = flags.MarkDeprecated("certification-project-id", "please use --certification-component-id instead")
+
+	// Use a bound package-level var here. We are going to leave the rest of the code
+	// using the Viper id of 'certification_project_id' in order to minimize changes
+	// in the overall code base.
+	// When --certification-project-id is fully removed, this should become bound in Viper
+	flags.StringVar(&componentID, "certification-component-id", "", fmt.Sprintf("Certification component ID from connect.redhat.com/component/view/{certification-component-id}/images\n"+
+		"URL paramater. This value may differ from the component PID on the overview page. (env: PFLT_CERTIFICATION_COMPONENT_ID)"))
+	// Here, we are forcing an env binding, so we can check that later. This should also
+	// be moved to "automatic" once the old project id is removed
+	_ = viper.BindEnv("certification_component_id")
+	checkContainerCmd.MarkFlagsMutuallyExclusive("certification-project-id", "certification-component-id")
 
 	flags.String("platform", rt.GOARCH, "Architecture of image to pull. Defaults to runtime platform.")
 	_ = viper.BindPFlag("platform", flags.Lookup("platform"))
@@ -210,12 +225,25 @@ func checkContainerPositionalArgs(cmd *cobra.Command, args []string) error {
 		}
 	})
 
-	// --submit was specified
 	viper := viper.Instance()
+
+	// If the new flag is set, use that
+	if cmd.Flag("certification-component-id").Changed {
+		cmd.Flag("certification-project-id").Changed = true
+		cmd.Flag("certification-component-id").Changed = false
+		viper.Set("certification_project_id", componentID)
+	}
+
+	// However, if the new env var is set, that's the priority
+	if viper.IsSet("certification_component_id") {
+		viper.Set("certification_project_id", viper.GetString("certification_component_id"))
+	}
+
+	// --submit was specified
 	if submit {
 		// If the flag is not marked as changed AND viper hasn't gotten it from environment, it's an error
 		if !cmd.Flag("certification-project-id").Changed && !viper.IsSet("certification_project_id") {
-			return fmt.Errorf("certification Project ID must be specified when --submit is present")
+			return fmt.Errorf("certification component ID must be specified when --submit is present")
 		}
 		if !cmd.Flag("pyxis-api-token").Changed && !viper.IsSet("pyxis_api_token") {
 			return fmt.Errorf("pyxis API Token must be specified when --submit is present")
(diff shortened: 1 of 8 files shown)


cmd/preflight/cmd/check_container.go
```diff
@@ -82,9 +85,21 @@ func checkContainerCmd(runpreflight runPreflight) *cobra.Command {
...
 		"URL paramater. This value may differ from the PID on the overview page. (env: PFLT_CERTIFICATION_PROJECT_ID)"))
 	_ = viper.BindPFlag("certification_project_id", flags.Lookup("certification-project-id"))
+	_ = flags.MarkDeprecated("certification-project-id", "please use --certification-component-id instead")
+
+	// Use a bound package-level var here. We are going to leave the rest of the code
+	// using the Viper id of 'certification_project_id' in order to minimize changes
+	// in the overall code base.
+	// When --certification-project-id is fully removed, this should become bound in Viper
+	flags.StringVar(&componentID, "certification-component-id", "", fmt.Sprintf("Certification component ID from connect.redhat.com/projects/{certification-component-id}/overview\n"+
+		"URL paramater. This value may differ from the CID on the overview page. (env: PFLT_CERTIFICATION_COMPONENT_ID)"))
```
acornett21, line 98: This should still said PID, or `Component PID`.

acornett21: No PID is PID as a word.

bcrochet: Updated.

cmd/preflight/cmd/check_container.go
```diff
@@ -82,9 +85,21 @@ func checkContainerCmd(runpreflight runPreflight) *cobra.Command {
...
+	flags.String("certification-project-id", "", fmt.Sprintf("Certification project ID from connect.redhat.com/projects/{certification-project-id}/overview\n"+
 		"URL paramater. This value may differ from the PID on the overview page. (env: PFLT_CERTIFICATION_PROJECT_ID)"))
 	_ = viper.BindPFlag("certification_project_id", flags.Lookup("certification-project-id"))
+	_ = flags.MarkDeprecated("certification-project-id", "please use --certification-component-id instead")
+
+	// Use a bound package-level var here. We are going to leave the rest of the code
+	// using the Viper id of 'certification_project_id' in order to minimize changes
+	// in the overall code base.
+	// When --certification-project-id is fully removed, this should become bound in Viper
+	flags.StringVar(&componentID, "certification-component-id", "", fmt.Sprintf("Certification component ID from connect.redhat.com/projects/{certification-component-id}/overview\n"+
```
acornett21, line 97:
```suggestion
	flags.StringVar(&componentID, "certification-component-id", "", fmt.Sprintf("Certification component ID from https://connect.redhat.com/component/view/{certification-component-id}\n"+
```

Looks like this route moved, and we can use the suggestion.

acornett21 (approved):

cmd/preflight/cmd/check_container.go
```diff
@@ -82,9 +85,21 @@ func checkContainerCmd(runpreflight runPreflight) *cobra.Command {
...
+	// Use a bound package-level var here. We are going to leave the rest of the code
+	// using the Viper id of 'certification_project_id' in order to minimize changes
+	// in the overall code base.
+	// When --certification-project-id is fully removed, this should become bound in Viper
+	flags.StringVar(&componentID, "certification-component-id", "", fmt.Sprintf("Certification component ID from connect.redhat.com/component/view/{certification-component-id}/images\n"+
+		"URL paramater. This value may differ from the component PID on the overview page. (env: PFLT_CERTIFICATION_COMPONENT_ID)"))
+	// Here, we are forcing an env binding, so we can check that later. This should also
+	// be moved to "automatic" once the old project id is removed
+	_ = viper.BindEnv("certification_component_id")
+	checkContainerCmd.MarkFlagsMutuallyExclusive("certification-project-id", "certification-component-id")
```
komish, line 102: I'm confused. Why not just bind this to viper now? It'll be bound at a different endpoint anyway, and you can  resolve the two in the places you need, such as in runtime.NewConfigFrom?

komish, line 102: I'm confused. Why not just bind this to viper now? It'll be bound at a different endpoint anyway, and you can resolve the two in the places you need, such as in runtime.NewConfigFrom?

container/check_container.go
```diff
@@ -125,11 +125,24 @@ func WithDockerConfigJSONFromFile(s string) Option {
 	}
 }
 
+// DEPRECATED: Use WithCertificationProject instead
 // WithCertificationProject adds the project's id and pyxis token to the check
 // allowing for the project's metadata to change the certification (if necessary).
 // An example might be the Scratch or Privileged flags on a project allowing for
 // the corresponding policy to be executed.
 func WithCertificationProject(id, token string) Option {
```
komish, line 133:
```suggestion
// WithCertificationProject adds the project's id and pyxis token to the check
// allowing for the project's metadata to change the certification (if necessary).
// An example might be the Scratch or Privileged flags on a project allowing for
// the corresponding policy to be executed.
//
// Deprecated: Use WithCertificationComponent instead
func WithCertificationProject(id, token string) Option {
```

Writing **Deprecated** (over DEPRECATED) seems to correctly trigger staticcheck when linting. Also, placing it at the end (or, potentially just _separated by one blank comment link_) makes staticcheck report the suggestion correctly. I wouldn't think it matters, but [Godoc](https://go.dev/blog/godoc) specifies the Title Case here.

Also note that the statement is saying to use the same function it's deprecating. Should read "use WithCertification**Component** instead".

Finally, I see two places where this is called in the project that need updating.

```
cmd/preflight/cmd/check_container.go:294:3   staticcheck  SA1019: container.WithCertificationProject is deprecated: Use WithCertificationComponent instead 
cmd/preflight/cmd/check_container.go:304:17  staticcheck  SA1019: container.WithCertificationProject is deprecated: Use WithCertificationComponent instead 
```

cmd/preflight/cmd/check_container.go
```diff
@@ -210,28 +225,41 @@ func checkContainerPositionalArgs(cmd *cobra.Command, args []string) error {
...
+	if cmd.Flag("certification-component-id").Changed {
+		cmd.Flag("certification-project-id").Changed = true
+		cmd.Flag("certification-component-id").Changed = false
+		viper.Set("certification_project_id", componentID)
+	}
+
+	// However, if the new env var is set, that's the priority
+	if viper.IsSet("certification_component_id") {
+		viper.Set("certification_project_id", viper.GetString("certification_component_id"))
+	}
```
komish, line 240: Is this only necessary because `certification_component_id` is bound explicitly to env, separate from the global variable?

cmd/preflight/cmd/check_container.go
```diff
@@ -210,28 +225,41 @@ func checkContainerPositionalArgs(cmd *cobra.Command, args []string) error {
...
 
-	// --submit was specified
 	viper := viper.Instance()
+
+	// If the new flag is set, use that
+	if cmd.Flag("certification-component-id").Changed {
+		cmd.Flag("certification-project-id").Changed = true
+		cmd.Flag("certification-component-id").Changed = false
+		viper.Set("certification_project_id", componentID)
+	}
```
komish, line 235: If I'm understanding, this is just  "promotion" of the value from one flag to the other, right? And updating their Changed state is required to avoid MutualExclusion from throwing an error?

Semantic & Nit: it reads weird to me that we're feeding the value of the new flag into the old flag. Disregard at will.

komish (changes requested):
Couple of notes left.

cmd/preflight/cmd/check_container.go
bcrochet: It is awkward, but this makes it so that we don't have to go change a bunch of other places that use the project-id verbage. It does seem very awkward. 

It occurs to me now to that each of these places that are changed should be marked with a TODO so that they are easier to find later.

cmd/preflight/cmd/check_container.go
bcrochet: I try a bunch of different permutations of all this, and this seemed to be the path of least resistance. It's been a bit since I looked at it, so I can't give you a full explanation right now. :)

cmd/preflight/cmd/check_container.go
bcrochet: Correct.

container/check_container.go
bcrochet: That's a good suggestion. I'll implement that.

komish (approved):
/lgtm

PR #1234, opened 2025-01-02
removing kube-rbac-proxy from test data

acornett21:
- Removing `kube-rbac-proxy` from tests data
- Fixes: #1231


komish (approved):
/lgtm

PR #1236, opened 2025-01-10
feat: allow Setuid/Setgid bits to be removed

bcrochet:
Removal of setuid/setgid bits should be allowed, as this can be seen only as increasing security. Caveat emptor should this removal cause a workload to subsequently fail.


acornett21 (approved):

komish (approved):
/lgtm

PR #1240, opened 2025-02-02
Clarify docker-config is required for public Dockerhub registries

damemi:
While submitting images to Red Hat, I noticed we weren't able to pull even though the images are in a public DockerHub registry. Red Hat support explained that the `--docker-config` flag is required for DockerHub registries. Updating here to help others, as I assumed it wasn't needed since this was a public registry.

### DIFF
--- a/docs/RECIPES.md
+++ b/docs/RECIPES.md
@@ -158,7 +158,7 @@ You will also need:
 - A Docker Auth file, can be specified multiple ways
   - If PFLT_DOCKERCONFIG envvar is set, or --docker-config is passed on the command line, this will be used to pull images.
     If also submitting, the file specified will be sent to Red Hat. If the image under test is on a public registry, there is
-    no need to pass this parameter.
+    no need to pass this parameter. For DockerHub registries, this is required even if the registry is public.
 
 ### Testing a Container
 Running container policy checks against a container iteratively until all tests pass.


acornett21:
Hi @damemi, I think there is confusion, if an image is public (ie no auth) we do not require a docker auth file, no matter the backing registry. Are you sure the image is public?

damemi:
Hi @acornett21, sorry, I think there is a lot more detail I could add here and this might not be the right place for it.

Preflight doesn't require the docker config (sorry for the confusion), it's Red Hat's container catalog that requires the docker config when hosting images on `registry.connect.redhat.com`.

However, the docker config needs to be submitted when running preflight, otherwise pulls from `registry.connect.redhat.com` (for the submitted, proxied image) will fail with a 401.

Do you think there is a better place to document this? It seems relevant to the `--submit` section of the preflight docs, or if it's mentioned somewhere else maybe a link to that would be good

acornett21:
It shouldn't be a 401 from the proxy, though is be rate limited and maybe DockerHub returns a 401 for rate limited requests, that's why when using the proxy we ask for a file. I thought this was in our [workflow guide ](https://docs.redhat.com/en/documentation/red_hat_software_certification/2024/html/red_hat_software_certification_workflow_guide/proc_running-the-certification-test-suite_openshift-sw-cert-workflow-working-with-containers#proc_running-the-certification-test-suite_openshift-sw-cert-workflow-working-with-containers), but maybe not, I'll double check.

The workflow guide is the more correct place for this nuance of using the proxy with DockerHub.

damemi:
@acornett21 true, maybe the 401 isn't coming from the proxy but rather from Dockerhub. I totally understand if the workflow is a better place to explain this. I didn't see anything about it there, and I'm happy to open a PR if those docs are open source. Thanks

acornett21:
@damemi I'll put in a Jira to our docs team to update [this](https://docs.redhat.com/en/documentation/red_hat_software_certification/2024/html-single/red_hat_software_certification_workflow_guide/index#certification_testing_for_containerized_applications) section to include a note about DockerHub, and it's low rate limits that require Certification to have auth submitted.

Thanks for pointing this out, I thought this was already documented.

damemi:
@acornett21 sounds good! Thank you

PR #1246, opened 2025-02-26
adding crane retry logic for crane ListTags lookup

acornett21:
## Motivation
We have this logic in our pull logic, but not in our `list-tags` logic, so this check fails if a registry flakes out.

## Changes
Exposing the existing option method that is used in our pull logic and adding it to our `HasUniqueTag` check.

### DIFF
--- a/internal/policy/container/has_unique_tag.go
+++ b/internal/policy/container/has_unique_tag.go
@@ -4,10 +4,12 @@ import (
 	"context"
 	"fmt"
 	"strings"
+	"time"
 
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/internal/authn"
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/internal/check"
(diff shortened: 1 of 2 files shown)


internal/policy/container/has_unique_tag.go
```diff
@@ -58,6 +60,7 @@ func (p *hasUniqueTagCheck) getDataToValidate(ctx context.Context, image string)
 	options := []crane.Option{
 		crane.WithContext(ctx),
 		crane.WithAuthFromKeychain(authn.PreflightKeychain(ctx, authn.WithDockerConfig(p.dockercfg))),
+		option.RetryOnceAfter(5 * time.Second),
```
itroyano, line 63: complete nit: a constant to make it configurable.

komish (approved):
/lgtm

internal/policy/container/has_unique_tag.go
acornett21: I don't really want to expose this out to end users, then have something else to trouble-shoot.

PR #1247, opened 2025-02-27
adding a new container policy for konflux pipeline execution

acornett21:
## Motivation
Today, for konflux all policies are ran and since the introduction of the Red Hat trademark violation policies, we have been unable to ship a new image to be used in konflux. This means they are missing out on some bugfixes and performance enhancements.

## Changes
- ~~Introduce a new `konflux` sub command ie `preflight check konflux` . This cmd builds out the proper `Options` to be used in/by the engine.~~
- Add a new `PolicyKonflux` and corresponding container policy with matching tests.
- Revert `HasRequiredLabelsCheck` to only check for required labels.
- Add `HasProhibitedLabelsCheck` to check for Red Hat Trademark violations within a containers labels, and exclude this from the `PolicyKonflux`.
- Update `ResolveSubmitter` to also look for `konflux` flag so a `NoopSubmitter` is returned even if a user tried to provide a pyixs component/key and the `konflux` env.

### DIFF
--- a/internal/engine/engine.go
+++ b/internal/engine/engine.go
@@ -758,6 +758,7 @@ func InitializeContainerChecks(ctx context.Context, p policy.Policy, cfg Contain
 			&containerpol.MaxLayersCheck{},
 			&containerpol.HasNoProhibitedPackagesCheck{},
 			&containerpol.HasRequiredLabelsCheck{},
+			&containerpol.HasNoProhibitedLabelsCheck{},
 			&containerpol.RunAsNonRootCheck{},
 			&containerpol.HasModifiedFilesCheck{},
 			containerpol.NewBasedOnUbiCheck(pyxis.NewPyxisClient(
@@ -774,6 +775,7 @@ func InitializeContainerChecks(ctx context.Context, p policy.Policy, cfg Contain
 			&containerpol.MaxLayersCheck{},
 			&containerpol.HasNoProhibitedPackagesCheck{},
 			&containerpol.HasRequiredLabelsCheck{},
+			&containerpol.HasNoProhibitedLabelsCheck{},
 			&containerpol.HasModifiedFilesCheck{},
 			containerpol.NewBasedOnUbiCheck(pyxis.NewPyxisClient(
 				cfg.PyxisHost,
@@ -788,6 +790,7 @@ func InitializeContainerChecks(ctx context.Context, p policy.Policy, cfg Contain
 			containerpol.NewHasUniqueTagCheck(cfg.DockerConfig),
 			&containerpol.MaxLayersCheck{},
 			&containerpol.HasRequiredLabelsCheck{},
+			&containerpol.HasNoProhibitedLabelsCheck{},
 			&containerpol.RunAsNonRootCheck{},
 			&containerpol.HasProhibitedContainerName{},
 		}, nil
@@ -797,7 +800,23 @@ func InitializeContainerChecks(ctx context.Context, p policy.Policy, cfg Contain
 			containerpol.NewHasUniqueTagCheck(cfg.DockerConfig),
 			&containerpol.MaxLayersCheck{},
 			&containerpol.HasRequiredLabelsCheck{},
+			&containerpol.HasNoProhibitedLabelsCheck{},
 			&containerpol.HasProhibitedContainerName{},
+		}, nil
+	case policy.PolicyKonflux:
+		return []check.Check{
+			&containerpol.HasLicenseCheck{},
+			containerpol.NewHasUniqueTagCheck(cfg.DockerConfig),
+			&containerpol.MaxLayersCheck{},
+			&containerpol.HasNoProhibitedPackagesCheck{},
+			&containerpol.HasRequiredLabelsCheck{},
+			&containerpol.RunAsNonRootCheck{},
+			&containerpol.HasModifiedFilesCheck{},
+			containerpol.NewBasedOnUbiCheck(pyxis.NewPyxisClient(
+				cfg.PyxisHost,
+				cfg.PyxisAPIToken,
+				cfg.CertificationProjectID,
+				&http.Client{Timeout: 60 * time.Second})),
 		}, nil
 	}
 
@@ -819,7 +838,7 @@ func makeCheckList(checks []check.Check) []string {
 func checkNamesFor(ctx context.Context, p policy.Policy) []string {
 	var c []check.Check
 	switch p {
-	case policy.PolicyContainer, policy.PolicyRoot, policy.PolicyScratchNonRoot, policy.PolicyScratchRoot:
+	case policy.PolicyContainer, policy.PolicyRoot, policy.PolicyScratchNonRoot, policy.PolicyScratchRoot, policy.PolicyKonflux:
 		c, _ = InitializeContainerChecks(ctx, p, ContainerCheckConfig{})
 	case policy.PolicyOperator:
 		c, _ = InitializeOperatorChecks(ctx, p, OperatorCheckConfig{})
@@ -856,4 +875,10 @@ func ScratchRootContainerPolicy(ctx context.Context) []string {
 // container policy with root exception.
 func RootExceptionContainerPolicy(ctx context.Context) []string {
 	return checkNamesFor(ctx, policy.PolicyRoot)
+}
+
+// KonfluxContainerPolicy returns the names of checks to be used in
+// a konflux pipeline
+func KonfluxContainerPolicy(ctx context.Context) []string {
+	return checkNamesFor(ctx, policy.PolicyKonflux)
 }
--- a/internal/policy/container/has_prohibited_labels.go
+++ b/internal/policy/container/has_prohibited_labels.go
@@ -1,0 +1,65 @@
+package container
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/internal/check"
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/internal/image"
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/internal/log"
+
+	"github.com/go-logr/logr"
+)
+
+var trademarkLabels = []string{"name", "vendor", "maintainer"}
+
+var _ check.Check = &HasNoProhibitedLabelsCheck{}
+
+type HasNoProhibitedLabelsCheck struct{}
+
+func (p *HasNoProhibitedLabelsCheck) Validate(ctx context.Context, imgRef image.ImageReference) (result bool, err error) {
+	labels, err := getContainerLabels(imgRef.ImageInfo)
+	if err != nil {
+		return false, fmt.Errorf("could not retrieve image labels: %v", err)
+	}
+
+	return p.validate(ctx, labels)
+}
+
+func (p *HasNoProhibitedLabelsCheck) validate(ctx context.Context, labels map[string]string) (bool, error) {
+	logger := logr.FromContextOrDiscard(ctx)
+
+	trademarkViolationLabels := []string{}
+	for _, label := range trademarkLabels {
+		if violatesRedHatTrademark(labels[label]) {
+			trademarkViolationLabels = append(trademarkViolationLabels, label)
+		}
+	}
+
+	// TODO: We should be reporting this in the results, not in a log message
+	if len(trademarkViolationLabels) > 0 {
+		logger.V(log.DBG).Info("labels violate Red Hat trademark", "trademarkViolationLabels", trademarkViolationLabels)
+	}
+
+	return len(trademarkViolationLabels) == 0, nil
+}
+
+func (p *HasNoProhibitedLabelsCheck) Name() string {
+	return "HasNoProhibitedLabels"
+}
+
+func (p *HasNoProhibitedLabelsCheck) Metadata() check.Metadata {
+	return check.Metadata{
+		Description:      "Checking if the labels (name, vendor, maintainer) violate Red Hat trademark.",
+		Level:            "good",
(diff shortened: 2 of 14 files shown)


cmd/preflight/cmd/list_checks.go
```diff
@@ -38,6 +38,7 @@ func printChecks(w io.Writer) {
 		"automatically applied for container checks if preflight determines a scratch exception flag has been added to your Red Hat Connect project"))
 	fmt.Fprintln(w, formattedPolicyBlock("Container Scratch (Root) Exception", engine.ScratchRootContainerPolicy(context.TODO()),
 		"automatically applied for container checks if preflight determines scratch and root exception flags have both been added to your Red Hat Connect project"))
+	// todo-adam should we add konflux here?
```
bcrochet, line 41: My gut says no, but I'm open to the idea.

internal/policy/container/has_required_labels.go
```diff
@@ -86,3 +67,9 @@ func (p *HasRequiredLabelsCheck) Help() check.HelpText {
 			"and validate that they do not violate Red Hat trademark.",
 	}
 }
+
+// getDataForValidate is a helper function to obtain the labels from an images configfile
+func getDataForValidate(image cranev1.Image) (map[string]string, error) {
```
bcrochet, line 72: Might move this to a `labels.go` or something. Since it is shared, it should be easier to find.

internal/runtime/config.go
```diff
@@ -74,6 +75,8 @@ func (c *Config) storeContainerPolicyConfiguration(vcfg viper.Viper) {
 	c.Platform = vcfg.GetString("platform")
 	c.Insecure = vcfg.GetBool("insecure")
 	c.Offline = vcfg.GetBool("offline")
+	// todo should we have a flag for this? trying to hide this option as much as possible
```
bcrochet, line 78: No flag. Env only.

bcrochet (changes requested):
Responses to some of the todos, and a suggestion about shared code.

cmd/preflight/cmd/list_checks.go
acornett21: My gut said no as well, we'll let others chime in.

internal/engine/engine.go
```diff
@@ -797,8 +800,24 @@ func InitializeContainerChecks(ctx context.Context, p policy.Policy, cfg Contain
...
+	case policy.PolicyKonflux:
+		return []check.Check{
+			&containerpol.HasLicenseCheck{},
+			containerpol.NewHasUniqueTagCheck(cfg.DockerConfig),
+			&containerpol.MaxLayersCheck{},
+			&containerpol.HasNoProhibitedPackagesCheck{},
+			&containerpol.HasRequiredLabelsCheck{},
+			&containerpol.RunAsNonRootCheck{},
+			&containerpol.HasModifiedFilesCheck{},
+			containerpol.NewBasedOnUbiCheck(pyxis.NewPyxisClient(
```
komish, line 815: Is this one of those checks that Konflux may want to opt out of?

internal/policy/container/labels.go
```diff
@@ -0,0 +1,9 @@
+package container
```
komish, line 1: This file seems extraneous. If you want this to be a library function that can be re-used by both "prohibited" and "required" label checks, might be better to move this to another package. Here, it feels like cruft.

internal/policy/container/has_prohibited_labels.go
```diff
@@ -0,0 +1,65 @@
...
+	logger := logr.FromContextOrDiscard(ctx)
+
+	trademarkViolationLabels := []string{}
+	for _, label := range trademarkLabels {
+		if violatesRedHatTrademark(labels[label]) {
+			trademarkViolationLabels = append(trademarkViolationLabels, label)
+		}
+	}
+
+	// TODO: We should be reporting this in the results, not in a log message
```
komish, line 39: Does this TODO need addressing before merge?

internal/policy/container/has_prohibited_labels.go
```diff
@@ -0,0 +1,65 @@
...
+	return len(trademarkViolationLabels) == 0, nil
+}
+
+func (p *HasProhibitedLabelsCheck) Name() string {
+	return "HasProhibitedLabelsCheck"
+}
+
+func (p *HasProhibitedLabelsCheck) Metadata() check.Metadata {
+	return check.Metadata{
+		Description:      "Checking if the labels (name, vendor, maintainer) violate Red Hat trademark.",
```
komish, line 53: Clarity: should this be something like more generic than "violate the Red Hat trademark", such as "contains prohibited content such as trademark violations"?

internal/policy/container/has_prohibited_labels.go
```diff
@@ -0,0 +1,65 @@
...
+		Level:            "good",
+		KnowledgeBaseURL: certDocumentationURL,
+		CheckURL:         certDocumentationURL,
+	}
+}
+
+func (p *HasProhibitedLabelsCheck) Help() check.HelpText {
+	return check.HelpText{
+		Message:    "Check HasRequiredLabel encountered an error. Please review the preflight.log file for more information.",
```
komish, line 62:
```suggestion
		Message:    "Check HasProhibitedLabel encountered an error. Please review the preflight.log file for more information.",
```

internal/lib/lib.go
```diff
@@ -10,11 +10,11 @@ import (
...
 
-// ResolveSubmitter will build out a ResultSubmitter if the provided pyxisClient, pc, is not nil.
-// The pyxisClient is a required component of the submitter. If pc is nil, then a noop submitter
-// is returned instead, which does nothing.
-func ResolveSubmitter(pc PyxisClient, projectID, dockerconfig, logfile string) ResultSubmitter {
-	if pc != nil {
+// ResolveSubmitter will build out a ResultSubmitter if the provided pyxisClient, pc, is not nil,
+// The pyxisClient is a required component of the submitter. If pc is nil or if konflux is false,
+// then a noop submitter is returned instead, which does nothing.
+func ResolveSubmitter(pc PyxisClient, projectID, dockerconfig, logfile string, konflux bool) ResultSubmitter {
```
komish, line 16: NIT: 

```suggestion
func ResolveSubmitter(pc PyxisClient, projectID, dockerconfig, logfile string, isKonflux bool) ResultSubmitter {
```

internal/lib/lib.go
```diff
@@ -10,11 +10,11 @@ import (
...
 )
 
-// ResolveSubmitter will build out a ResultSubmitter if the provided pyxisClient, pc, is not nil.
-// The pyxisClient is a required component of the submitter. If pc is nil, then a noop submitter
-// is returned instead, which does nothing.
-func ResolveSubmitter(pc PyxisClient, projectID, dockerconfig, logfile string) ResultSubmitter {
-	if pc != nil {
+// ResolveSubmitter will build out a ResultSubmitter if the provided pyxisClient, pc, is not nil,
+// The pyxisClient is a required component of the submitter. If pc is nil or if konflux is false,
+// then a noop submitter is returned instead, which does nothing.
```
komish, line 15:
```suggestion
// then a noop submitter is returned instead.
```

internal/lib/lib.go
komish: Also, I think you can get by without changing this function signature.

When callers call ResolveSubmitter, they already have the context as to whether they're in a Konflux environment. As I understand it, they can just _skip calling ResolveSubmitter_ instead of passing that information into the check because they have the same information this function has.

Less code to change here, IMO.

internal/policy/container/has_prohibited_labels.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/internal/log"
+
+	"github.com/go-logr/logr"
+)
+
+var trademarkLabels = []string{"name", "vendor", "maintainer"}
+
+var _ check.Check = &HasProhibitedLabelsCheck{}
+
+type HasProhibitedLabelsCheck struct{}
```
komish, line 18: This naming is unclear to me. This check asserts that labels don't have prohibited value, which *today* translates into trademark violations. This check does *not* check that given label keys are missing, as the name might suggest.

internal/policy/container/has_prohibited_labels.go
```diff
@@ -0,0 +1,65 @@
...
+		Level:            "good",
+		KnowledgeBaseURL: certDocumentationURL,
+		CheckURL:         certDocumentationURL,
+	}
+}
+
+func (p *HasProhibitedLabelsCheck) Help() check.HelpText {
+	return check.HelpText{
+		Message:    "Check HasRequiredLabel encountered an error. Please review the preflight.log file for more information.",
+		Suggestion: "Ensure the following labels to your Dockerfile or Containerfile: name, vendor, maintainer do not violate Red Hat trademark.",
```
komish, line 63:
```suggestion
		Suggestion: "Ensure the name, vendor, and maintainer label on your image do not violate the Red Hat trademark.",
```

komish (changes requested):
Comments left in-line.

internal/policy/container/labels.go
acornett21: @bcrochet suggested this, so that it's easier to find, since I had originally kept it in the original labels check.

internal/policy/container/has_prohibited_labels.go
acornett21: No, this TODO is there since the beginning, just carrying it over to this check.

internal/policy/container/has_prohibited_labels.go
acornett21: I don't think so, this is what we had before, and the policy docs have a link to Red Hat Trademark docs.

internal/engine/engine.go
acornett21: I think that's out of scope for this, if an image is UBI then none of these checks apply so the caller in konflux can remove preflight from their pipeline.

internal/lib/lib.go
acornett21: No, that's incorrect, this is needed for the case that a partner looks at our code and see's a env to bypass checks, and then passes in the flags to build a pyxis client. If this code was not there we would allow less checks to be executed by a partner that read our code. Basically this is a safeguard for that situation.

internal/policy/container/has_prohibited_labels.go
acornett21: This has always been a problem with our naming, I can change it to `HasNoProhibitedLabelsCheck`

komish: Eh. I didn't catch that this was the old name. I don't love it, given what it's doing now, but I suppose I can't  say much given the old name was basically this.

komish: Up to you how you want to handle this.

internal/lib/lib.go
komish: I'm not sure I'm following. You can forcefully set a NoopSubmitter based on cfg.Konflux at the call point. I understand wanting to make sure that the partner doesn't find themselves in a situation they don't want, but the way I see it, it's the same precaution either way.

The way I see it, if we had followed a Konflux-only flow, we wouldn't be doing any submitter resolution at all, and would just be passing in a hard-coded NoopSubmitter.

Up to you; feel free to resolve if you think this is sufficient.

internal/policy/container/has_prohibited_labels.go
bcrochet: It might be time to do a review of our check naming? As @komish has stated, some of them are just not named in a way that is unambiguous.

internal/policy/container/labels.go
bcrochet: It's being used in two files. It does seem extra, but I actually thought there should probably be more shared code that could end up in here. I say establish this file now, and we refactor these two checks to strip out more of the shared code.

internal/runtime/config.go
```diff
@@ -74,6 +75,7 @@ func (c *Config) storeContainerPolicyConfiguration(vcfg viper.Viper) {
 	c.Platform = vcfg.GetString("platform")
 	c.Insecure = vcfg.GetBool("insecure")
 	c.Offline = vcfg.GetBool("offline")
+	c.Konflux = vcfg.GetBool("konflux")
```
bcrochet, line 78: Is this enough to get set? Setting `PFLT_KONFLUX=1` in the environment enables this?

bcrochet (commented):
Couple more questions and comments.

cmd/preflight/cmd/list_checks.go
acornett21: I removed this todo, and we'll leave this out of the `list-checks` cmd.

internal/policy/container/has_prohibited_labels.go
acornett21: I changed this `HasNoProhibitedLabelsCheck`, I hope this helps.

internal/runtime/config.go
acornett21: Yes, `PFLT_KONFLUX=1` or `PFLT_KONFLUX=true` are all that's needed for viper to do it's magic.

internal/policy/container/labels.go
komish: Understood. The final point I'll stand on here is that `getDataForValidate` is a common name used through all checks. It's not a required naming convention by any means, but it's just been the common practice. It can change at any time.

That said, this name is used as a method on one of the Check types. Here, it's in the global namespace for the package, which I think makes it easier to confuse things.

Can we name this something _else_?

It's a bit of a nit, but if too many small nits get through, things can get messy fast.

acornett21: Renamed this to `getContainerLabels`

Issue #1263, opened 2025-04-28
Update `HasUniqueTag` Logic to Only Use One Page of Data


acornett21:
## Issue
Currently `cranes` library implementation of `ListTags` uses `pagination` and gets `all` of the tags for a given container. This can result is slow performance of this check for images that have thousands/millions of tags.

## Solution
Move away from calling `ListTags` directly and use other `crane` methods to get the same details, but do not iterate. Sample logic might be something similar to the below, but without the `for` loop.

https://github.com/google/go-containerregistry/blob/main/cmd/crane/cmd/list.go#L48-L80

PR #1268, opened 2025-05-01
Only check first page of tags in `HasUniqueTag`

caxu-rh:
An image passes the check if it has more than one tag (`latest` can be one of them) or if it has only one tag which is not `latest`.

We can validate these scenarios using just the first page of tags returned from the registry, rather than requesting all tags for a repository which may be a lot more.

---

Some follow-up areas of conversation/questions annotated below.

Addresses #1263.

### DIFF
--- a/internal/policy/container/has_unique_tag.go
+++ b/internal/policy/container/has_unique_tag.go
@@ -6,12 +6,15 @@ import (
 	"strings"
 	"time"
 
+	"github.com/go-logr/logr"
+
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/internal/authn"
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/internal/check"
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/internal/image"
-	"github.com/redhat-openshift-ecosystem/openshift-preflight/internal/option"
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/internal/log"
 
-	"github.com/google/go-containerregistry/pkg/crane"
+	"github.com/google/go-containerregistry/pkg/name"
+	"github.com/google/go-containerregistry/pkg/v1/remote"
 )
 
 var _ check.Check = &hasUniqueTagCheck{}
@@ -30,6 +33,7 @@ type hasUniqueTagCheck struct {
 }
 
 func (p *hasUniqueTagCheck) Validate(ctx context.Context, imgRef image.ImageReference) (bool, error) {
+	logger := logr.FromContextOrDiscard(ctx)
 	imgRepo := fmt.Sprintf("%s/%s", imgRef.ImageRegistry, imgRef.ImageRepository)
 
 	tags := make([]string, 0)
@@ -40,6 +44,8 @@ func (p *hasUniqueTagCheck) Validate(ctx context.Context, imgRef image.ImageRefe
 		if err != nil {
 			return false, fmt.Errorf("failed to get tags list for %s: %v", imgRepo, err)
 		}
+
+		logger.V(log.DBG).WithValues("tags", tags).Info(fmt.Sprintf("got tags list for %s", imgRepo))
 	}
 
 	// if tags is of length zero we know that either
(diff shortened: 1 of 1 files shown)


internal/policy/container/has_unique_tag.go
```diff
@@ -57,13 +57,48 @@ func (p *hasUniqueTagCheck) Validate(ctx context.Context, imgRef image.ImageRefe
...
 func (p *hasUniqueTagCheck) getDataToValidate(ctx context.Context, image string) ([]string, error) {
-	options := []crane.Option{
-		crane.WithContext(ctx),
-		crane.WithAuthFromKeychain(authn.PreflightKeychain(ctx, authn.WithDockerConfig(p.dockercfg))),
-		option.RetryOnceAfter(5 * time.Second),
+	repo, err := name.NewRepository(image)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse image name: %v", err)
+	}
+	puller, err := remote.NewPuller([]remote.Option{
```
caxu-rh, line 64: The code was previously using options provided by `github.com/redhat-openshift-ecosystem/openshift-preflight/internal/option`, which supplied `crane.Option`s, but we now need `remote.Option`s.

Not sure if it's preferable to modify `github.com/redhat-openshift-ecosystem/openshift-preflight/internal/option` for this or directly use the upstream package (I might be missing the reason to bring that internal in the first place).

internal/policy/container/has_unique_tag.go
```diff
@@ -57,13 +57,48 @@ func (p *hasUniqueTagCheck) Validate(ctx context.Context, imgRef image.ImageRefe
 }
 
 func (p *hasUniqueTagCheck) getDataToValidate(ctx context.Context, image string) ([]string, error) {
-	options := []crane.Option{
-		crane.WithContext(ctx),
-		crane.WithAuthFromKeychain(authn.PreflightKeychain(ctx, authn.WithDockerConfig(p.dockercfg))),
-		option.RetryOnceAfter(5 * time.Second),
+	repo, err := name.NewRepository(image)
```
caxu-rh, line 60: There are some options for `name.NewRepository` which set default registry and tag if those are not specified (e.g. `nginx` -> `docker.io/library/nginx:latest`). I have not set them here yet - should they be set as Docker Hub (`docker.io`) and `latest`?

internal/policy/container/has_unique_tag.go
```diff
@@ -57,13 +57,48 @@ func (p *hasUniqueTagCheck) Validate(ctx context.Context, imgRef image.ImageRefe
...
+
+	if lister.HasNext() {
+		tags, err := lister.Next(ctx)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get tags: %v", err)
+		}
+
+		results := []string{}
+		for _, tag := range tags.Tags {
+			if strings.HasPrefix(tag, "sha256-") {
```
caxu-rh, line 91: This condition is borrowed from code in `go-containerregistry`, which seems to identify digest-based tags as starting with `sha256-`, while other code in Preflight seems to identify digest-based tags as starting with `sha256:`. Not sure which is right here.

internal/policy/container/has_unique_tag.go
acornett21: We allow partners to use "any" registry of their choosing, as long as it's  [Open Container Initiative](https://opencontainers.org/) (OCI) compliant.

We only use `name.NewRepository` in tests and there we set options `WeakValidation`, but I see that this is default. At face value, I think what is here will work. We can test a few repos and see.

https://github.com/google/go-containerregistry/blob/098045d5e61ff426a61a0eecc19ad0c433cd35a9/pkg/name/options.go#L29

internal/policy/container/has_unique_tag.go
acornett21: I noticed the switch in packages for the `Option` struct when looking to just add `WithPageSize`, as well. The reason preflight has it's own internal option, so so that we could have our own keychain to handle auth as we need, and how we need it when pulling the content. 

This check has always had it's own options, and just relied on only the keychain. We have to switch to `remote.Options` since that's what the method takes, so as long as we preserve all of the `.with` I think we should be fine here.

caxu-rh: Leaving this as-is for now, using the option types from `remote`.

internal/policy/container/has_unique_tag.go
```diff
@@ -57,13 +57,44 @@ func (p *hasUniqueTagCheck) Validate(ctx context.Context, imgRef image.ImageRefe
...
+
+	if lister.HasNext() {
+		tags, err := lister.Next(ctx)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get tags: %v", err)
+		}
+
+		results := []string{}
+		results = append(results, tags.Tags...)
+		return results, nil
```
komish, line 94: I think you can remove results and just return tags.Tags
```suggestion
		return tags.Tags, nil
```

internal/policy/container/has_unique_tag.go
```diff
@@ -57,13 +57,44 @@ func (p *hasUniqueTagCheck) Validate(ctx context.Context, imgRef image.ImageRefe
...
+		return nil, fmt.Errorf("failed to create puller: %v", err)
+	}
+
+	lister, err := puller.Lister(ctx, repo)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create lister: %v", err)
+	}
+
+	if lister.HasNext() {
+		tags, err := lister.Next(ctx)
```
komish, line 87: You may need to do a nil check on `tags`.

caxu-rh: Thanks, adjusted - now treating a nil `tags` as if there are no tags available (not as an error condition), please advise if that's a bad assumption.

internal/policy/container/has_unique_tag.go
```diff
@@ -57,13 +57,44 @@ func (p *hasUniqueTagCheck) Validate(ctx context.Context, imgRef image.ImageRefe
...
-	return crane.ListTags(image, options...)
+	options := []remote.Option{
+		remote.WithContext(ctx),
+		remote.WithAuthFromKeychain(authn.PreflightKeychain(ctx, authn.WithDockerConfig(p.dockercfg))),
+		remote.WithRetryBackoff(remote.Backoff{
+			Duration: 5 * time.Second,
+			Factor:   1.0,
+			Jitter:   0.1,
+			Steps:    2,
+		}),
```
komish, line 73: Do we see any benefit in restricting the number of tags we want via the [remote.WithPageSize](https://github.com/google/go-containerregistry/blob/main/pkg/v1/remote/options.go#L288-L297) option? If I'm reading correctly, I see the default set to [1000](https://github.com/google/go-containerregistry/blob/main/pkg/v1/remote/options.go#L108), so we're still doing a large-ish pull for something we can evaluate with a smaller pull.

internal/policy/container/has_unique_tag.go
komish: Should be fine.

internal/policy/container/has_unique_tag.go
komish: cc @acornett21

acornett21: We could probably add this now that we aren't getting multiple pages. I'm not sure what value we would want, technically we only care if there is `one more tag then latest` , but going with just `WithPageSize(2)` seems risky.

https://github.com/redhat-openshift-ecosystem/openshift-preflight/blob/4b330505988d594b55cbb9e022b5669d7f19e179/internal/policy/container/has_unique_tag.go#L74

komish: For me, anything smaller than 10 seems fit for purpose while also reducing the query.  Sound reasonable enough?

Also, worth adding a log line here at DEBUG/TRACE verbosity, listing "which tags were returned"?

komish: @caxu-rh Any concern with reducing the page size? You can decide on the value here. Just add the log line so we know what tags were pulled in the event of an issue and I think this should be set.

komish (approved):
/lgtm

PR #1280, opened 2025-06-24
Add test cases to improve test coverage

caxu-rh:
~~This change also sets the test image `src` in `cmd/preflight/cmd/check_container_test.go` to have architecture `runtime.GOOS` instead of hardcoded to `amd64`. Tests which use this image expect the image to be the same architecture as the machine where tests are running (usually amd64 in CI, but could be arm64 if running tests locally).~~

Some test fixtures are only available for amd64, so leaving it as-is.

### DIFF
--- a/internal/policy/container/has_modified_files_test.go
+++ b/internal/policy/container/has_modified_files_test.go
@@ -473,6 +473,40 @@ var _ = Describe("HasModifiedFiles", func() {
 		})
 	})
 
+	When("parsing the platform distribution value from /etc/os-release", func() {
+		var mmfs afero.Fs
+		BeforeEach(func() {
+			mmfs = afero.NewMemMapFs()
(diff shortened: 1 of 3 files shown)


internal/policy/container/has_modified_files_test.go
```diff
@@ -473,6 +473,40 @@ var _ = Describe("HasModifiedFiles", func() {
 		})
 	})
 
+	When("parsing the platform distribution value from /etc/os-release", func() {
+		var mmfs afero.Fs
+		BeforeEach(func() {
+			mmfs = afero.NewMemMapFs()
+			Expect(mmfs.Mkdir("/etc", 0o755)).Should(Succeed())
```
bcrochet, line 480: Very nit but `To` instead of `Should`?

PR #1286, opened 2025-07-07
adding maintainer label to test docker files

acornett21:
- Adding `maintainer` label so that developers that build with these files for their own testing, can pass preflights checks.


komish (approved):
/lgtm

Issue #1016, opened 2023-06-16
Update GH Action for `main` to Run Preflight Image


acornett21:
## Is your feature request related to a problem? Please describe.
We have ran into a few issues where the image build by `main` can not be executed. This usually has lead to alot of after the fact troubleshooting.

## Describe the solution you'd like.
After our image is built, and before we push the image to quay, we should put a blocking step in our process to execute

```
podman run -it --rm <image-id-from-build-process> version
```

The above will make sure that the binary built, as well as the container can be used and contains no errors in either the binary or the image itself.

## Describe alternatives you've considered.
NA

## Additional context.
NA

PR #1296, opened 2025-07-31
add a run validate step to main to make sure image is built correctly

acornett21:
- Fixes: #1016 

Assisted-by: Gemini

### DIFF
--- a/.github/workflows/build-main.yml
+++ b/.github/workflows/build-main.yml
@@ -42,6 +42,15 @@ jobs:
         dockerfiles: |
           ./Dockerfile
 
+    - name: Run and Validate Image
+      run: |
+        if ! podman run --rm ${IMAGE_WITH_TAG} version; then
+            echo "Image validation failed. The 'podman run' command returned a non-zero exit code."
+            exit 1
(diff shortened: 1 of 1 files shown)


acornett21:

.github/workflows/build-main.yml
```diff
@@ -42,6 +42,13 @@ jobs:
         dockerfiles: |
           ./Dockerfile
 
+    - name: Run and Validate Image
+      run: |
+        if ! podman run --rm ${{ steps.build-image.outputs.image-with-tag }} version; then
+            echo "Image validation failed. The 'podman run' command returned a non-zero exit code."
+            exit 1
+        fi
+
```
caxu-rh, line 51: **Nit:** Assign `${{ steps.build-image.outputs.image-with-tag }}` to an environment variable in the GitHub Actions step (as a key in a new `env:`), then use the environment variable in the inline script.

Not a blocker; if proceeding without this adjustment, I can also include it as part of #1298.

acornett21: Fix.

PR #1298, opened 2025-08-01
Improve security of GitHub Actions workflows

caxu-rh:
This change addresses several attack vectors and weaknesses which have been seen in the wild over the past few years:
- [Template injection](https://securitylab.github.com/resources/github-actions-untrusted-input/). Instead of directly referencing GitHub Actions variables `${{ ... }}` in scripts, where they are added literally, assign the value to an environment variable and use the environment variable in the script. The shell will expand the environment variable and escape any problematic inputs.
- [Pin actions](https://www.stepsecurity.io/blog/pinning-github-actions-for-enhanced-security-a-complete-guide). Third-party actions are often referenced by tags, which are floating and can be re-assigned to a different ref. Instead, refer to third-party actions by a full Git SHA-1 commit hash. The intended tag is still left as a comment, which Dependabot can use for automatic updates. In March 2025, [a supply chain attack conducted on the `tj-actions/changed-files` action](https://github.com/advisories/GHSA-mrrh-fwg8-r2c3) resulted in the potential compromise of secrets for any workflows which ran the action without pinning it to a known good version.
- ~~[Excessive token permissions](https://docs.github.com/en/actions/tutorials/authenticate-with-github_token). The GitHub token used inside the workflow run has write access to most things in the repo, unless permissions are explicitly specified. [Tools are also available to audit workflows and identify the minimal required set of permissions](https://github.com/GitHubSecurityLab/actions-permissions).~~ We'll tackle this in a follow-up PR.
- [Don't persist credentials in Git checkouts](https://yossarian.net/til/post/actions-checkout-can-leak-github-credentials/). By default, when using the `actions/checkout` action, a copy of the workflow credential (e.g. `GITHUB_TOKEN`) will be persisted in the checked-out repository's `.git/config`. This makes it easier to perform some further operations in the repo, e.g. pushing a commit or tag later. This can pose an issue if the checked-out repository is then uploaded as an artifact, which can be publicly downloaded.

This change also introduces scanning using [zizmor](https://zizmor.sh/), a static analyzer for GitHub Actions workflows, to run on pushes to main and all pull requests.

### DIFF
--- a/.github/workflows/build-multiarch.yml
+++ b/.github/workflows/build-multiarch.yml
@@ -55,7 +55,7 @@ jobs:
 
       # Authenticate to container image registry to push the image
       - name: Podman Login
-        uses: redhat-actions/podman-login@v1
+        uses: redhat-actions/podman-login@4934294ad0449894bcd1e9f191899d7292469603 # v1.7
         with:
           registry: ${{ secrets.registry }}
           username: ${{ secrets.user }}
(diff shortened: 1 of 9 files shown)


.github/workflows/build-multiarch.yml
```diff
@@ -55,37 +55,49 @@ jobs:
...
         if: ${{ inputs.sign == true && github.event.release && github.event.action == 'published' }}
         run: |
           cosign verify \
-            --certificate-identity https://github.com/${{ github.repository }}/.github/workflows/build-multiarch.yml@refs/tags/${{ inputs.tag }} \
+            --certificate-identity https://github.com/${{ github.repository }}/.github/workflows/build-multiarch.yml@refs/tags/${INPUT_TAG} \
             --certificate-oidc-issuer https://token.actions.githubusercontent.com \
-            ${{ secrets.registry }}/${{ inputs.name }}:${{ inputs.tag }}
+            ${{ secrets.registry }}/${INPUT_NAME}:${INPUT_TAG}
+        env:
+          INPUT_NAME: ${{ inputs.name }}
```
acornett21, line 102: Is name used here? Or is this just for consistency?

acornett21:
It will probably take a couple of passes to double check everything on this. Since this is such a sweeping change, I think once we feel this is no longer WIP, we setup your fork to cut releases to your own quay and test those workflows as well.

caxu-rh:
I've reduced the scope of this PR to skip explicitly adding permissions on jobs/workflows for now, instead just leaving TODOs.

In follow-up PRs, we should audit the necessary scope of the tokens for each job using a tool like https://github.com/GitHubSecurityLab/actions-permissions, and then assign the necessary permissions based on findings.

acornett21 (approved):

bcrochet (commented):
Would like to see a final pass that all of the release tags for actions used are up-to-date. Otherwise, lgtm

bcrochet (approved):
/lgtm

PR #1301, opened 2025-08-08
updating GetPackageList to support ubi10-micro

acornett21:
## Motivation
- It seems that UBI 10 Micro's rpm db has moved locations, causing users of this version/variant of UBI to fail preflight. And historical symlink is not present, like in other UBI flavors.
- Context in the [upstream](https://fedoraproject.org/wiki/Changes/RelocateRPMToUsr) driving the rpmdb move.

## Changes
- changed rpmdbPath to a list that can be extensible, since not all paths are in `/var/lib/rpm`
- changed from nested `if` logic to a loop to iterate over rpmdbPaths

Assisted-by: Claude

### DIFF
--- a/internal/rpm/rpm.go
+++ b/internal/rpm/rpm.go
@@ -12,21 +12,32 @@ import (
 	_ "github.com/glebarez/go-sqlite"
 )
 
-// GetPackageList returns the list of packages in the rpm database from either
-// /var/lib/rpm/rpmdb.sqlite, or /var/lib/rpm/Packages if the former does not exist.
+// GetPackageList returns the list of packages in the rpm database from
+// /var/lib/rpm/rpmdb.sqlite, /var/lib/rpm/Packages or /usr/lib/sysimage/rpm/rpmdb.sqlite.
 // If neither exists, this returns an error of type os.ErrNotExists
 func GetPackageList(ctx context.Context, basePath string) ([]*rpmdb.PackageInfo, error) {
-	rpmdirPath := filepath.Join(basePath, "var", "lib", "rpm")
-	rpmdbPath := filepath.Join(rpmdirPath, "rpmdb.sqlite")
+	rpmdbPaths := []string{
+		// Explicitly check /usr/lib/sysimage/rpm. A compatibility symlink from
+		// /var/lib/rpm may not necessarily exist.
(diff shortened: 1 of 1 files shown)


internal/rpm/rpm.go
```diff
@@ -12,21 +12,29 @@ import (
...
 // If neither exists, this returns an error of type os.ErrNotExists
 func GetPackageList(ctx context.Context, basePath string) ([]*rpmdb.PackageInfo, error) {
-	rpmdirPath := filepath.Join(basePath, "var", "lib", "rpm")
-	rpmdbPath := filepath.Join(rpmdirPath, "rpmdb.sqlite")
-
-	if _, err := os.Stat(rpmdbPath); errors.Is(err, os.ErrNotExist) {
-		// rpmdb.sqlite doesn't exist. Fall back to Packages
-		rpmdbPath = filepath.Join(rpmdirPath, "Packages")
+	rpmdbPaths := []string{
+		filepath.Join(basePath, "usr", "lib", "sysimage", "rpm", "rpmdb.sqlite"),
```
caxu-rh, line 20: **Nit:** Consider adding a comment to record why we need to check both locations. Otherwise, a future reading of changelogs such as https://fedoraproject.org/wiki/Changes/RelocateRPMToUsr would suggest that this is unnecessary because the compatibility symlink _should_ exist.

Something like:

```suggestion
		// Explicitly check /usr/lib/sysimage/rpm. A compatibility symlink from
		// /var/lib/rpm may not necessarily exist.
		filepath.Join(basePath, "usr", "lib", "sysimage", "rpm", "rpmdb.sqlite"),
```

internal/rpm/rpm.go
```diff
@@ -12,21 +12,31 @@ import (
...
+		var err error
+		if _, err = os.Stat(path); err == nil {
+			rpmdbPath = path
+			break
 		}
+		errs = append(errs, err)
+	}
+
+	if rpmdbPath == "" {
+		return nil, errors.Join(errs...)
```
bcrochet, line 39: This does still need the extra context that we had before. Just the arg to fmt.Errorf needed to change.

bcrochet (changes requested):
Comment about the error context.

komish:
Underlying context in the upstream driving the rpmdb move. 

https://fedoraproject.org/wiki/Changes/RelocateRPMToUsr

/lgtm

PR #1304, opened 2025-08-12
moving away from docker.io for base image due to rate limiting

acornett21:
## Motivation
- docker.io has a very low rate limit for 
- DCI now runs in a connected env
- This means our jobs will always fail

## Changes
- Move to `go-toolset` image from `registry.access.redhat.com`, this registry is already used for the final image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,11 +3,15 @@ ARG release_tag=0.0.0
 ARG ARCH=amd64
 ARG OS=linux
 
-FROM docker.io/golang:1.24 AS builder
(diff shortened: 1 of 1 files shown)


tonyskapunk:
> from change #1304:
> 
>     * FAILURE https://www.distributed-ci.io/jobs/716bec75-a2ec-4981-b11a-570aea364b66/jobStates

This failure will be addressed by https://github.com/redhatci/ansible-collection-redhatci-ocp/pull/767 in connected environments.

bcrochet (commented):
EDIT: Never mind. Figured it out.

PR #1314, opened 2025-09-15
Capture controller pod logs and status as part of artifacts

caxu-rh:
- Add new functions for getting the pods which match a deployment's selector label
- Add new functions for getting logs from a pod's container(s)
  - This introduces a new dependency, `k8s.io/client-go`, since `controller-runtime` basically only covers CRUD operations on resources, not logs
- Add logic in DeployableByOlmCheck to use these functions and write as artifacts

### DIFF
--- a/internal/policy/operator/deployable_by_olm.go
+++ b/internal/policy/operator/deployable_by_olm.go
@@ -657,6 +657,41 @@ func (p *DeployableByOlmCheck) cleanUp(ctx context.Context, operatorData operato
 		if err := p.writeToFile(ctx, deployment); err != nil {
 			logger.Error(err, "could not write deployment to storage")
 		}
+
+		pods, err := p.openshiftClient.GetDeploymentPods(ctx, deploymentName, operatorData.InstallNamespace)
+		if err != nil {
+			logger.Info(fmt.Sprintf("warning: unable to retrieve deployment pods: %s", err))
+			continue
+		}
+
+		for _, pod := range pods {
+			jsonManifest, err := json.Marshal(pod.Status)
+			if err != nil {
+				logger.Error(err, "unable to marshal to json")
+			}
+
+			filename := fmt.Sprintf("%s-PodStatus.json", pod.Name)
+			if artifactWriter := artifacts.WriterFromContext(ctx); artifactWriter != nil {
+				if _, err := artifactWriter.WriteFile(filename, bytes.NewReader(jsonManifest)); err != nil {
+					logger.Error(err, "failed to write the PodStatus to the file")
+				}
+			}
+
+			logs, err := p.openshiftClient.GetPodLogs(ctx, pod.Name, pod.Namespace)
+			if err != nil {
+				logger.Info(fmt.Sprintf("warning: unable to retrieve pod logs: %s", err))
+				continue
+			}
+
+			for container, logContents := range logs {
+				filename := fmt.Sprintf("%s-%s.log", pod.Name, container)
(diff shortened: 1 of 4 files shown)


internal/openshift/openshift.go
```diff
@@ -519,3 +525,108 @@ func (oe *openshiftClient) GetDeployment(ctx context.Context, name string, names
...
+	_, err = io.Copy(buf, logs)
+	if err != nil {
+		return nil, fmt.Errorf("failed to copy log stream: %s/%s/%s: %v", namespace, pod, container, err)
+	}
+
+	return buf, nil
+}
+
+// GetPodLogs can return an ErrNotFound
+func (oe *openshiftClient) GetPodLogs(ctx context.Context, name string, namespace string) (map[string]*bytes.Buffer, error) {
```
acornett21, line 594: This doesn't seem to be used, are we just adding it for completeness?

internal/policy/operator/deployable_by_olm.go
```diff
@@ -647,6 +657,41 @@ func (p *DeployableByOlmCheck) cleanUp(ctx context.Context, operatorData operato
...
+			}
+
+			logs, err := p.openshiftClient.GetPodLogs(ctx, pod.Name, pod.Namespace)
+			if err != nil {
+				logger.Info(fmt.Sprintf("warning: unable to retrieve pod logs: %s", err))
+				continue
+			}
+
+			for container, logContents := range logs {
+				filename := fmt.Sprintf("podLogs-%s-%s.json", pod.Name, container)
```
acornett21, line 687: Do we know what this looks like? Does the `json` make sense?

internal/openshift/openshift.go
caxu-rh: `GetPodLogs` is used [over here](https://github.com/redhat-openshift-ecosystem/openshift-preflight/pull/1314/files#diff-07f6ce8cd49e3852e3165e0cd817d81a80b6d1878f214ac914a1887545cd8e1dR680) (line 680), if that's what you're asking.

internal/policy/operator/deployable_by_olm.go
acornett21: I'd go with `.log`, since we know that can be stored in pyxis properly.

internal/openshift/openshift.go
caxu-rh: Yeah, I kept that public for completeness/consistency with existing functions; can make it private if preferred (or absorb the functionality into `GetPodLogs`).

internal/openshift/openshift.go
```diff
@@ -519,3 +525,109 @@ func (oe *openshiftClient) GetDeployment(ctx context.Context, name string, names
...
+		return nil, fmt.Errorf("could not retrieve pod: %s/%s: %v", namespace, name, err)
+	}
+	return &pod, nil
+}
+
+func (oe *openshiftClient) getContainerLogs(ctx context.Context, namespace string, pod string, container string) (*bytes.Buffer, error) {
+	podLogOpts := corev1.PodLogOptions{
+		Container: container,
+	}
+	req := oe.K8sInterface.CoreV1().Pods(namespace).GetLogs(pod, &podLogOpts)
```
bcrochet: Are we pulling this in just to get a typed interface? It should be possible to get what you need without introducing this new field.

https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/client#example-Client-Get

PR #1316, opened 2025-09-25
Add new flag to use non-zero exit code when checks fail

caxu-rh:
Open to feedback on how to structure/propagate the boolean flag. This approach introduces dependency on `viper` in `cli`; alternatively, we could add the flag into the `CheckConfig` struct (but this would require changing how `CheckConfig` structs are initialized/created in several places). Similarly, the flag could also be added as an additional parameter in `RunPreflight`, but this would require changing places which expect the existing function signature.

### DIFF
--- a/internal/cli/cli.go
+++ b/internal/cli/cli.go
@@ -14,6 +14,7 @@ import (
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/internal/formatters"
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/internal/lib"
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/internal/log"
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/internal/viper"
 
 	"github.com/go-logr/logr"
 )
@@ -83,6 +84,10 @@ func RunPreflight(
 	}
 
 	logger.Info(fmt.Sprintf("Preflight result: %s", convertPassedOverall(results.PassedOverall)))
+
+	if !results.PassedOverall && viper.Instance().GetBool("exit_with_failure") {
+		return fmt.Errorf("one or more checks did not pass")
+	}
 
 	return nil
 }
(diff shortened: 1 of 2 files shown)


internal/cli/cli.go
```diff
@@ -84,6 +85,10 @@ func RunPreflight(
 
 	logger.Info(fmt.Sprintf("Preflight result: %s", convertPassedOverall(results.PassedOverall)))
 
+	if !results.PassedOverall && viper.Instance().GetBool("exit_with_failure") {
+		return fmt.Errorf("one or more checks did not pass")
```
bcrochet, line 89: What about this... How about we just return an error with results.PassedOverall and leave the viper out of this.

Check the error and the bool further up without introducing viper here.

bcrochet (changes requested):
Alternative...

internal/cli/cli.go
acornett21: I think the intent here is so that we don't break peoples existing CI that depend on our error codes, and we make this an opt in feature via a flag.

bcrochet: I'm not suggesting we get rid of the flag... I'm suggesting we don't push the flag down into the cli package. 

For instance... The cli package would now have a `cli.DidNotPassOverallError`. (Please don't use that name)

```
if !results.PassedOverall {
  return cli.DidNotPassOverallError
}
```

_Meanwhile, somewhere in the cmd tree_

```
if errors.Is(err, cli.DidNotPassOverallError) && viper.Instance().GetBool("exit_with_failure") {
  return err
}

// Check for other errors as usual
```

This keeps the existence of this flag from escaping cmd.

bcrochet (changes requested):
Great direction, but another suggestion.

acornett21 (approved):

cmd/preflight/main.go
```diff
@@ -1,13 +1,23 @@
...
-		log.Fatal(err)
+		if errors.Is(err, &cmd.ChecksErroredError{}) {
+			log.Println(err)
+			os.Exit(1)
+		} else if errors.Is(err, &cmd.ChecksFailedError{}) {
+			log.Println(err)
+			os.Exit(2)
+		} else {
+			log.Fatal(err)
+		}
```
bcrochet:
```suggestion
		if errors.Is(err, &cmd.ChecksErroredError{}) {
			log.Println(err)
			os.Exit(1)
		}
		if errors.Is(err, &cmd.ChecksFailedError{}) {
			log.Println(err)
			os.Exit(2)
		}
		
		// Default, original behaviour
		log.Fatal(err)
```

PR #1318, opened 2025-10-24
removing the deprecated certification-project-id flag

acornett21:
## Motivation
The flag has been deprecated from sometime now, and it was found out that we were overriding precedences of `flags`, `env`, and `configs` with taking `envs` over `flags`. Since there is a bug in the custom logic, that goes against cobra's order, it makes sense to just remove this flag.

## Changes

- Removes the `certification-project-id` flag
- Update supporting vars to reference `component` flag 

Build-depends: https://github.com/redhatci/ansible-collection-redhatci-ocp/pull/810

### DIFF
--- a/cmd/preflight/cmd/check_container.go
+++ b/cmd/preflight/cmd/check_container.go
@@ -34,10 +34,7 @@ import (
 	"github.com/spf13/pflag"
 )
 
-var (
-	submit      bool
-	componentID string
-)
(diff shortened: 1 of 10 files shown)


bcrochet (changes requested):
Comment removal

bcrochet (approved):
/lgtm

PR #1322, opened 2025-10-31
Restrict permissions in go build workflow

caxu-rh:
Now that we have `zizmor` audits in place, we can proceed with further securing workflows by limiting the `GITHUB_TOKEN` only to permissions which are needed.

This PR starts with the `go.yml` workflow which is easy to validate (since it ran on even this PR).

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -6,10 +6,10 @@ on:
     - main
     - release-*
 
+permissions: {}
+
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


caxu-rh:
Validated via [this workflow run](https://github.com/redhat-openshift-ecosystem/openshift-preflight/actions/runs/18981855794) that none of the jobs in `go.yml` require additional permissions.

acornett21 (approved):

PR #1330, opened 2025-11-21
Reverse an order of layers for Pyxis

Allda:
Pyxis schema expects image layers to be sorted from base image to top layer. This commit reverts the order to align with Pyxis expected schema.

JIRA: ISV-6472


acornett21:
@Allda As part of this, once this is released would you want all previous versions of preflight invalidated in pyxis, so you don't have to do data migration multiple times?

PR #1337, opened 2025-12-02
Avoid using third-party action for finding latest tag

caxu-rh:
The [logic in this third-party action](https://github.com/reloc8/action-latest-release-version/blob/master/action.yml) is simple enough that it's probably just easier to include the logic in the workflow, avoiding a third-party dependency and also simplifying things a little bit.

Also increase Dependabot cooldown to 7 days (small fix to pass lint after zizmor audit rule change; happy to move this into another PR first, but it's necessary for CI here to pass).

### DIFF
--- a/.github/workflows/build-main.yml
+++ b/.github/workflows/build-main.yml
@@ -21,16 +21,12 @@ jobs:
     - uses: actions/checkout@1af3b93b6815bc44a9784bd300feb67ff0d1eeb3 # v6.0.0
       with:
         persist-credentials: false
-    - name: Fetch latest release version
-      uses: reloc8/action-latest-release-version@b8d6337f30390558e7874a044d6a3c1314314bab # 1.0.0
-      id: fetch-latest-release
 
     - name: Set release tag and short SHA
       run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/build-main.yml
```diff
@@ -21,16 +21,12 @@ jobs:
...
         persist-credentials: false
-    - name: Fetch latest release version
-      uses: reloc8/action-latest-release-version@b8d6337f30390558e7874a044d6a3c1314314bab # 1.0.0
-      id: fetch-latest-release
 
     - name: Set release tag and short SHA
       run: |
-        echo "RELEASE_TAG=${RELEASE_TAG}" >> "${GITHUB_ENV}"
+        git fetch --tags
+        echo "RELEASE_TAG=$(git describe --abbrev=0 --tags)" >> "${GITHUB_ENV}"
```
acornett21: It seems smart to get ride of this since it hasn't been updated in 4 years, do you think we'd need the below line?
```
git fetch --prune --unshallow || true
```
Or not needed in our case since we have a fresh clone here, and the above would always just be `true`?

acornett21: Just wanted to make sure we were thinking the same thing.

PR #1348, opened 2026-01-09
Optimize image layer extraction to reduce memory usage

bcrochet:
## Summary

This PR optimizes the container image layer extraction process to significantly reduce memory usage and improve extraction speed. The changes replace `mutate.Extract()` with a custom layer-by-layer extraction approach that properly manages resources.

### Core Optimization
- Replaced `mutate.Extract()` with custom `extractImageLayers()` function
- Processes layers individually instead of flattening all layers first
- Explicitly closes each layer reader immediately after processing
- Prevents memory accumulation from deferred closes in loops

### Code Refactoring
- Enhanced `untar()` function to support Docker overlay filesystem semantics
- Added whiteout handling (`.wh.` prefix detection) for proper layer overlay
- Consolidated duplicate tar extraction logic into single reusable function
- Removed ~140 lines of duplicate code

## Performance Improvements

Tested with `quay.io/redhat-services-prod/app-sre-tenant/qontract-reconcile-master/qontract-reconcile-master:latest`:

### Memory Usage
- **Main branch**: 6MB → 14MB (8MB growth during extraction)
- **Optimized**: 8MB → 10MB (2MB growth during extraction)
- **Result: 75% reduction in memory growth**

### Extraction Speed
- **Main branch**: 23 seconds
- **Optimized**: 15 seconds
- **Result: 35% faster extraction**

### Layer-by-Layer Memory Release
The optimization properly releases memory after each layer:
```
Layer 6: 8 MB
Layer 5: 9 MB  
Layer 4: 8 MB (dropped back down)
Layer 3: 9 MB
Layer 2: 18 MB (spike during processing)
Layer 1: 10 MB (released again)
```

## Technical Details

The key insight is that `mutate.Extract()` returns a single reader for all flattened layers, and when iterating through layers to get readers, deferred `Close()` calls accumulate until the entire function completes. This optimization:

1. Gets each layer individually from the image
2. Opens the layer reader
3. Extracts files using `untar()` with whiteout/overlay support
4. **Immediately closes the layer reader** before moving to the next layer
5. Allows garbage collector to reclaim memory between layers

The refactored `untar()` function now handles both traditional tar extraction (when `fileMap=nil`) and Docker layer semantics (when `fileMap` is provided), making it reusable and maintainable.

## Testing

- Built and tested with the same image used in performance measurements
- All checks pass successfully
- Pre-commit hooks pass
- Memory instrumentation added for validation (trace-level logging)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/internal/engine/engine.go
+++ b/internal/engine/engine.go
@@ -15,6 +15,7 @@ import (
 	"path"
 	"path/filepath"
 	"regexp"
+	goruntime "runtime"
 	"slices"
 	"sort"
 	"strings"
@@ -39,8 +40,8 @@ import (
 
 	"github.com/google/go-containerregistry/pkg/crane"
 	"github.com/google/go-containerregistry/pkg/name"
+	v1 "github.com/google/go-containerregistry/pkg/v1"
 	"github.com/google/go-containerregistry/pkg/v1/cache"
-	"github.com/google/go-containerregistry/pkg/v1/mutate"
 )
 
 // New creates a new CraneEngine from the passed params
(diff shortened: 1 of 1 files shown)


caxu-rh (commented):
A couple of thoughts:
- Is there some performance comparisons available for images with more layers and in a more constrained execution environment (possibly suggest GC to try and trim more often with `GOMEMLIMIT=...`)? The deferred-close memory savings would be more pronounced under these conditions.
- There's a lot of what _appears_ to be similar code introduced here and already existing in `internal/policy/container/has_modified_files.go`. It may be worth considering unifying them into a common implementation (in a new place under `internal`?) if possible. (The code in both of these places are both already heavily copied from the implementation in `go-containerregistry`).
- In lieu of trying to work around the issue of deferred closes happening inside loops, it may be better to try and get this fixed upstream, and continue being able to use the `mutate` package for this. I opened https://github.com/google/go-containerregistry/pull/2190 to try and fix this for `mutate.Extract`, regardless of how we choose to proceed here.

PR #1363, opened 2026-02-03
engine: Clean dirty page cache every 100 written MBs

jhutar:
Sometimes in the OpenShift deployment container with Preflight gets killed by OOMKiller. Based on journalctl log from this we think it is caused by dirty (not written to the disk) page cache. This happens when the disk IO can not keep up with speed of writing untared files to disk and pagecache fills whole memory limit of a container.

Generated-by: Gemini


jhutar:
Yes, I agree this is not sufficient! Closing for now as I think there is better plan in works.

PR #1364, opened 2026-02-06
engine: support filter/allowlist for untar

caxu-rh:
Pardon the big diff, fortunately some of the changes are just moving a lot of things around and additional tests/test helpers. High level summary of changes:

- Extend the `Check` interface to add `RequiredFilePatterns() []string`. The changes in `internal/policy` are pretty much just implementing this new function in each check.
- Move the `untar` and `resolveLinkPaths` functions out of `engine.go` into `untar.go`.
  - `untar` now calls `untarOnce`, a new function added in `untar.go`
  - Tests for link path resolution and untar directory traversal protection have been moved from `engine_test.go` to `untar_test.go`. The test helper function `writeTarballWithLink` is also moved to `untar_test.go`, since it is only used there now.
- Hoist array of rpmdb paths up into a public variable in `internal/rpm/rpm.go`. This allows `HasProhibitedPackagesCheck` to refer to it for its list of required file patterns (rather than keeping it up to date in multiple places).
- Create a list of file patterns common to operator bundles in `internal/bundle/bundle.go`. All operator checks that rely on loading bundles from directory refer to this list as their required file patterns.

Changes in the engine/untar process:
- Collect/deduplicate required file patterns from all checks (and convert into relative paths)
- `untar` now accepts the context, destination temporary directory, the **`v1.Image`** (not an `io.Reader` anymore), and a **list of file patterns (string slice)**. `untar` then calls `untarOnce` one or more times.
  - `untarOnce` calls `mutate.Extract` and opens an `io.Reader`. `untarOnce` iterates through the tar stream, similar to the previous implementation of `untar`.
    - tar entries that do not `filepath.Match` to any pattern in the list of file patterns are skipped
    - Directory-type entries are skipped. We already have code that makes the required directory structure for any regular files/links anyways.
    - When encountering a hardlink/symlink entry, the target of the link might not be covered in the list of file patterns. We create the symlink.
      - If the target file has already been created (on this `untarOnce` invocation, or an earlier one - tracked via the `state` map), no further action is needed.
      - Otherwise, add the target file to a list of link targets encountered in this pass, and also to the list of file patterns to look for.
  - After iterating through all tar entries, if there are any link targets encountered in this pass that were not found later in the pass (i.e. the link occurs after the target in the tar archive), we return those from `untarOnce` to `untar`. It will take another pass to find them.
- `untar` keeps running `untarOnce` as long as there are still link targets that need to be resolved.

### DIFF
--- a/internal/engine/engine.go
+++ b/internal/engine/engine.go
@@ -1,12 +1,10 @@
 package engine
 
 import (
-	"archive/tar"
 	"bytes"
 	"context"
 	"crypto/md5"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"io"
 	"io/fs"
@@ -15,7 +13,6 @@ import (
 	"os"
 	"os/exec"
 	"path"
-	"path/filepath"
 	"regexp"
 	"slices"
 	"strings"
@@ -41,7 +38,6 @@ import (
 	"github.com/google/go-containerregistry/pkg/crane"
 	"github.com/google/go-containerregistry/pkg/name"
 	"github.com/google/go-containerregistry/pkg/v1/cache"
-	"github.com/google/go-containerregistry/pkg/v1/mutate"
 )
 
 // New creates a new CraneEngine from the passed params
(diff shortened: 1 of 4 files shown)


internal/engine/engine.go
```diff
@@ -147,26 +143,18 @@ func (c *craneEngine) ExecuteChecks(ctx context.Context) error {
...
 	}
 
-	// Wrap this critical section in a closure to that we can close the
-	// mutate.Extract reader sooner than the end of the checks
-	if err := func() error {
-		// export/flatten, and extract
-		logger.V(log.DBG).Info("exporting and flattening image")
-		fs := mutate.Extract(img)
-		defer fs.Close()
+	requiredFilePatterns := []string{}
```
bcrochet: Let's have a bit of a guess here, and `make([]string, 0, SWAG)` to pre-allocate. Or even, loop over the checks and get the sizes first, then `make`.

bcrochet (changes requested):
Just a couple of comments/questions.

Also, I think it might be nice (but up to you) to make 2 commits here? The first being the one that adds the fields to the checks, then moves the untar around?

Would make future reviews easier. WDYT?

internal/engine/untar.go
```diff
@@ -0,0 +1,243 @@
...
+		switch header.Typeflag {
+		// skip all directories, we'll only create the needed directory
+		// structure for the files/symlinks that need to be created
+		case tar.TypeDir:
+			continue
+
+		// if it's a file create it
+		case tar.TypeReg:
+			dirname := filepath.Dir(header.Name)
+			if _, err := dstRoot.Stat(dirname); err != nil {
```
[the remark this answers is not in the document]
caxu-rh: Good catch (this was copied over directly from the existing `untar` implementation). Removing the outer `Stat` call but keeping the error checking on the `MkdirAll`, to keep the error message clear in case it fails for some other reason.

internal/bundle/bundle.go
```diff
@@ -40,6 +40,12 @@ var ocpToKubeVersion = map[string]string{
 
 const latestReleasedVersion = "4.21"
 
+var BundleFiles = []string{
+	"/manifests/*",
+	"/metadata/annotations.yaml",
```
acornett21: `/metadata` can have more then just this file, but if we are only listing the files that we read/need this should be fine.

bcrochet (changes requested):
Note about `MkdirAll`

PR #1377, opened 2026-02-26
engine: support more robust handling of links in tar stream

caxu-rh:
# Problem(s)

This PR addresses several corner cases in the handling of hardlinks/symlinks in the tar stream:
1. A tar entry may be a hardlink to a symlink to a location on disk. The symlink may be a relative path, which, if evaluated from the location of the hardlink, resolves successfully. We currently rewrite these relative paths into absolute paths during extraction, but this relative path to absolute path conversion is performed from the location of the symlink, thus breaking the hardlink.
2. Hardlinks can only be created when the target file exists. Since we are now filtering/being selective about which files we extract, it's possible that a hardlink is eligible for extraction for which we have skipped over the link target. Unlike the symlink case, we cannot just create the link and then come back for the target; we must extract the link target and then try to create the link.
3. A file might not match the filters for extraction via its canonical path, but one of its parent directories may be a symlink which, if traversed through that path, _does_ match a filter for extraction. For example, `rpmdb.sqlite` may be canonically at `/usr/share/rpm/rpmdb.sqlite`, with a symlink `/usr/lib/sysimage/rpm -> ../../share/rpm`, thereby making the file available at `/usr/lib/sysimage/rpm/rpmdb.sqlite`.

Problems 2 and 3 were introduced with #1364, problem 1 has been present since #1179.

# Existing process

Walk through the tar stream; extract regular file entries, symlinks, and hardlinks which match the filters. In the case where the symlink/hardlink points to a target that is not extracted yet, make additional passes through the tar stream as needed to resolve and extract the target (and any intermediary symlinks), repeating until there are no outstanding files to search for.

# New process

Walk through the tar stream once, collect a list (`allFiles`) of the names of all regular files, symlinks, and hardlinks, and also build a graph of all the symlinks and hardlinks.

Using the symlink/hardlink graph, build a map that shows, for a given link, what are the other paths by which that location is reachable (basically, backward traversal of the graph) - this occurs in `BuildDirectoryAliasMap`. In the example described above, there would be an entry in the map for `/usr/share/rpm -> /usr/lib/sysimage/rpm`, since you can arrive at `/usr/share/rpm` by traversing the `/usr/lib/sysimage/rpm` symlink.

Then, take all files in the tar stream (collected in `allFiles`) and determine all paths by which that location is reachable (this occurs in `ExpandFilePathAliases`). For example, we provide input `/usr/share/rpm/rpmdb.sqlite`, and it provides output `["/usr/share/rpm/rpmdb.sqlite", "/usr/lib/sysimage/rpm/rpmdb.sqlite"]` - and if any of these "aliases" match the filter pattern, we'll want to extract the original file and any links needed to be able to traverse to the file at the location.

Next, take a look at any files we want to extract - if they are symlinks or hardlinks, traverse them forward in the graph to find any additional links/files that we also need to extract.

Finally, walk through the tar stream again and extract the exact list of files we need. When encountering hardlinks, if the target already exists, we can create it, otherwise save it for later.

After traversing the tar stream the second time, we may still have some hardlinks that need to be created. Sort them topologically to create the remaining hardlinks in the appropriate order.

# Performance

I built a container locally and tried running it on an OpenShift cluster. Runtime is about the same as before, if not a little faster. Memory usage during the image extraction phase looks about the same (probably a few MiB more to store the maps/graphs described above).

# How do other tools handle this?

- `oc image extract` allows filtering by paths but doesn't handle links (tries to link to non-existent files, does not resolve and extract link targets, etc.)
- `crane export` does not allow filtering by paths
- `tar` allows for the extraction of specific files in the archive but will not resolve "virtual paths" that only exist when traversing links

# Alternatives

- Ack and note the limitations and adjust filter paths as necessary when we need to. This may require staying on top of changes in Fedora/UBI.
- Solve only problem 2 mentioned above; problem 1 and 3 require most of the refactor here while problem 2 alone might be able to be solved more simply.

# Todo

- More unit tests for `graph.go`. Getting this up early to get some feedback before investing more time.

### DIFF
--- a/internal/engine/engine.go
+++ b/internal/engine/engine.go
@@ -357,7 +357,7 @@ func generateBundleHash(ctx context.Context, bundlePath string) (string, error) 
 	slices.Sort(keys)
 
 	for _, k := range keys {
-		hashBuffer.WriteString(fmt.Sprintf("%s  %s\n", k, files[k]))
+		fmt.Fprintf(&hashBuffer, "%s  %s\n", k, files[k])
 	}
 
 	artifactsWriter := artifacts.WriterFromContext(ctx)
--- a/internal/engine/graph.go
+++ b/internal/engine/graph.go
@@ -1,0 +1,404 @@
+package engine
+
+import (
+	"archive/tar"
+	"path/filepath"
+	"slices"
+	"strings"
+
+	"github.com/go-logr/logr"
+
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/internal/log"
+)
+
+// linkType a convenience type just to make the consuming functions more clear.
+type linkType byte
+
+const (
+	hardlink linkType = tar.TypeLink
+	symlink  linkType = tar.TypeSymlink
+)
+
+// String returns the string representation of the LinkType.
+func (lt linkType) String() string {
+	if lt == tar.TypeLink {
+		return "hardlink"
+	}
+	return "symlink"
+}
+
+type linkNode struct {
+	Name              string
+	Deps              *linkNode
+	OriginalLinkname  string   // For symlinks, the original target string before resolution
+	Type              linkType // Type of link (symlink or hardlink)
+	VirtualLinkTarget string   // For hardlinks pointing to symlinks, the symlink's target string
+}
+
+func (n *linkNode) IsHardlink() bool {
+	return n.Type == hardlink
+}
+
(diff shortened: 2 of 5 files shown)


bcrochet:
Please rebase so the Gemini reviews can run.

caxu-rh:
So far we seem to be getting by without covering these corner cases. We can re-visit this in the future if needed.

Issue #1411, opened 2026-04-07
Add tests for scorecard check Help() fatal error branches


bcrochet:
## Summary

Both `ScorecardBasicSpecCheck.Help()` and `ScorecardOlmSuiteCheck.Help()` have a `fatalError` branch annotated with `//coverage:ignore`.

## Affected Files

| File | Line |
|------|------|
| `internal/policy/operator/scorecard_basic_check_spec.go` | 68 |
| `internal/policy/operator/scorecard_olm_suite.go` | 68 |

Both follow the same pattern:
```go
func (p *ScorecardXxxCheck) Help() check.HelpText {
    if p.fatalError {
        //coverage:ignore
        return check.HelpText{
            Message: "There was a fatal error...",
            Suggestion: "If the logs are showing a context timeout...",
        }
    }
    return check.HelpText{...}
}
```

## Requested Changes

1. Delete `//coverage:ignore` from line 68 in both files
2. Add tests that construct each check with `fatalError: true` and assert the `Help()` return contains the fatal error message and suggestion text

bcrochet:
**CodeRabbit AI Agent Prompts** (from PR #1400):

### `internal/policy/operator/scorecard_basic_check_spec.go` (lines 68-73)
```
Verify each finding against the current code and only fix it if needed.

In `@internal/policy/operator/scorecard_basic_check_spec.go` around lines 68 - 73,
Remove the //coverage:ignore that hides the fatal branch in Help() and the
similar ignore around RequiredFilePatterns(); instead make these code paths
testable by keeping their logic visible and add unit tests that exercise the
fatal Help() return (the error path in Help()) and the RequiredFilePatterns()
function so coverage captures them; locate the Help() method and the
RequiredFilePatterns() function in scorecard_basic_check_spec.go and ensure
tests assert the Help() fatal Message/Suggestion content and the expected file
pattern list.
```

### `internal/policy/operator/scorecard_olm_suite.go` (lines 68-73)
```
Verify each finding against the current code and only fix it if needed.

In `@internal/policy/operator/scorecard_olm_suite.go` around lines 68 - 73,
Remove the "coverage:ignore" pragmas on the user-facing branches in
scorecard_olm_suite.go (specifically the Help() return block and the
RequiredFilePatterns() return paths) so these code paths are included in
coverage; then add unit tests that exercise the Help() method and
RequiredFilePatterns() function to assert the returned check.HelpText.Message
[...]

PR #1434, opened 2026-04-15
check: add tests for scorecard check Help() fatal error branches

bcrochet:
Remove `//coverage:ignore` from fatal error branches in `Help()` for both `ScorecardBasicSpecCheck` and `ScorecardOlmSuiteCheck`, and add tests asserting the fatal help text.

Refs: #1411


acornett21 (approved):
/lgtm

Issue #1412, opened 2026-04-07
Add tests for certificationInputBuilder.finalize() nil-guard validation


bcrochet:
## Summary

`internal/pyxis/builder.go` has `//coverage:ignore` on two nil-guard checks in `finalize()`:

- Line 50: `if b.CertImage == nil` → returns `"a CertImage was not provided and is required"`
- Line 54: `if b.TestResults == nil` → returns `"test results were not provided and are required"`

## Requested Changes

1. Delete `//coverage:ignore` from lines 50 and 54
2. Add tests in `builder_test.go`:
   - Construct a builder with `CertImage = nil`, call `finalize()`, assert the CertImage error
   - Construct a builder with `TestResults = nil`, call `finalize()`, assert the TestResults error

## Context

`builder_test.go` already tests the builder pattern with `WithCertImage()`, `WithPreflightResults()`, etc. The nil-guard tests just need to omit those builder calls before calling `finalize()`.

bcrochet:
**CodeRabbit AI Agent Prompt** (from PR #1400):

```
Verify each finding against the current code and only fix it if needed.

In `@internal/pyxis/builder.go` around lines 50 - 55, Remove the //coverage:ignore
comments on the CertImage and TestResults validation guards so those branches
are covered by tests: locate the method that checks b.CertImage == nil and
b.TestResults == nil (the blocks that return fmt.Errorf("a CertImage was not
provided and is required") and fmt.Errorf("test results were not provided and
are required")), delete the coverage ignore annotations, and add unit tests that
construct the builder with b.CertImage == nil and with b.TestResults == nil to
assert the function returns the respective errors.
```

bcrochet:
Resolved via merged PR.

PR #1435, opened 2026-04-15
pyxis: add tests for certificationInputBuilder.finalize() nil-guard validation

bcrochet:
Remove `//coverage:ignore` from nil-guard checks in `finalize()` and add tests for nil `CertImage` and nil `TestResults` cases.

Refs: #1412


acornett21 (approved):
/lgtm

Issue #1413, opened 2026-04-07
Add test for ResultWriterFile.OpenFile error path


bcrochet:
## Summary

`internal/runtime/result_writer.go` line 20 has `//coverage:ignore` on the error return from `os.OpenFile`.

## Affected Code

```go
func (f *ResultWriterFile) OpenFile(name string) (io.WriteCloser, error) {
    file, err := os.OpenFile(name, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0o600)
    if err != nil {
        //coverage:ignore
        return nil, err
    }
    ...
}
```

## Requested Changes

1. Delete `//coverage:ignore` from line 20
2. Add a test in `result_writer_test.go` that triggers the `OpenFile` failure — for example, pass a path to a non-existent directory (e.g., `/nonexistent/dir/file.txt`) and assert the error is returned

## Context

`result_writer_test.go` already uses `os.MkdirTemp()` for file I/O tests. The error test just needs an invalid path.

bcrochet:
**CodeRabbit AI Agent Prompt** (from PR #1400):

```
Verify each finding against the current code and only fix it if needed.

In `@internal/runtime/result_writer.go` around lines 20 - 21, Remove the
"//coverage:ignore" suppression before the error return in
internal/runtime/result_writer.go so the OpenFile error branch is counted by
coverage; keep the existing "return nil, err" but delete the coverage comment
and add a unit test that triggers the OpenFile failure (e.g., by using a temp
path with restricted permissions or a filesystem mock) to assert the function
(the code path returning nil, err) behaves correctly.
```

bcrochet:
Resolved via merged PR.

PR #1436, opened 2026-04-15
runtime: add test for ResultWriterFile.OpenFile error path

bcrochet:
Remove `//coverage:ignore` from the `OpenFile` error path in `result_writer.go` and add a test triggering the error with a non-existent directory.

Refs: #1413


acornett21 (approved):
/lgtm

Issue #1414, opened 2026-04-07
Remove coverage exclusions from openshiftClient error-return paths and verify testability with fake clients


bcrochet:
## Summary

`internal/openshift/openshift.go` has `//coverage:ignore` on error-return branches in namespace operations that should be testable with the existing `controller-runtime` fake client infrastructure.

## Key Affected Lines

| Line | Method | Error |
|------|--------|-------|
| 78 | `CreateNamespace` | `"could not create namespace: %s: %v"` |
| 95 | `DeleteNamespace` | `"could not delete namespace: %s: %v"` |

(Plus additional annotations throughout the file on other methods)

## Requested Changes

1. Delete `//coverage:ignore` from error-return paths that can be exercised with fake clients
2. Evaluate each annotation: the fake client from `controller-runtime` can be configured to return errors via interceptors or reactor functions
3. For paths that truly require a live cluster (e.g., specific API server behaviors), keep `//coverage:ignore` but add a comment explaining why
4. Add tests using `fakecr.NewClientBuilder()` with error-returning interceptors for the removable paths

## Context

`openshift_test.go` already uses `fakecr.NewClientBuilder()` and `fakecg.NewClientset()`. The `controller-runtime` fake client supports `WithInterceptorFuncs()` to inject errors on specific operations.

bcrochet:
**CodeRabbit AI Agent Prompt** (from PR #1400):

```
Verify each finding against the current code and only fix it if needed.

In `@internal/openshift/openshift.go` around lines 77 - 79, Remove the broad
test-coverage suppressions on core error paths by deleting the inline
"//coverage:ignore" comments around the namespace creation error return (the
branch that returns "nil, fmt.Errorf(\"could not create namespace: %s: %v\",
name, err)"), and similarly remove or narrow the ignores at the other cited
locations (around lines handling errors in the same file) so these branches can
be exercised by unit tests using mocked/fake Kubernetes clients; ensure only
non-testable, platform-specific code retains any coverage skips and leave the
actual error-return statements (fmt.Errorf returns) intact.
```

bcrochet:
Resolved via merged PR.

PR #1437, opened 2026-04-15
openshift: remove coverage exclusions from openshiftClient error-return paths

bcrochet:
Remove 34 `//coverage:ignore` annotations from `openshiftClient` error-return paths and add 573 lines of unit tests using fake Kubernetes clients to cover all error branches.

Refs: #1414


acornett21 (approved):
/lgtm

PR #1442, opened 2026-04-15
pyxis: reduce per-line coverage exclusions in createArtifact

bcrochet:
Remove 7 `//coverage:ignore` annotations from happy-path lines in `createArtifact` (logger init, defer, variable declarations, return). Keep annotations only on the 6 error-return branches. Verified 100% coverage maintained.

Refs: #1419


acornett21 (approved):
/lgtm

Issue #1420, opened 2026-04-07
Remove coverage exclusion from artifact ImageID assignment in submit.go


bcrochet:
## Summary

`internal/pyxis/submit.go` line 112 has `//coverage:ignore` on a simple assignment inside the artifacts loop:

```go
for _, artifact := range artifacts {
    //coverage:ignore
    artifact.ImageID = certImage.ID
```

This is a normal execution path, not an error branch.

## Requested Changes

1. Delete `//coverage:ignore` from line 112 (the assignment)
2. Keep `//coverage:ignore` on line 115 (the `createArtifact` error return) if desired — that involves Pyxis API calls
3. No new test needed if integration tests already iterate over artifacts; otherwise add a test verifying `ImageID` is set correctly on each artifact before `createArtifact` is called

bcrochet:
**CodeRabbit AI Agent Prompt** (from PR #1400):

```
Verify each finding against the current code and only fix it if needed.

In `@internal/pyxis/submit.go` around lines 111 - 118, Remove the stray coverage
exclusion placed before the assignment to artifact.ImageID inside the artifacts
loop; specifically, delete the "//coverage:ignore" that precedes the line
setting artifact.ImageID = certImage.ID so that the assignment and loop behavior
are testable, and keep or re-add "//coverage:ignore" only on the error branch
that wraps the call to p.createArtifact(ctx, &artifact) (or directly around the
fmt.Errorf return) if you need to exclude the HTTP-dependent error path; ensure
the loop over artifacts and the assignment to ImageID are covered by tests that
exercise the artifact-handling logic.
```

bcrochet:
Further investigation is required for this one.

PR #1443, opened 2026-04-15
pyxis: remove coverage exclusion from artifact ImageID assignment in submit.go

bcrochet:
Remove stray `//coverage:ignore` from `artifact.ImageID = certImage.ID` assignment. The line is already exercised by existing tests. Verified 100% coverage maintained.

Refs: #1420


bcrochet:
Closing this PR. The `//coverage:ignore` on `artifact.ImageID = certImage.ID` cannot be removed because the entire `SubmitResults` function is integration-only (makes real Pyxis API calls) and has no unit test coverage. Removing the annotation drops coverage below 100% because the line becomes visible to the coverage tool but is never executed in tests.

To properly fix this, a full `SubmitResults` integration test with mocked Pyxis endpoints would be needed, which is beyond the scope of this issue.

PR #1445, opened 2026-04-15
container: narrow coverage exclusions in Pyxis resolution flow

bcrochet:
Narrow `//coverage:ignore` placement in `check_container.go` Pyxis resolution flow. Remove 6 annotations from non-error lines (API calls, logger, conditionals), keeping annotations only on hard-to-test error returns. Verified 100% coverage maintained.

Refs: #1422


acornett21 (approved):
/lgtm

PR #1449, opened 2026-04-17
option: add tests for GenerateCraneOptions and remove coverage exclusions

bcrochet:
## Summary

- Add tests for `GenerateCraneOptions` covering both secure and insecure code paths using a fake `CraneConfig` implementation, bringing the `internal/option` package from 25% to 100% statement coverage.
- Remove the two `//coverage:ignore` directives in `option.go` that are no longer needed now that tests exercise all paths.


acornett21 (approved):
/lgtm

PR #1453, opened 2026-04-21
engine: fix globbing inside directory symlinks

caxu-rh:
## Context
Checks only pull matching paths out of the image tar (e.g. `licenses/**`). In some images, /licenses is a symlink to /usr/share/licenses, but the tar still lists files as usr/share/licenses/..., not licenses/..., so `licenses/**` never matched those files.

When the extractor followed the symlink, it only added the bare target path (usr/share/licenses) as the next filter. Glob matching treats that as that path only, not everything under it, so usr/share/licenses/pkg/LICENSE did not match and nested license files were never extracted (including on a second pass that still only had the bare path).

## Changes
- After resolving a symlink target, add both usr/share/licenses and `usr/share/licenses/**` so descendants in the tar match.
- At the start of each untarOnce pass, for patterns without glob characters, automatically add `path/**` next to each literal path. That way a follow-up pass whose list is only usr/share/licenses still expands to include `usr/share/licenses/**`.
- When clearing “pending symlink targets”, treat a target like usr/share/licenses as satisfied once any extracted path is that path or under it (longest-prefix match), not only an exact name hit.
- Test cases for the above

### DIFF
--- a/internal/engine/untar.go
+++ b/internal/engine/untar.go
@@ -20,6 +20,60 @@ import (
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/internal/log"
 )
 
+// expandLiteralPatternsWithDescendantGlob adds a "path/**" pattern for each pattern that has no
+// glob metacharacters, so a follow-up pass that only lists a symlink's target directory (e.g.
+// usr/share/licenses) still matches nested files in the tarball.
+//
+// Patterns containing `[` are skipped: we treat * ? [ as glob syntax; literal `[` in a path is rare.
+func expandLiteralPatternsWithDescendantGlob(patterns []string) []string {
+	if len(patterns) == 0 {
+		return patterns
+	}
+	out := slices.Clone(patterns)
+	for _, p := range patterns {
+		if p == "" || strings.ContainsAny(p, "*?[") {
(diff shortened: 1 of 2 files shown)


caxu-rh:
Test images:

`quay.io/rh-ee-caxu/test-licenses-symlink:latest` - copies the scenario described in the original issue, symlinking `/licenses -> /usr/share/licenses`. Note that there are no licenses in `/usr/share/licenses` itself, but rather this directory contains subdirectories which contain licenses

```
FROM registry.redhat.io/ubi10/ubi:10.1-1776646961
RUN ln -s /usr/share/licenses /licenses
```

`quay.io/rh-ee-caxu/test-licenses-symlink:direct` - symlinks `/licenses -> /usr/share/licenses/foo`, and contains one non-empty text file at `/usr/share/licenses/foo/LICENSE`

```
FROM registry.redhat.io/ubi10/ubi:10.1-1776646961
RUN mkdir -p /usr/share/licenses/foo && \
    echo "bar" > /usr/share/licenses/foo/LICENSE && \
    ln -s /usr/share/licenses/foo /licenses
```

acornett21 (approved):

PR #1455, opened 2026-04-24
HasLicense: use resolved FileInfo for license walk entries

caxu-rh:
Wrap WalkDir DirEntry with resolvedDirEntry so validate()'s Info() reflects os.Stat after following symlinks, not symlink inode metadata.

Add tests: symlink to empty file, symlink to empty dir, symlink to dirs-only tree; keep existing /licenses -> directory with files coverage.

Made-with: Cursor

---

This is a follow-up for some corner cases missed in #1453 (and includes tests which should guard against regression). You can confirm the difference by turning on trace logging and noticing the `licenseCount` on `quay.io/rh-ee-caxu/test-licenses-symlink:latest` (should be 168, not 1).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved license discovery to correctly handle symlinked license directories, resolve mount roots, and prevent symlink-based path traversal so only valid regular files are considered.

* **Tests**
  * Added comprehensive tests covering symlink edge cases in license validation (including symlinked license dirs, escaped symlinks, empty targets and directory-only trees); symlink scenarios are platform-aware.

### DIFF
--- a/internal/policy/container/has_license.go
+++ b/internal/policy/container/has_license.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"path/filepath"
 	"slices"
+	"strings"
 
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/internal/check"
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/internal/image"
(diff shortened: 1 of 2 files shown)


internal/policy/container/has_license.go
```diff
@@ -55,16 +93,33 @@ func (p *HasLicenseCheck) getDataToValidate(ctx context.Context, mountedPath str
...
 		}
-		// Only include regular files, not directories
-		if !d.IsDir() {
-			files = append(files, d)
+		if d.IsDir() {
+			return nil
+		}
+		resolved, errEv := filepath.EvalSymlinks(p)
+		if errEv != nil || resolved == "" || !pathWithinMount(mountRoot, resolved) {
+			return nil
```
acornett21, line 117: Would we want to maybe log a message here at level `debug/trace` so we aren't silently skipping on broken links?

acornett21 (approved):

bcrochet:
@coderabbitai resume

bcrochet (approved):
/lgtm

PR #1466, opened 2026-05-13
build: switch golangci-lint from local install to go tool dependency

bcrochet:
Replace the manual golangci-lint binary installation via go-install-tool with the native Go tool directive (go tool golangci-lint). This removes the dedicated Makefile target and bin/ artifact in favor of Go's built-in tool management, simplifying the build setup and ensuring version consistency through go.mod.

An added benefit is that the tool version can be tracked by dependabot.

## Summary by CodeRabbit

* **Chores**
  * Streamlined development tool management through centralized Go modules configuration
  * Enabled automated dependency updates for build tooling and utilities

[Review Change Stack](https://app.coderabbit.ai/change-stack/redhat-openshift-ecosystem/openshift-preflight/pull/1466)


Makefile
```diff
@@ -40,16 +40,21 @@ endef
...
 	git diff --exit-code
 
 .PHONY: tidy
 tidy:
 	go mod tidy
 	git diff --exit-code
 
+.PHONY: tidy-tools
+tidy-tools:
+	go mod -modfile=tools/go.mod tidy
```
acornett21, line 55:
```suggestion
	go mod tidy -modfile=tools/go.mod
```

tools/go.mod
```diff
@@ -0,0 +1,220 @@
+module github.com/redhat-openshift-ecosystem/openshift-preflight/tools
+
+go 1.25.7
```
acornett21, line 3: I'd think we'd want to use the same version of go here, as the main module.

acornett21 (changes requested):
@bcrochet Not sure what you wanted to do with this PR. I thought the last time you mentioned this, was that dependabot did not actually do the updates when you tested in your fork?

If this does work and you want to iterate, below are a couple of comments. If not feel free to close, totally understand.

bcrochet:
Shelving.

PR #1467, opened 2026-05-16
build: set GOAMD64=v2 and GOPPC64=power8 for backward compatibility

### LINKED ISSUE #1339: Runtime error on AMD64 `v1/v2` processors: requires `v3` microarchitecture support
## Root Cause

   - The error originates from the Go runtime, indicating that the runtime architecture level is lower than the GOAMD64 version the binary was compiled with
   - The GOAMD64 default value was updated in UBI10 go-toolset to `v3`, while UBI9 uses `v2`, and golang default is `v1`
   - This change means binaries built with UBI10 are incompatible with older AMD64 processors that only support v2 microarchitecture

See below article for more info:
- https://access.redhat.com/solutions/7066628

vikasmulaje:
UBI10 go-toolset defaults to GOAMD64=v3 and GOPPC64=power9, which produces binaries that crash at startup on older processors that only support the v2 (amd64) or power8 (ppc64le) microarchitecture level.

Explicitly pin GOAMD64=v2 and GOPPC64=power8 in both the Dockerfile and the Makefile to restore the same hardware compatibility baseline that UBI9-built releases provided.

Fixes #1339

## Summary by CodeRabbit

* **Chores**
  * Enhanced compatibility with older processor architectures by configuring appropriate baseline defaults during the build process, ensuring the software runs reliably on older hardware.
  * Standardized microarchitecture configurations consistently across all build environments and platforms.
  * Maintained build flexibility by allowing compile-time customization of processor settings for different deployment scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/redhat-openshift-ecosystem/openshift-preflight/pull/1467?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,6 +13,12 @@ ARG OS
 # which prohibits copying from `/tmp` during make build cmd
 USER root
 
+# Override UBI10 Go toolset microarchitecture defaults to maintain backward
+# compatibility with the same hardware baseline as UBI9-built releases.
+# See https://github.com/redhat-openshift-ecosystem/openshift-preflight/issues/1339
+ENV GOAMD64=v2
+ENV GOPPC64=power8
+
 # Build the preflight binary
 COPY . /go/src/preflight
 WORKDIR /go/src/preflight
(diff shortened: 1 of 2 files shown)


caxu-rh (approved):

bcrochet (approved):
/lgtm

PR #1469, opened 2026-05-21
Filter test image layer hashes to remove irrelevant layers

komish:
The RootFS.Layer hash `sha256:5f70bf18a086007016e948b04aed3b82103a36bea41755b6cddfaf10ace3c6ef` effectively means "no content", and if found in a test image, could match with base images that also incorrectly have this layer hash as their final hash.

This PR adds a function to filter layers we want to exclude from the test image. It only excludes one layer for the moment, but it's been built to potentially exclude more if needed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improve certified image layer filtering to exclude a known empty layer before building filter lists, preventing empty layers from affecting results.

* **Tests**
  * Added tests for layer filtering: excluded-layer removal, absent excluded layer, empty input, input of only excluded layers, and multiple occurrences.

[Review Change Stack](https://app.coderabbit.ai/change-stack/redhat-openshift-ecosystem/openshift-preflight/pull/1469?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/internal/pyxis/layers.go
+++ b/internal/pyxis/layers.go
@@ -10,12 +10,46 @@ import (
 	"github.com/shurcooL/graphql"
 )
 
+// filterExcludedLayers removes layer hashes that are known to be common or
+// problematic when matching up against a base image.
+func filterExcludedLayers(uncompressedLayerHashes []cranev1.Hash) []cranev1.Hash {
+	excludedHashes := map[string]struct{}{
+		// This hash represents an empty layer, where the contents is only the
+		// tar end-of-stream marker 1kb of zeros (dd if=/dev/zero bs=1024
+		// count=1 2>/dev/null | sha256sum).
+		//
+		// Legacy build tools would add this before "empty_layer" became an
+		// option for operations like LABEL, ENV, etc. that did not modify the
+		// filesystem. There are also more modern cases where we see this behavior
+		// e.g. https://github.com/containers/buildah/issues/6860
+		//
+		// Because it is effectively an empty layer, we do not want to match
+		// against base images that also have an empty base layer as their final
+		// layer diff (any image can have an empty layer).
+		//
+		// For this reason, we will not send this to the backend to find an
+		// image that contains this layer
+		"sha256:5f70bf18a086007016e948b04aed3b82103a36bea41755b6cddfaf10ace3c6ef": {},
+	}
+
+	filtered := make([]cranev1.Hash, 0, len(uncompressedLayerHashes))
+	for _, layer := range uncompressedLayerHashes {
+		if _, isExcluded := excludedHashes[layer.String()]; !isExcluded {
+			filtered = append(filtered, layer)
+		}
+	}
+
+	return filtered
+}
+
 // CertifiedImagesContainingLayers takes uncompressedLayerHashes and queries to a Red Hat Pyxis,
 // returning existing certified images from registry.access.redhat.com that contain any of the
 // IDs as its uncompressed top layer id.
 func (p *pyxisClient) CertifiedImagesContainingLayers(ctx context.Context, uncompressedLayerHashes []cranev1.Hash) ([]CertImage, error) {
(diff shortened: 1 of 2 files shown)


internal/pyxis/layers.go
```diff
@@ -10,12 +10,46 @@ import (
 	"github.com/shurcooL/graphql"
 )
 
+// filterExcludedLayers removes layer hashes that are known to be common or
+// problematic when matching up against a base image.
+func filterExcludedLayers(uncompressedLayerHashes []cranev1.Hash) []cranev1.Hash {
+	excludedHashes := map[string]bool{
```
bcrochet: Is the bool necessary? Or are we just looking for existence in the set?

I would lean toward the "zero cost" map[string]struct{}, but if you feel the bool makes it easier to read than

```
if _, matches := excludedHashes[layer.String()]; !matches {
   ...
}
```

Or whatever you want to call that...

bcrochet (commented):
Leaving as a comment and not a request for changes. Do as you will with it.

internal/pyxis/layers.go
komish: Commit https://github.com/redhat-openshift-ecosystem/openshift-preflight/commit/efc928bfecf3e35f42f5490b823209d90bf93aa2 should implement this to keep it consistent.

caxu-rh (commented):
Is this omitting of layers (the particular one listed here) also relevant/appropriate for non-UBI images (images which have an exception to be based on `scratch`)?

Would this layer show up in `FROM scratch` and in such a case do we want to report it?

komish:
@caxu-rh 

Scratch exceptions shouldn't be in scope, as they [do not have BasedOnUBI in their policy](https://github.com/redhat-openshift-ecosystem/openshift-preflight/blob/9c826fb95055f84cabb51e2626be13cd9ffe1286/internal/engine/engine.go#L715-L733).

Let me know if that doesn't answer your question - I'm not sure if I understood it fully.

caxu-rh:
You answered it, thanks. Forgot about that piece in `engine` that opts in to the specific relevant checks depending on policy.

bcrochet (approved):
/lgtm

komish:
https://github.com/redhat-openshift-ecosystem/openshift-preflight/commit/4f5e808f00f9486e2a2d6b2eddcd8b6192b0c6a0 should contain fixes to code rabbit's review.

cc @acornett21

acornett21 (approved):

tonyskapunk:
> from change #1469:
> 
>     * FAILURE https://www.distributed-ci.io/jobs/5d4ca11b-fb33-4e8f-9e6c-c15a3f6983c3/jobStates

The failure is due to:
```
could not submit to pyxis: could not create test results: status code: 400: body: {\"type\": \"about:blank\", \"title\": \"Bad Request\", \"detail\": \"Validation error: 'github.com/redhat-openshift-ecosystem/openshift-preflight' version '1.19.0' is not supported. Supported versions are: ['1.16.0', '1.17.0', '1.17.1', '1.17.2', '1.18.0']\
```
Not sure if this is something that needs attention on the pyxis side.

acornett21:
@tonyskapunk This has to be some weird timing issue where the above CI pulled the released image while CI was going on, but had yet to update pyixs. Below shows it was done in the release.

```
❯ curl -s https://catalog.redhat.com/api/containers/v1/tools?page_size=100 | jq '.data | map(select(.enabled_for_testing == true))'
[
  {
    "_id": "6a15eb8fea68ae777e41cda7",
    "commit": "8a0341d649c56b0052f96079bd8426880385298a",
    "creation_date": "2026-05-26T18:50:55.436000+00:00",
    "enabled_for_testing": true,
    "last_update_date": "2026-05-26T18:50:55.436000+00:00",
    "version": "1.19.0"
  }
]
```
https://github.com/redhat-openshift-ecosystem/openshift-preflight/actions/runs/26466977169/job/77934828566

Are you able to re-run or confirm my suspession?

PR #1498, opened 2026-07-20
move operator-sdk install to builder, switch to local gpg key verify

acornett21:
## Summary by CodeRabbit

* **Bug Fixes**
  * Hardened container builds by verifying the Operator SDK binary with a bundled release public key, including fingerprint and signature/checksum validation, before copying artifacts into the runtime image.
  * Ensured required CA trust files are carried into the micro runtime so HTTPS works reliably.
* **Chores**
  * Reduced runtime image footprint by switching to a micro-focused UBI base image.
  * Note: This added the need to copy in CA certs so `crypto/tls` during container pulls still work.
* **Documentation**
  * Added guidance for updating the bundled Operator SDK signing key via `make update-operator-sdk-key`, including how to override the signing fingerprint.


caxu-rh (approved):

acornett21:
Putting this PR on hold, since I'm investigating if `operator-sdk` can go away as a dependency. We'd likely keep the switch to `ubi-micro`, but not the rest.

acornett21:
Going to close this in favor of implementing the below:
- #1499

PR #1500, opened 2026-07-24
deployablebyolm: add random namespace suffix to avoid collisions

### LINKED ISSUE #1355: Allow set install namespace for OLM deployable test

I've been trying to incorporate the preflight check operator into our pipeline to avoid issues during the final certification period, the big problem that I have faced is that it requires an OpenShift cluster, this it's perfect, but spinning a cluster every time that we need to check the CI it's a lot of time wasted on this, but I was thinking on running this test against an already running OpenShift cluster, but when running two test in parallel conflicts due to preflight using the package name as the target namespace to install the operator here https://github.com/redhat-openshift-ecosystem/openshift-preflight/blob/main/internal/policy/operator/deployable_by_olm.go#L313

I've been browsing the code and thinking on adding this as an option to the preflight run and replace the value here https://github.com/redhat-openshift-ecosystem/openshift-preflight/blob/main/internal/policy/operator/deployable_by_olm.go#L313 and avoid using the `appName` as the namespace.

In short, a new variable to specify the namespace to test the operator install process will be amazing.

There's a variable PFLT_NAMESPACE and argument `--namespace` that contains the namespace for the scorecard, probably we can use the same one to use only one namespace for everything ?

## Additional context.

This will be really useful to run multiple checks in one cluster and also to trigger [operator-pipeline](https://github.com/redhat-openshift-ecosystem/operator-pipelines) multiple times in the same cluster which is my ultimate goal

acornett21:
- Fixes: #1355 

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduced operator namespace collisions by generating install and target namespaces from the app name, truncating to DNS-label limits and appending an optional suffix when needed.

* **Tests**
  * Expanded coverage for namespace suffixing and truncation, including new checks for `operatorMetadata` namespace generation.
  * Updated existing test fixtures and added long-name OLM manifest/annotation fixtures to validate metadata and namespace behavior.

### DIFF
--- a/internal/policy/operator/deployable_by_olm.go
+++ b/internal/policy/operator/deployable_by_olm.go
@@ -24,6 +24,7 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	apiruntime "k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/rand"
 	"k8s.io/apimachinery/pkg/util/validation"
 	"k8s.io/client-go/kubernetes"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -69,6 +70,7 @@ type DeployableByOlmCheck struct {
 	validImages         bool
 	csvTimeout          time.Duration
 	subscriptionTimeout time.Duration
+	namespaceSuffix     func(int) string
 }
 
 func (p *DeployableByOlmCheck) initClient() error {
@@ -153,9 +155,10 @@ func NewDeployableByOlmCheck(
 	opts ...Option,
 ) *DeployableByOlmCheck {
 	c := &DeployableByOlmCheck{
-		dockerConfig: dockerConfig,
-		indexImage:   indexImage,
-		channel:      channel,
+		dockerConfig:    dockerConfig,
+		indexImage:      indexImage,
+		channel:         channel,
+		namespaceSuffix: rand.String,
 	}
 
 	for _, opt := range opts {
@@ -332,13 +335,18 @@ func (p *DeployableByOlmCheck) operatorMetadata(ctx context.Context, bundleRef i
 		deploymentNames = append(deploymentNames, deployment.Name)
 	}
 
+	namespace := appName[:min(len(appName), maxAppNameLen)]
+	if suffix := p.namespaceSuffix(namespaceSuffixLen); suffix != "" {
+		namespace = fmt.Sprintf("%s-%s", namespace, suffix)
+	}
+
 	return &operatorData{
 		CatalogImage:     catalogImage,
 		Channel:          channel,
 		PackageName:      packageName,
 		App:              appName,
-		InstallNamespace: appName,
-		TargetNamespace:  appName + "-target",
+		InstallNamespace: namespace,
+		TargetNamespace:  namespace + targetSuffix,
 		InstallModes:     installModes,
 		DeploymentNames:  deploymentNames,
 	}, nil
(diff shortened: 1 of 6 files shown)


acornett21:
@coderabbitai full review

caxu-rh (approved):

acornett21:
@bcrochet Are you okay with this change, since you commented on the original issue, just want to be sure? Thanks!

internal/policy/operator/deployable_by_olm_test.go
```diff
@@ -37,6 +37,7 @@ var _ = Describe("DeployableByOLMCheck", func() {
 		now := metav1.Now()
 		og.Status.LastUpdated = &now
 		deployableByOLMCheck = *NewDeployableByOlmCheck("test_indeximage", "", "", WithCSVTimeout(1*time.Second), WithSubscriptionTimeout(1*time.Second))
+		deployableByOLMCheck.namespaceSuffix = func(int) string { return "" }
```
bcrochet: Why define this base version differently than the "default"? Shouldn't this default to a len(5) string?

internal/policy/operator/deployable_by_olm.go
```diff
@@ -332,13 +335,20 @@ func (p *DeployableByOlmCheck) operatorMetadata(ctx context.Context, bundleRef i
 		deploymentNames = append(deploymentNames, deployment.Name)
 	}
 
+	// random suffix avoids namespace collisions during parallel runs
+	// keep TargetNamespace (appName + "-" + suffix + "-target") within the 63-char DNS label limit
+	namespace := appName[:min(len(appName), 50)]
+	if suffix := p.namespaceSuffix(5); suffix != "" {
```
bcrochet: These "magic numbers" should be constants for future readability. And if the 50 could be calculated, even better.

bcrochet (changes requested):
Couple of requests.

internal/policy/operator/deployable_by_olm.go
acornett21: I've moved these values to `default` and used some of the built in k8s defaults to construct this which should be easier to read.

internal/policy/operator/deployable_by_olm_test.go
acornett21: I've updated this test with an `abcde` like the other tests, which is what I think you are asking.

acornett21:
@bcrochet I think I have addressed your comments, can you re-review?

internal/policy/operator/deployable_by_olm_test.go
```diff
@@ -160,6 +161,30 @@ var _ = Describe("DeployableByOLMCheck", func() {
 		})
 	})
 
+	Describe("operatorMetadata namespace generation", func() {
+		Context("When namespaceSuffix returns a non-empty value", func() {
+			It("Should append the suffix to InstallNamespace and TargetNamespace", func() {
+				deployableByOLMCheck.namespaceSuffix = func(int) string { return "abcde" }
```
bcrochet: This shouldn't still be necessary?

internal/policy/operator/deployable_by_olm_test.go
```diff
@@ -160,6 +161,30 @@ var _ = Describe("DeployableByOLMCheck", func() {
...
+				data, err := deployableByOLMCheck.operatorMetadata(testcontext, imageRef)
+				Expect(err).ToNot(HaveOccurred(), "failed to load operator metadata")
+				Expect(data.InstallNamespace).To(Equal("p-testPackage-abcde"), "install namespace should include the generated suffix")
+				Expect(data.TargetNamespace).To(Equal("p-testPackage-abcde-target"), "target namespace should derive from the suffixed install namespace")
+			})
+		})
+		Context("When appName exceeds the maximum length", func() {
+			It("Should truncate to keep namespaces within the 63-char DNS label limit", func() {
+				deployableByOLMCheck.namespaceSuffix = func(int) string { return "abcde" }
```
bcrochet: Same.

bcrochet (changes requested):
Just two more bits.

internal/policy/operator/deployable_by_olm_test.go
acornett21: correct, the `BeforeEach` now covers this setup.

bcrochet (approved):
/lgtm

PR #1507, opened 2026-07-31
engine: eagerly pull image layers with retry before extraction

bcrochet:
## Description

`crane.Pull`/`cache.Image` only fetch the image manifest eagerly; each layer's bytes were previously fetched lazily, the first time they were read during `untar`. go-containerregistry's retry support only retries the initial HTTP round trip, not a body read that fails partway through streaming a layer, so a transient network drop mid-layer surfaced as a hard failure deep inside tar extraction.

This adds `pullLayers`, which eagerly downloads every layer's uncompressed content into the on-disk filesystem cache (keyed by `DiffID`, the same lookup `untar`'s use of `mutate.Extract` performs) using a bounded concurrency pool, retrying each layer independently with backoff on failure. This isolates registry/network flakiness to a single step before extraction begins, so `untar` only ever operates on local data.

## Testing

- `make vet`
- `make lint`
- `make fmt` / `make tidy` (no diffs)
- `make test`
- `make cover` (100% coverage maintained)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved container image processing by downloading and verifying all image layers before extraction.
  * Added automatic retries with backoff for temporary layer download failures.
  * Added clearer error reporting for unavailable, corrupted, or incomplete image layers.
  * Improved handling of cancellations, interrupted transfers, cache inconsistencies, and failed retries.
  * Limited simultaneous layer downloads to improve stability and resource usage.
  * Added integrity checks to detect corrupted layer contents before use.

### DIFF
--- a/internal/engine/pull.go
+++ b/internal/engine/pull.go
@@ -1,0 +1,150 @@
+package engine
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"time"
+
+	"github.com/go-logr/logr"
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/cache"
+	"golang.org/x/sync/errgroup"
+
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/internal/log"
+)
+
+// errLayerContentMismatch indicates that a layer's content, once fully read,
+// does not hash to the digest the layer itself reports. This can happen when
+// a previously-cached copy of the layer on disk is corrupted or was left
+// truncated by an interrupted download (see pullLayerWithRetry).
+var errLayerContentMismatch = errors.New("layer content does not match expected digest")
+
+const (
+	// pullLayerMaxAttempts is the number of times to attempt to fully download a
+	// single layer before giving up. go-containerregistry's own retry support
+	// (remote.WithRetryBackoff, etc.) only retries the initial HTTP round trip; it
+	// does not retry a read that fails partway through streaming a layer's body.
+	// This is the layer of retry that covers that gap.
+	pullLayerMaxAttempts = 3
+	// pullLayerConcurrency bounds how many layers are pulled at the same time,
+	// mirroring crane's own default job concurrency.
+	pullLayerConcurrency = 4
+)
+
+// pullLayerRetryBaseDelay is the initial backoff delay between failed
+// attempts to pull a single layer. Subsequent attempts double this delay.
+// It's a var (rather than a const) so tests can shrink it.
+var pullLayerRetryBaseDelay = 2 * time.Second
+
+// pullLayers eagerly downloads the full, uncompressed content of every layer in
+// img, verifying it against the layer's own digest. Reading through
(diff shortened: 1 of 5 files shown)


internal/engine/pull.go
```diff
@@ -0,0 +1,117 @@
...
+		return nil
+	}
+
+	return fmt.Errorf("failed to pull layer %s after %d attempts: %w", diffID, pullLayerMaxAttempts, lastErr)
+}
+
+// pullLayerOnce makes a single attempt to fully read a layer's uncompressed
+// content, discarding the bytes as they're read.
+func pullLayerOnce(layer v1.Layer) error {
+	rc, err := layer.Uncompressed()
```
nammn, line 134: Could we have stale cache layers? Example the case when a single layer failed and writes a stale cache but succeeds at the second try? 

i was digging a bit on the caching layer you mentioned and I couldn't find in the layer where we "throw away" the broken layer. It seems the dep just writes to the cache file as they go: https://github.com/google/go-containerregistry/blob/v0.21.7/pkg/v1/cache/fs.go#L83-L95

and re-use them if it exists: https://github.com/google/go-containerregistry/blob/v0.21.7/pkg/v1/cache/fs.go#L120-L132

am I over complicating things?

nammn (commented):
Thanks a lot! One Q regarding layer caching

internal/engine/pull.go
bcrochet: If you read the second link you sent more closely, you'll see that the validation of the layer happens in the `tarball.LayerFromFile` call. It will catch an incomplete tarball and return the UnexpectedEOF. That triggers a delete of the incomplete file.

nammn: From reading the code it seems that we only return `io.ErrUnexpectedEOF` if we go into `UnzipReadCloser`.  `Uncompress`  with raw `tar` files doesn't seem to go there.
So to my limited understanding it seems that the deletion logic only runs if the cache file itself is compressed and its decompressor fails during `fs.Get`

bcrochet: The compressed and the uncompressed layers will have different digests. The cache layer only caches the uncompressed layer. So, if the error occurs before being written as the uncompressed layer, it will not be cached anyway. And if the error occurred during decompression, it would be an invalid file, and get handled as such. 

~~I haven't found the exact "a-ha" spot in googe/go-containerregistry, but I have high confidence that this will work as intended.~~

I stand corrected. I will add some additional protection against this scenario. Good catch.

bcrochet: I've updated the PR. LMKWYT.

nammn: thanks a lot - this is great - once this is out, we will update and use it!

nammn (approved):
thanks a lot!

acornett21:
I've tested this PR and the memory profile is similar to what `main`'s profile is, as well as the profiles we have from when `preflight` was having performance issues. 

I put this on hold for now, to merge other PR's and do a release. This would be slated for a release next week or the week after.

bcrochet:
@acornett21 Ready to remove the hold?

PR #1515, opened 2026-08-06
feat: add --exit-with-failure flag for non-zero exit on check failures

### LINKED ISSUE #1235: Add flag to exit with non-zero status code when not passing
We use preflight check in our CI infrastructure to ensure our images will pass the preflight check once we submit it. There is currently no way to tell preflight check to exit with a non-zero code in case of any failed tests.

It would be helpful to have a flag like `--exit-with-failure` or something similar, that will return a non-zero status code when the preflight check did not pass. The [operator pipeline](https://github.com/redhat-openshift-ecosystem/operator-pipelines) has `--exit-with-pipelinerun-error` for example. 

The workaround is to parse the output and check if `passed` is true.

hardengl:
## Summary

Add an opt-in `--exit-with-failure` flag that causes preflight to exit with a non-zero status code when checks fail (exit 2) or encounter errors (exit 1). Default behavior (exit 0 regardless of check results) is preserved for backward compatibility.

This is a re-implementation of reverted PR #1316, with both bugs that caused the revert now fixed:
1. Default behavior is **unchanged** — preflight exits 0 regardless of check results
2. Flag logic is consistent across all code paths (container and operator modes)

## Architecture

```
cobra flag (check.go PersistentFlag)
  → viper.BindPFlag("exit_with_failure")
    → runtime.Config.ExitWithFailure (NewConfigFrom)
      → CheckConfig.ExitOnFailure (passed from RunE)
        → RunPreflight returns typed error
          → main.go maps to exit code
```

- Flag defined in \`check.go\` as PersistentFlag (shared by both subcommands)
- Config read through \`runtime.Config\` (correct pattern, not direct viper reads)
- Check commands (\`check_container.go\`, \`check_operator.go\`) don't contain any flag-specific logic — they just pass \`cfg.ExitWithFailure\` naturally
- \`RunPreflight\` returns \`ChecksErroredError\` or \`ChecksFailedError\` typed errors
- \`main.go\` translates typed errors to exit codes (1 for errors, 2 for failures)
- Both error types implement \`Is(target error) bool\` for proper \`errors.Is\` matching

## Exit Codes

| Condition | Exit Code |
|-----------|-----------|
| All checks pass | 0 |
| Flag not set (any result) | 0 |
| Checks errored + flag set | 1 |
| Checks failed + flag set | 2 |

## Testing

- Tests added to existing \`internal/cli/cli_test.go\` (20/20 pass)
- Covers: failures, errors, priority (errors > failures), success, disabled mode, results-file-written-before-exit
- Error type matching via \`errors.Is\` tested
- No new test failures introduced (6 pre-existing cmd test failures are ARM fixture issues on main)

## Documentation

- \`docs/CONFIG.md\` updated with \`PFLT_EXIT_WITH_FAILURE\` entry

Closes #1235

### DIFF
--- a/cmd/preflight/cmd/check_container.go
+++ b/cmd/preflight/cmd/check_container.go
@@ -182,6 +182,7 @@ func checkContainerRunE(cmd *cobra.Command, args []string, runpreflight runPrefl
 			cli.CheckConfig{
 				IncludeJUnitResults: cfg.WriteJUnit,
 				SubmitResults:       cfg.Submit,
+				ExitOnFailure:       cfg.ExitWithFailure,
 			},
 			formatter,
 			&runtime.ResultWriterFile{},
--- a/internal/cli/cli.go
+++ b/internal/cli/cli.go
@@ -21,6 +21,29 @@ import (
 type CheckConfig struct {
 	IncludeJUnitResults bool
 	SubmitResults       bool
+	ExitOnFailure       bool
+}
+
+// ChecksFailedError is returned when one or more checks have a "failed" result
+// and ExitOnFailure is enabled.
+type ChecksFailedError struct{}
+
+func (e *ChecksFailedError) Error() string { return "one or more checks failed" }
+
+func (e *ChecksFailedError) Is(target error) bool {
+	_, ok := target.(*ChecksFailedError)
+	return ok
+}
+
+// ChecksErroredError is returned when one or more checks encountered an error
+// during execution and ExitOnFailure is enabled.
+type ChecksErroredError struct{}
+
+func (e *ChecksErroredError) Error() string { return "one or more checks encountered an error" }
+
+func (e *ChecksErroredError) Is(target error) bool {
+	_, ok := target.(*ChecksErroredError)
+	return ok
 }
 
 var stdout io.Writer = os.Stdout
@@ -88,6 +111,15 @@ func RunPreflight(
 	}
 
 	logger.Info(fmt.Sprintf("Preflight result: %s", convertPassedOverall(results.PassedOverall)))
+
+	if cfg.ExitOnFailure {
+		if len(results.Errors) > 0 {
+			return &ChecksErroredError{}
+		}
+		if len(results.Failed) > 0 {
+			return &ChecksFailedError{}
+		}
+	}
 
 	return nil
 }
--- a/cmd/preflight/cmd/check.go
+++ b/cmd/preflight/cmd/check.go
@@ -23,6 +23,9 @@ func checkCmd() *cobra.Command {
 	checkCmd.PersistentFlags().String("artifacts", "", "Where check-specific artifacts will be written. (env: PFLT_ARTIFACTS)")
 	_ = viper.BindPFlag("artifacts", checkCmd.PersistentFlags().Lookup("artifacts"))
 
+	checkCmd.PersistentFlags().Bool("exit-with-failure", false, "Exit with a non-zero status code when checks fail or encounter errors. (env: PFLT_EXIT_WITH_FAILURE)")
+	_ = viper.BindPFlag("exit_with_failure", checkCmd.PersistentFlags().Lookup("exit-with-failure"))
+
 	checkCmd.AddCommand(checkOperatorCmd(cli.RunPreflight))
 	checkCmd.AddCommand(checkContainerCmd(cli.RunPreflight))
 
--- a/cmd/preflight/cmd/check_operator.go
+++ b/cmd/preflight/cmd/check_operator.go
@@ -139,6 +139,7 @@ func checkOperatorRunE(cmd *cobra.Command, args []string, runpreflight runPrefli
 		cli.CheckConfig{
 			IncludeJUnitResults: cfg.WriteJUnit,
 			SubmitResults:       false, // operator results are not submitted.
+			ExitOnFailure:       cfg.ExitWithFailure,
 		},
 		formatter,
 		&runtime.ResultWriterFile{},
--- a/cmd/preflight/main.go
+++ b/cmd/preflight/main.go
@@ -1,15 +1,28 @@
 package main
 
 import (
-	"log"
+	"errors"
+	"fmt"
+	"os"
 
 	"github.com/redhat-openshift-ecosystem/openshift-preflight/cmd/preflight/cmd"
+	"github.com/redhat-openshift-ecosystem/openshift-preflight/internal/cli"
 )
 
 func main() {
 	//coverage:ignore
 	if err := cmd.Execute(); err != nil {
 		//coverage:ignore
-		log.Fatal(err)
+		switch {
+		case errors.Is(err, &cli.ChecksErroredError{}):
+			fmt.Fprintln(os.Stderr, err)
+			os.Exit(1)
+		case errors.Is(err, &cli.ChecksFailedError{}):
+			fmt.Fprintln(os.Stderr, err)
+			os.Exit(2)
+		default:
+			fmt.Fprintln(os.Stderr, err)
+			os.Exit(1)
+		}
 	}
 }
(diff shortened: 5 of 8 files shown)


cmd/preflight/cmd/check_container.go
```diff
@@ -140,6 +141,12 @@ func checkContainerRunE(cmd *cobra.Command, args []string, runpreflight runPrefl
 		return err
 	}
 
+	exitOnFailure := viper.Instance().GetBool("exit_with_failure")
```
acornett21: This should be added to the `config.go` and then the config should be used for this.

cmd/preflight/cmd/check_container.go
```diff
@@ -140,6 +141,12 @@ func checkContainerRunE(cmd *cobra.Command, args []string, runpreflight runPrefl
...
+
+	// Track the worst check result across all platforms so we can report
+	// an appropriate exit code after processing all platforms.
+	var worstCheckErr error
```
acornett21: This doesn't really read clearly as what is trying to be accomplished, this also makes the logic below overly nested and hard to determine what the actually cause of the error was. I don't have a great name suggestion, right now.

cmd/preflight/cmd/root.go
```diff
@@ -61,6 +61,9 @@ func rootCmd() *cobra.Command {
 	rootCmd.PersistentFlags().String("loglevel", "", "The verbosity of the preflight tool itself. Ex. warn, debug, trace, info, error. (env: PFLT_LOGLEVEL)")
 	_ = viper.BindPFlag("loglevel", rootCmd.PersistentFlags().Lookup("loglevel"))
 
+	rootCmd.PersistentFlags().Bool("exit-with-failure", false, "Exit with a non-zero exit code when checks fail (exit 2) or encounter errors (exit 1). (env: PFLT_EXIT_WITH_FAILURE)")
```
acornett21: This new value needs to be added to CONFIG.md

acornett21 (changes requested):

internal/cli/cli.go
```diff
@@ -21,6 +21,19 @@ import (
...
+type ChecksFailedError struct{}
+
+func (e *ChecksFailedError) Error() string {
+	return "one or more checks failed"
+}
+
+type ChecksErroredError struct{}
+
+func (e *ChecksErroredError) Error() string {
+	return "one or more checks encountered an error"
```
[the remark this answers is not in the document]
acornett21: ➕

cmd/preflight/cmd/check_container.go
```diff
@@ -182,12 +189,21 @@ func checkContainerRunE(cmd *cobra.Command, args []string, runpreflight runPrefl
...
 				IncludeJUnitResults: cfg.WriteJUnit,
 				SubmitResults:       cfg.Submit,
+				ExitOnFailure:       exitOnFailure,
 			},
 			formatter,
 			&runtime.ResultWriterFile{},
 			resultSubmitter,
 		); err != nil {
-			return err
+			if errors.Is(err, &cli.ChecksErroredError{}) || errors.Is(err, &cli.ChecksFailedError{}) {
```
acornett21: There isn't anything that tests this logic.  `check_container_test.go` already has an injectable runpreflight mock and a multi-platform manifest list fixture (manifestListSrc) — a mock that returns a different result per call would let us assert all platforms get processed and errors win over failures.

internal/cli/errors_test.go
```diff
@@ -0,0 +1,259 @@
+package cli
```
acornett21: cli_test.go already exists, these should be folded into that test file.

acornett21:
@caxu-rh Would you mind reviewing this as well, since you implemented the original fix, and I know you have thoughts on what the proper approach should have been?

caxu-rh:
Left a few reactions on some of the same things you pointed out. Overall I'm still more partial to checking the flag and deciding whether to bubble up errors in `cmd/preflight/cmd/root.go` and keeping `cmd/preflight/cmd/check_{container,operator}.go` not caring about whether the exit-with-failure flag is set.

hardengl:
## Verification Summary

This is a re-implementation of reverted PR #1316, with both bugs that caused the revert now fixed.

**Background**: PR #1316 added `--exit-with-failure` but was reverted due to:
1. It changed the default exit behavior (breaking existing CI pipelines)
2. The flag logic was inverted in some code paths

**This PR fixes both issues**:
1. Default behavior is **unchanged** — preflight exits 0 regardless of check results (preserving backward compatibility)
2. The `--exit-with-failure` flag is opt-in only and correctly returns non-zero when ANY check fails
3. Flag logic is consistent across all code paths (container and operator modes)

**Testing**:
- Unit tests cover both `--exit-with-failure` enabled and disabled paths
- Verified default behavior (exit 0 on check failure) is preserved
- Verified `--exit-with-failure` correctly returns exit code 1 when checks fail
- Verified `--exit-with-failure` returns exit code 0 when all checks pass

**Impact**: Enables CI/CD pipelines to gate on preflight results without parsing output, while maintaining full backward compatibility for existing users.

Ready for review and merge. @komish @bcrochet

hardengl:
@acornett21 @caxu-rh Thank you both for the detailed review! I agree with the feedback:

1. **Move flag to `config.go`** — will add `ExitWithFailure` to the viper config struct
2. **Add to `CONFIG.md`** — will document the new config option
3. **Move exit logic to `root.go`** — per @caxu-rh's suggestion, the check commands shouldn't care about this flag. The decision of whether to exit non-zero should happen at the root level after all checks complete
4. **Add tests** — will use the existing `runpreflight` mock pattern in `check_container_test.go`
5. **Simplify nesting** — will flatten the logic with the new approach

I'll push an updated commit addressing all of these. Thanks for pointing me to the right patterns!

hardengl:
## Full Rework Pushed

Addressed all review feedback from @acornett21 and @caxu-rh:

### Changes Made

1. **Flag moved to `check.go`** as a `PersistentFlag` (shared by both `check container` and `check operator`)
2. **Config flows through `runtime.Config`** — added `ExitWithFailure` field, read from viper in `NewConfigFrom`
3. **Documented in `docs/CONFIG.md`** — `PFLT_EXIT_WITH_FAILURE` with exit code semantics
4. **Check commands don't care about the flag** — they just pass `cfg.ExitWithFailure` to `CheckConfig.ExitOnFailure` (same pattern as `WriteJUnit` and `Submit`)
5. **`RunPreflight` returns typed errors** — `ChecksErroredError` (exit 1) or `ChecksFailedError` (exit 2), with errors taking priority
6. **`main.go` handles exit code mapping** — translates typed errors to os.Exit codes
7. **`Is()` methods added** to both error types for proper `errors.Is` matching across instances
8. **Tests folded into `cli_test.go`** (not a separate file) — covers failures, errors, priority, success, disabled mode, and results-file-written-before-exit
9. **All assertions have descriptive messages** and stdout is properly saved/restored with `BeforeEach`/`AfterEach`

### Architecture Summary

```
cobra flag (check.go PersistentFlag)
  → viper.BindPFlag("exit_with_failure")
    → runtime.Config.ExitWithFailure (NewConfigFrom)
      → CheckConfig.ExitOnFailure (passed from RunE)
        → RunPreflight returns typed error
          → main.go maps to exit code
```

### Test Results
- `internal/cli`: 20/20 pass ✅
- `cmd/preflight/cmd`: 91/97 pass (6 pre-existing failures on main — ARM test fixture issue, unrelated)

All feedback items addressed. Ready for re-review.

hardengl:
@acornett21 @caxu-rh — Full rework is pushed addressing all your feedback:
- Flag in `check.go` PersistentFlag
- Config through `runtime.Config` (not direct viper reads)
- Tests in `cli_test.go` (not separate file)
- `CONFIG.md` updated
- Check commands don't know about the flag — just pass cfg naturally
- `main.go` handles exit codes

DCI passed. Would appreciate a re-review + `/ok-to-test` when you get a chance. Thanks for the guidance on the architecture!

